PR #8, opened 2017-09-06
MOD: clarified phrasing

lianilychee:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,14 @@
-How to work on the tool on your local machine:
+## Installation and Setup for Local Development
 
-1. Make sure you have [Node.js](https://nodejs.org/en/) installed (I used v6.11.1).
+1. Install [Node.js](https://nodejs.org/en/) (I used v6.11.1).
 2. Open the node command prompt and change into the directory that you want to clone the app to.
 3. Clone the repo to your local machine with `git clone https://github.com/adanaher/cliff-effects.git`.
-4. Change into the new app directory with `cd cliff-effects`.
+4. Change into the new app directory with `cd cliff-effects`.  For Windows machines, use `dir` instead of `cd`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,14 +1,14 @@
-How to work on the tool on your local machine:
+## Installation and Setup for Local Development
 
-1. Make sure you have [Node.js](https://nodejs.org/en/) installed (I used v6.11.1).
+1. Install [Node.js](https://nodejs.org/en/) (I used v6.11.1).
```
knod: Maybe "If you don't have [node.js](...) already, install it. ... To check if you have node.js, open a terminal window and type `node --version`. If a number of the format #.#.# gets printed, you have node. If not, you don't.

I dunno. Something like that.

README.md
```diff
@@ -1,14 +1,14 @@
...
 2. Open the node command prompt and change into the directory that you want to clone the app to.
 3. Clone the repo to your local machine with `git clone https://github.com/adanaher/cliff-effects.git`.
-4. Change into the new app directory with `cd cliff-effects`.
+4. Change into the new app directory with `cd cliff-effects`.  For Windows machines, use `dir` instead of `cd`.
```
knod: We talked about this, but as a refresher - `dir` creates a directory while `cd` navigates to it, so this edit needs to be removed.

lianilychee: Noted.  Just FYI, just double checked, and `dir` is definitely analog to `ls`, while 'mkdir' creates.

lianilychee:
@knod requested changes implemented.

README.md
knod: Thanks for the clarification! The making a directory thing was a guess of ours (neither of us have a windows machine), but the cd thing was for sure. Good to know about `mkdir`.

Issue #12, opened 2017-09-06
MOD:  "Household Annual Income" Page Percentages


lianilychee:
* [x] Verify CALC of "% of Federal Poverty Level"
* [x] Verify SOURCE of "% of Federal Poverty Level"
* [x] Replace "% of State Median Income" with "Area Median Income"
* [x] Switch phrasing of title from "% of Federal Poverty Level" -> "Federal Poverty Level Percentage".  Do the same for State Med level.
* [ ] Add reference to source of calc. info, incl. year, as clear linked text right underneath associated %ages in comment in codebase

danaher2:
Please provide a mockup of how the percentages should look with the references below.

lianilychee:
![image](https://user-images.githubusercontent.com/6895034/30123496-2d16b490-9300-11e7-957e-d26f49feef3c.png)

lianilychee:
![image](https://user-images.githubusercontent.com/6895034/30123623-a371fce4-9300-11e7-8455-53ec8126c43a.png)

`2010 Nat'l Census data` is stand-in text for the actual sources.

knod:
@lianilychee: Wouldn't it be more important to show the expiration date of the document?

**Technical notes:**
- If so, maybe we can use that nifty expiration date idea to fill in things like dates
- Either way, we may want to make the same type of provisions for where the data came from. For example, return a `sources` object as part of the result. Ex:
```js
result.sources = {
  primary: {
    label: "2010 National Census Data",
    link: "http://national-census-data.com" // or an html string or a React component
  },
  // ...
}
```

AnnieLa:
https://www.zanebenefits.com/blog/2017-federal-poverty-level-guidelines 

Use the data found on this blog as the basis for federal poverty levels by household size for 2017

zzyyfff:
Area Median Income calculation resources:

https://www.huduser.gov/portal/datasets/il.html#2017_data
https://www.huduser.gov/portal/datasets/il/il17/HUD-sec8-FY17.pdf

lianilychee:
@knod comment obsolete now; we just removed the citation from the front-end.

knod:
@lianilychee: If you no longer want the citation, could you remove it from the list of things to do? It's #5, right?

knod:
A note: We can't figure out Area Median Incomes in the way we can figure out State Median Incomes and Federal Poverty Levels. Area median incomes will be either 30%, 50%, or 80%. Is that still useful?

[...]

PR #43, opened 2017-09-20
Addresses Issue #12 : Two items addressed

WillMcIntosh:
Replaced "% of State Median Income" with "Area Median Income Percentage"
Switched phrasing of "% of Federal Poverty Level" to "Federal Poverty Level Percentage"
I couldn't find "State Med Level" so that item is still unchecked

Additional notes: I got a package-lock.json file when I first ran npm run build. I talked with Moss and we decided to include it to be tracked. I had also added *.swp to gitignore since vim creates temp files when editing, but then removed it.


knod:
Just a note: the value currently paired with the 'Area Median Income Percentage' label doesn't actually represent the area median income percentage value, and that value is possibly not calculable in the way that is desired. There's a comment on it in the thread. I'm going to merge this pull request and then hide that element just in case it slips our mind to make it accurate.

Other than that, looks great!

PR #59, opened 2017-09-26
Fixes graph calculations changing values of client

knod:


WillMcIntosh:
I  cloned this bugs branch onto my machine to test the graph changing the client object. This PR creates a new client object which fixed this bug for me.

knod:
To clarify the purpose of the pull request, to make the graph work, the code now creates a new object containing the values of the current client object, and then only the new object's values get changed.

PR #63, opened 2017-10-03
Created a test suite for the roundMoney function in math.js

yhou8:
Right now roundMoney fails to pass 2 tests that I think are important. The most important is that it fails to  round 1.005 to 1.01, which is a result of representing money using binary floating point. We might want to switch to a decimal number library after the mvp so results are more consistent with what people expect. I also feel that roundMoney should convert -0 to 0 to reduce the possibility of displaying -0 somewhere on the website since the concept of -0 can be confusing.

### DIFF
--- a/src/helpers/math.test.js
+++ b/src/helpers/math.test.js
@@ -1,0 +1,33 @@
+const math = require('./math');
+
+test('Rounding up', () => {
+    expect(math.roundMoney(1.005)).toBe(1.01);
+});
+
+test('Rounding down', () => {
+    expect(math.roundMoney(1.0049)).toBe(1.0);
+});
+
+// While 0 and -0 work the same in most cases, we want to avoid displaying -0 since the concept of -0 can be confusing for some people
+test('Rounding to -0', () => {
+    var roundedNegZero = math.roundMoney(-0);
+    expect(Object.is(roundedNegZero, 0)).toBe(true);
+});
+
+// This is a more of a pedantic than practical case since values representing money should not reach anywhere close to Number.MAX_VALUE
+// However, rounding Number.MAX_VALUE to two decimal places should not cause it to become infinity
+/*
+test('Rounding a very large number', () => {
+    expect(math.roundMoney(Number.MAX_VALUE)).toBeCloseTo(Number.MAX_VALUE);
+});
+*/
+
+test('Rounding infinity', () => {
+    expect(math.roundMoney(Infinity)).toBe(Infinity);
+});
+
+test('Rounding NaN', () => {
+    var roundedNan = math.roundMoney(NaN);
+    expect(isNaN(roundedNan)).toBe(true);
+});
+


src/helpers/math.test.js
```diff
@@ -0,0 +1,33 @@
...
+test('Rounding a very large number', () => {
+    expect(math.roundMoney(Number.MAX_VALUE)).toBeCloseTo(Number.MAX_VALUE);
+});
+*/
+
+test('Rounding infinity', () => {
+    expect(math.roundMoney(Infinity)).toBe(Infinity);
+});
+
+test('Rounding NaN', () => {
```
knod: This is actually a good question to ask - do we want rounding money to return NaN? A lot of times it could be convenient because some values start out as `undefined` and returning `0` instead of `NaN` could be a useful shortcut for getting monetary values. On the other hand it could easily fail silently if an actual issue came up. Would it be useful to return 0 if `undefined` and `NaN` if the value is a non-number in some other way?

Maybe it'd be more useful to find an alternative solution that handles that. I just know it's been a bug-prone annoyance that I've been seeing a lot and it would be great to have a uniform solution.

yhou8:
There are several conventions for where to place tests. One convention I've seen was to place unit tests next to where their functions are defined and reserve src/tests for integration tests, which test larger parts of a system. The documentation for the testing framework, Jest, suggests putting unit tests for <file>.js in <file>.test.js, like the App.test.js that was created with create-react-app. I think this makes it simpler to import the functions to test. We can define our own structure for where to place tests as long as we're consistent and document it.

While we don't need roundMoney to be perfectly accurate right now, I feel like it is a good idea to figure out whether floating point arithmetic is consistent with the requirements for calculating program eligibility and benefits early on. Accuracy issues can be hard to debug later on, and Javascript makes it especially difficult to consistently change our representation of money.

The NaN issue is a great point. I forgot that undefined can be coerced to NaN. I prefer if roundMoney treats null, undefined, NaN distinctly to help with debugging or ideally filter them out before using roundMoney since they do not really represent money. Undefined might be coerced to NaN before it reaches roundMoney, which is case that roundMoney cannot detect.

knod:
Thank you for some really useful perspectives. They clarify a lot and bring up good questions. tl;dr the PR looks good to me and the various questions we've brought up warrant continued discussion in their own issues.

The issue of where to place tests is an interesting one. Thanks for really digging into Jest and the conventions surrounding it. I think the decision will partly depend on how we want to develop this project, but I'm up for trying out the structure you've developed. We'll try it for a while and discuss it if something becomes an issue.

I agree about arithmetic accuracy. It would be better to discuss it sooner rather than later. I have more questions, so look out for that issue being added.

I like the idea of `null`, `undefined`, and `NaN` being identified separately. This one also deserves its own issue with further questions. The puzzle of how to handle `undefined` form values also needs addressing, but I agree it could and should be handled independently from and prior to `roundMoney()`. We'll need an issue for that too.

organiker:
Hi, I had a chance to look at the tests yesterday for a bit- 
For the discussion of `roundMoney()` failing 2 tests, it passes if you use `.ceil()` and `.floor()` instead of `.round()`, so depending on how critical it is to be able to round up and down, we might have to separate the `roundMoney()` method into two.
Depending on where things are with testing, I'd be happy to look at separating some of the math functions this coming Tuesday with @yhou8 or anyone else who's interested.

PR #72, opened 2017-10-09
Fixes #70, fixes #64, normalize property name to 'name'

### LINKED ISSUE #64: Entering 'rent' on 'previous expenses' raises an error
To reproduce:

 * Open the demo in a browser
 * Hit "intake new client"
 * Hit "submit"
 * Hit "create visit"
 * Hit "next" until on the "previous expenses" screen
 * Select monthly rent input field
 * Enter '10' 

Expected behavior:

 * Input should appear as typed, and be validated as appropriate

Actual behavior:

 * Screen is replaced with "TypeError: inputProps.id is undefined", with further backtraces

### LINKED ISSUE #70: Normalize the attribute used to store the property name of a form element
- [ ] Turn all instances of these to use the 'name' attribute instead, including the functions that use those attributes:
  - [ ] data-key
  - [ ] id

knod:


knod:
Source of this PR: https://github.com/knod/cliff-effects-1/tree/name-uniformity. Review needed.

knod:
One thing I'm concerned about is that the attribute `name` is used in html in a functional way. The particular use I'm concerned about is with checkboxes. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox (look for 'Handling multiple checkboxes'). It's possible that isn't going to be an issue at all as long as we don't actually submit form data and just send our own stuff to the backend. I'm not sure if there are other similar issues that need to be discussed, so if anyone knows of others, please bring them up.

Issue #16, opened 2017-09-06
ADD:  Page for user to review/confirm client information


lianilychee:
* [x] Mod status bar to account for the added page
* [x] Build the new page

STATUS BAR MOCKUP
![image](https://user-images.githubusercontent.com/6895034/30131486-ed217490-931a-11e7-8710-0c2fc1476d1e.png)

NEW PAGE SPEC

The new page should go after the MassHealth Qualifying Conditions Page, and before the Results page.  This allows the user to review/confirm inputs with the client, before jumping to the data viz.

![image](https://user-images.githubusercontent.com/6895034/30132332-01647aee-931e-11e7-99bd-d5fd1c09da79.png)

- Don't forget commas for household annual income value
- Copy for MassHealth negative:  "No qualifying conditions"

knod:
Not sure at what point we'll get to it, but would it be useful for the values to be changeable in place? Or for each item to have a link to it's step in the form?

carpeliam:
@knod thanks for the ideas! We can give it some thought. I think though that we can keep this issue scoped to not include those details.

knod:
I completely agree, you got my intention exactly. As you said, not anything to require in this issue.

shp54:
https://github.com/codeforboston/cliff-effects/pull/73

knod:
Just to understand more clearly, is this part of the MVP, or a piece that needs some work separate from that?

lianilychee:
The latter.  Jonathan and I were starting to familiarize ourselves with the codebase by watching Shameek tackle a simple issue.

knod:
I'm not sure this one is closed yet, as this page isn't fully functional. To close this issue, we either get rid of it, or complete it.

knod:
Page was removed.

PR #73, opened 2017-10-11
First crack at ConfirmInformation component

shp54:
Fixes issue #16 , needs refactored and broken out, and the display is not fixed, but http://i0.kym-cdn.com/entries/icons/original/000/005/600/its-something.jpg

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "chart.js": ">=2.6.x < 2.7.x",
     "gh-pages": "^1.0.0",
     "lodash": "^4.17.4",
+    "moment": "^2.18.0",
     "react": "^15.6.1",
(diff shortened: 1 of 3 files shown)


shp54:
ish16

knod:
Looks like a great start! A couple of notes:

1. Why was moment.js added? I didn't see it used.
2. Some of the properties there are a bit defunct, and some of them are not worked out yet. We should go over the properties listed in https://docs.google.com/spreadsheets/d/1hvMqhOtIaQr6uYyc-nizXLp8cNtKuajuEeJj6Ysbw08/edit#gid=0. One question we still need to answer is what to count as 'annual income'.

shp54:
Moment.js is a dependency of one of the charting libraries, and in order
for the app to build I needed that specific version of moment.js defined.

PR #77, opened 2017-10-11
Get rid of nested objects in props

shp54:
Props for a React component should always be flat - there are no extra props inside objects. When passing down props from a parent component, the spread notation can be used or the specific props explicitly passed down. I also prefer using the spread notation to other ways to merge, but that's open for discussion.

### DIFF
--- a/src/forms/citizenship.js
+++ b/src/forms/citizenship.js
@@ -18,8 +18,7 @@ import {
 */
 const CitizenshipContent = (props) => {
 
-	var origin = props.origin,
-      client = origin.pageState;
+	var client = props.pageState;
 
   return (
     <wrapper className={'field-aligner'}>
@@ -29,7 +28,7 @@ const CitizenshipContent = (props) => {
           name='citizenshipStatus'
           value='citizen'
           checked={client.citizenshipStatus === 'citizen'}
-          onChange={origin.storeComplex}
+          onChange={props.storeComplex}
         />
       </Form.Field>
       <br/>
@@ -39,7 +38,7 @@ const CitizenshipContent = (props) => {
           name='citizenshipStatus'
           value='immigrant'
           checked={client.citizenshipStatus === 'immigrant'}
-          onChange={origin.storeComplex}
+          onChange={props.storeComplex}
         />
       </Form.Field>
       <br/>
@@ -49,7 +48,7 @@ const CitizenshipContent = (props) => {
           name='citizenshipStatus'
           value='unknown'
           checked={client.citizenshipStatus === 'unknown'}
-          onChange={origin.storeComplex}
+          onChange={props.storeComplex}
         />
       </Form.Field>
     </wrapper>
@@ -75,7 +74,7 @@ const CitizenshipStep = function ( props ) {
         clarifier = {'Select your citizenship status.'}
         left      = {{name: 'Previous', func: props.previousStep}}
         right     = {{name: 'Next', func: props.nextStep}}>
-          <CitizenshipContent origin={props}/>
+          <CitizenshipContent {...props} />
       </FormPartsContainer>
     </Form>
   );
--- a/src/forms/citizenship.js
+++ b/src/forms/citizenship.js
@@ -1,5 +1,5 @@
 // REACT COMPONENTS
-import React, { Component } from 'react';
+import React from 'react';
 import { Form, Radio } from 'semantic-ui-react';
 
 // PROJECT COMPONENTS
--- a/src/forms/citizenship.js
+++ b/src/forms/citizenship.js
@@ -18,8 +18,6 @@ import {
 */
 const CitizenshipContent = (props) => {
 
-	var client = props.pageState;
-
   return (
     <wrapper className={'field-aligner'}>
       <Form.Field>
@@ -27,7 +25,7 @@ const CitizenshipContent = (props) => {
           label='US Citizen / National'
           name='citizenshipStatus'
           value='citizen'
-          checked={client.citizenshipStatus === 'citizen'}
+          checked={props.client.citizenshipStatus === 'citizen'}
           onChange={props.storeComplex}
         />
       </Form.Field>
@@ -37,7 +35,7 @@ const CitizenshipContent = (props) => {
           label='Lawfully present immigrant / AWSS'
           name='citizenshipStatus'
           value='immigrant'
-          checked={client.citizenshipStatus === 'immigrant'}
+          checked={props.client.citizenshipStatus === 'immigrant'}
           onChange={props.storeComplex}
         />
       </Form.Field>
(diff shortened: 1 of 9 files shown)


src/forms/current-benefits.js
```diff
@@ -20,27 +20,26 @@ import {
 */
 const CurrentBenefitsContent = (props) => {
 
-	var origin = props.origin,
-      client = origin.pageState;
+  var client = props.pageState;
```
knod: You instantiate `client` here, but doesn't seem to get used in the following lines. I personally do like the idea of this declaration up here to avoid `props.client` in all the other spots, so I'd go with that. Also, does `props` have both a `client` and a `pageState`?

src/forms/currentIncome.js
```diff
@@ -56,15 +56,15 @@ const IncomeForm = function ( props ) {
 
       <div style={{ textAlign: 'center' }}>
         <Header as='h4' textAlign='center'>
-          FOR A HOUSEHOLD SIZE OF <strong>{ otherProps.pageState.householdSize }</strong>:
+          FOR A HOUSEHOLD SIZE OF <strong>{ props.pageState.householdSize }</strong>:
```
knod: There's some inconsistency with `props.pageState` vs. `props.client` vs. `client = props.pageState`, etc. Is there a specific purpose behind it? If not, could it be made more consistent? We can possibly do that in a future iteration, though, if there isn't time for it now.

src/forms/citizenship.js
```diff
@@ -1,5 +1,5 @@
 // REACT COMPONENTS
-import React, { Component } from 'react';
+import React from 'react';
```
knod: It may be a good idea to add a note about this in the commit message, since removing excess imports is something different than removing nested props. If someone needs to go back to hunt for when this was done, it'd be helpful to be able to see it at least one commit message.

src/forms/previousExpenses.js
```diff
@@ -101,11 +95,11 @@ const Housing = function ( props ) {
       <FormHeading>Shelter</FormHeading>
 
       <MassiveToggle name={ time + 'Homeless' } value={ wasHomeless } storeChecked={ storeChecked }
-          label={'Was the household homeless at the last benefit assessment?'}/>
+          label='Was the household homeless at the last benefit assessment?' />
```
knod: This is more of a question of whether we want to keep things in brackets or not. There are a at least two reasons I can see for keeping things in brackets. First is consistency, second is flexibility. Consistency makes it easier to just automatically do the same thing over and over again. The code writer doesn't have to stop and remember which way they're supposed to do it. It also allows a reader to not stop to question the methodology being used. As far as flexibility goes, with brackets it's easy to add things like `someVar +` when needed without any other maneuvering. Maybe a good topic for discussion.

src/forms/previousExpenses.js
```diff
@@ -341,7 +330,7 @@ const ExpensesFormContent = function ( props ) {
 * 
 * @returns Component
 */
-// `props` is a cloned version of the original props. References broken.
+// `props` is a cloned version of the propsal props. References broken.
```
knod: Auto-replace all, needs correction ;)

src/forms/previousIncome.js
```diff
@@ -50,49 +45,48 @@ const IncomeForm = function ( props ) {
...
-    origin.storeComplex( evnt, inputProps );
+    props.storeComplex( evnt, inputProps );
 
   };  // End ensureCurrent()
 
 
   var sharedProps = { client: client, type: type, time: time,
                       storeComplex: ensureCurrent };
-
+
```
knod: The additional whitespace isn't terrible, but at some point we'll have to go back and fix it.

knod:
Thanks @shp54, some good changes! Check the review comments, of course, but my last question is whether or not you were going to continue the treatment for '/forms/formHelpers.js' - is that planned for this PR?

src/forms/current-benefits.js
shp54: I probably just forgot to remove that. `props` doesn't have have a `pageState` anymore, so I can probably just declare `client = props.client`

shp54: Or we can use ES6 destructuring like so:

`const CurrentBenefitsContent = ({ client, storeChecked })`

src/forms/currentIncome.js
shp54: `props.pageState` should never exist. All interaction with that should be passed through a `client` prop on the component.

src/forms/previousExpenses.js
shp54: Done.

src/forms/currentIncome.js
knod: I mean that instead of having `props.pageState` *and* `props.client` *and* etc. we should just have one way of doing this. Maybe just send it in as 'client' from the very start. I just see the object being used here (in this PR) in different ways and I propose that it should be made more consistent.

src/forms/previousExpenses.js
knod: To clarify, I'm not sure this PR is the place to do this and it may need further discussion before we go after this change.

src/forms/currentIncome.js
shp54: You're right. I think `props.pageState` shouldn't exist. Should be `props.client`. I forgot to change that. My mistake.

shp54:
Added destructuring to make props explicit.

src/forms/currentIncome.js
```diff
@@ -93,7 +92,7 @@ const CurrentIncomeStep = function ( props ) {
         clarifier = 'How much money does your household make now?'
         left      = {{name: 'Previous', func: props.previousStep}}
         right     = {{name: 'Next', func: props.nextStep}}>
-          <IncomeForm client={props.pageState} props={props}/>
+          <IncomeForm prevStep={props.prevStep} nextStep={props.nextStep} storeComplex={props.storeComplex} storeChecked={props.storeChecked} client={props.pageState} />
```
knod: Could we check where `prevStep` and `nextStep` are being used inside `<IncomeForm>`?

src/forms/formHelpers.js
```diff
@@ -325,26 +322,26 @@ const ColumnHeading = function ( props ) {
 * 
 * @function
 * @param {object} props
-* @property {object} props.__ - explanation
+* @property {object} __ - explanation
```
knod: Just because it's destructured doesn't mean it's not a property of `props`, so this should probably stay the same.

src/forms/formHelpers.js
```diff
@@ -142,9 +141,7 @@ const FormPartsContainer = function(props) {
 * 
 * @returns Component
 */
-const MassiveToggle = function ( props ) {
-
-  var time = props.time || '';
+const MassiveToggle = function (props) {
```
knod: Is there a specific reason not to destructure this one?

shp54: It uses `props.onChange || props.storeChecked`, which means that certain objects would fail destructuring if they passed the wrong parameter. This can be destructured after unifying the names.

knod: Sounds good.

src/forms/currentIncome.js
knod: We'll check on this after merging.

knod:
Just a note that this PR also removes old imports.

PR #87, opened 2017-10-13
Remove useless imports

shp54:


knod:
All form pages working

PR #91, opened 2017-10-15
Timeframe swap fixes #83

### LINKED ISSUE #83: Switch from 'previous' and 'current' to 'current' and 'future'
After a discussion with our client, it looks like there's been a mix-up with time frames, so we need to switch those two up, both in the text of the forms and in the property names. It seems desired for the mvp.
- [x] Switch form text
- [x] Switch property names

knod:


src/forms/previousExpenses.js
```diff
@@ -47,7 +49,8 @@ const Housing = function ({ client, type, time, storeComplex, storeChecked }) {
   /** Makes sure values are propagated to 'current' properties if needed */
   let ensureCurrChecked = function ( evnt, inputProps ) {
 
-	let keyOfCurr = inputProps.name.replace( 'previous', 'current' );
+  	let keyOfCurr = inputProps.name.replace( 'previous', 'current' );
```
nicklee100: IMO the use of whitespace is used more effectively in the function ensureCurrComplex on line 36 than here. Either way it is not consistent.

src/forms/previousExpenses.js
```diff
@@ -336,7 +277,8 @@ const PreviousExpensesStep = function ( props ) {
         clarifier = {'Monthly expenses that were expected to happen during the 12 months following that assessment'}
         left      = {{name: 'Previous', func: props.previousStep}}
         right     = {{name: 'Next', func: props.nextStep}}>
-        <ExpensesFormContent {...props} client={props.pageState} />
+        <ExpensesFormContent {...props} client={props.pageState}
+      />
```
nicklee100: I think what you mean to do is move the  ' > ' char on line 279 dow on its own line and keep the '/>' on 281 on line 280.

src/forms/currentExpenses.js
```diff
@@ -49,7 +49,7 @@ const Housing = function ({ client, type, time, storeComplex, storeChecked }) {
   /** Makes sure values are propagated to 'current' properties if needed */
   let ensureCurrChecked = function ( evnt, inputProps ) {
 
-  	let keyOfCurr = inputProps.name.replace( 'previous', 'current' );
+  	let keyOfCurr = inputProps.name.replace( 'current', 'future' );
```
nicklee100: use of whitespace is not consistent line 52 vs 38.

src/forms/currentIncome.js
```diff
@@ -60,8 +59,11 @@ const IncomeForm = function ({ storeComplex, client }) {
   };  // End ensureFuture()
 
 
-  var sharedProps = { client: client, type: type, time: time,
-                      storeComplex: ensureFuture };
+  var sharedProps = {
```
nicklee100: es6 has property value short hand   :) 

```
let sharedProps = {
   client: client,
   time: time,
   type: type
}
```
Do the same thing 

```
let sharedProps = {
   client,
   time,
   type
}
```
cool

not required but I suggest moving 'type: type' on its own line.

src/forms/futureIncome.js
```diff
@@ -87,7 +87,7 @@ const FutureIncomeStep = function ( props ) {
     <Form className = 'income-form'>
       <FormPartsContainer
         title     = 'Future Household Income'
-        clarifier = 'How much money do you think your household could make in the future?'
+        clarifier = 'How much money would your household make in the future?'
```
nicklee100: How come 'would' and not 'will' ?

src/forms/previousExpenses.js
knod: Fixed to match 36.

src/forms/previousExpenses.js
knod: No, I meant to move them both. The `/` lets you know that the node is self-closing.

src/forms/currentIncome.js
knod: Which ES6 features to use is definitely a discussion we should have. For now, I'll just put them on separate lines.

src/forms/futureIncome.js
knod: The discussion with Krisitin seemed to indicate that they would enter projected values so they could experiment to see what the effects would be on their benefits if certain factors of their situations changed.

nicklee100: okay that makes sense.

PR #92, opened 2017-10-15
Cleanup housing.js

knod:


knod:
The 'required fields' issue may need some discussion. We're debating how to be clear about what properties are being used in a calculation, and that could be one way of doing it. They wouldn't be feedback about which fields are required, but which *properties* are required and what type they need to be.

knod:
There's a helper function that helps check properties, so we can can use that in future if needed. In the meantime, other stuff is cleaned up.

PR #94, opened 2017-10-16
removed unesscary coding testing that was testing hard data object in house.js, brought b…

### LINKED ISSUE #93: Result.js has unnecessarily code - needs to be simplified
The point of Result.js is unclear due to unnecessarily code
Should be using `this` keyword in a class function 
Test functions outside of class are unnecessarily as they are testing hardcoded values

nicklee100:
…ack use of this statement, condensed comments at top  closes #93

### DIFF
--- a/src/helpers/Result.js
+++ b/src/helpers/Result.js
@@ -1,126 +1,27 @@
-
-/** 
-* All benefit programs' return value should be an instance of
-* this class.
-* 
-* @todo Implement `.expirationDate` or some interface for that so we can
-* keep track of when data needs to be updated. We can create some kind of
-* flag or alert in here.
-* @todo Figure out how to access this jsdoc definition externally.
-* 
+/**
+* All benefit programs' return value should be an instance of Result class.
+*
+* @todo Implement `.expirationDate`to track  when data needs to be update
 * @external
-* 
 * @class
-* @param {object} trial - Data to try out/validate
+* @param {object}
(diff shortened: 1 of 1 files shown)


src/helpers/Result.js
```diff
@@ -1,126 +1,27 @@
...
-* 
-* @todo Implement `.expirationDate` or some interface for that so we can
-* keep track of when data needs to be updated. We can create some kind of
-* flag or alert in here.
-* @todo Figure out how to access this jsdoc definition externally.
-* 
+/**
+* All benefit programs' return value should be an instance of Result class.
+*
+* @todo Implement `.expirationDate`to track  when data needs to be update
```
knod: Fix typo at end to 'updated'

knod:
Just to sum up a conversation we had elsewhere:

1. Unnecessary testing - the tests can later be replaced by type checking if we implement that, but for now this is what we're doing to make sure every calculation function puts out the kind of data that we need. It did look like the object was only being used in one place, but that's only because we currently have only one functioning calculation.

2. `this` - I think our policy regarding `this` needs to be discussed as a group before making those changes

3. I think the main concern here has been with clarity of the code. I completely agree that the code needs to be more clear, but I think there are other ways to achieve that clarity. We'll try out some different options and see if we can find something that fits.

knod:
Some alternatives: https://gist.github.com/knod/37452cbfef13d4b017300e837772fd6a

nicklee100:
I'm in favor of the gist above.

Issue #60, opened 2017-09-27
Allow deployment selectively to multiple locations


lianilychee:
open research Q.

WillMcIntosh:
## Notes from what we found last night
* It looks like we can change the "deploy" script in package.json to run `node my-deploy-file.js`
    * We wanted "deploy" to get the user's github repo and deploy there
* We talked about creating a "deploy-production" script that deploys the way we're deploying now, to ghpages
* There is also a node package that may be useful called [exp-deploy](https://www.npmjs.com/package/exp-deploy) which handles different deployment tasks such as staging, production, test

knod:
Preventing people from accidentally deploying to the production site is a good thought to start with. We may want to make a second issue (or adjust this issue) to cover being able to deploy to multiple sites. Either before or after the first prototype is done, we'll need a way to deploy to a live site and a way to deploy to a dev site, so it's going to have to be done one way or another.

What follows arose from a discussion between Will and myself. If we create a new issue, I'll move this comment to that issue.

So, my best guess at the different stages of complexity, from most 'simple' to more complex:
1. Require a confirmation before deploying. This assumes we're deploying to just one site.
2. (Maybe almost as 'simple' as 1) Ability to deploy to either live site or to dev site.
3. (A very short step from 2) Ability to deploy to live site, dev site, and your own gh-pages branch. If devs could deploy to their own gh-pages branch, they could easily show anyone whatever they're experimenting with. 
4. A way to customize where you're deploying to in general. This can be more or less complex depending on the kind of interaction you want to provide.

[Edit: after taking a closer think about what info we have and the complexity/simplicity of what we know, numbers 2-4 might be basically the same exact thing and be pretty darn simple to boot. Still don't know anything about 1. More research must be done.]

I think we might as well go for 2 or 3 because we've already found some solid leads on how to do those and haven't found ones for 1 yet. Here's a description of a behavior that can work for 2 (if you take off the last item) and 3 (if you keep the last item):
Assume `npm run` a the start of all of these:
- `deploy --dev` deploys to the dev site
- `deploy --live` deploys to the live site
- just plain `deploy` and/or `deploy --home` (or `--origin` or something) deploys to the gh-pages of the remote repo that this local repo calls 'origin'.

These might be useful:
https://docs.npmjs.com/cli/run-script
https://github.com/thlorenz/resolve-git-remote/blob/master/index.js
https://github.com/wilmoore/node-git-origin-url/blob/master/index.js

[...]

PR #96, opened 2017-10-16
Deploy scripts

WillMcIntosh:
Addresses issue #60 
I did this a few ways before settling on this way, which is by far the simplest. The node package gh-pages will default publish to remote/origin, which will be the user's fork if you forked the codeforboston repo as the readme instructions say to. Running `npm run deploy` will deploy to your forked repo, creating a gh-pages branch if you don't already have one. 

If you run `npm run deploy-dev` it will (theoretically) deploy to Code For Boston's gh-page just as `npm run deploy` did before. I haven't tried it because I didn't want to accidentally deploy without someone checking my code first. 

I can use a different approach if we want this to be more secure. This currently still doesn't prevent someone from just cloning the cliff-effects repo and deploying straight to it, but I think you have to be a contributor to do that if I'm not mistaken.


WillMcIntosh:
This doesn't seem to be working the way I thought it was after further testing with @knod. I'll push updates to my branch after troubleshooting more.

package.json
```diff
@@ -22,7 +22,8 @@
     "build": "react-scripts build",
     "test": "react-scripts test --env=jsdom",
     "eject": "react-scripts eject",
-    "deploy": "npm run build&&gh-pages -d build"
+    "deploy-dev": "npm run build && gh-pages -d build",
```
knod: Tested, but not yet working. Some issues have come up with deployment that we're trying to track down.

WillMcIntosh:
This actually seems to be a non issue... @danaher2 and I just tested it and if you fork the code for boston repo like the readme says, `npm run deploy` actually deploys to your personal gh-pages branch. You have to have a clone of the actual Code For Boston repo to deploy to the production site. Should we close this issue?

WillMcIntosh:
Closed after discussion last night

knod:
Just a reminder that if we ever do both a 'live'/'production' site and a 'dev' site, we'll need the code from this worked out. Also, it would be good to be able to deploy from our own forks, but that's a matter of convenience.

Issue #65, opened 2017-10-04
When using default household size of 1 number of dependents is not bound.


nicklee100:
To reproduce:

Open the demo in a browser

- Hit "intake new client"
- Hit "submit"
- Hit "create visit"
- Hit "next" until on the "Household Size" screen
- Leave the default value 1 in 'Number of members in the household'
- Increase value of number of 'Number of dependents'  and/or 'Number of children under 12' 

Expected behavior

- 'Number of dependents' and/or 'number of children under 12' should be limited by number of household size,   value should stay have 0

Actual behavior:

- values increase past actual house old value

PR #97, opened 2017-10-17
Fixes issue where number of dependents was not limited when household number was not specified.

shp54:


shp54:
Currently limits dependents to 0 when household size is not specified.

knod:
Fixes #65

PR #106, opened 2017-10-18
Nest client state

shp54:
I put some of the properties of the state object into a nested client object, in order to make state more readable (we now have an idea of what bits of state belong to the client) and easier to reason about (eligibility functions are called with `getSnapEligibility(this.state.client)` rather than `getSnapEligibility(this.state)` so they only have properties they need.) In addition, I also renamed `storeComplex` and `storeChecked`, and refactored them into a single method, because they do almost the same work except for one difference.


src/visitPage.js
```diff
@@ -89,37 +91,12 @@ class VisitPage extends Component {
...
-    truth.setState(
-      { [name]: value },
-      function () {
-        // console.log( truth );
-        if ( callback ) { callback( truth ); }
-      }  // This is given no arguments
-    );
+  setClientProperty = (e, data) => {
+    let propertyName = data.name
+    let value = data.value || data.checked  //This handles both complex values and checked values
```
knod: I believe this should be done the other way around. I know currently our check boxes don't have `value`s, but I believe that they can actually have `value`s, while other inputs shouldn't have a `checked` attribute. Unless someone else knows otherwise, we should check for `data.checked` first.

shp54: Good point. Will change.

shp54:
Deleting `props.pageState` from resultsGraph

shp54:
Updated with the changes @knod requested

src/visitPage.js
```diff
@@ -89,37 +91,12 @@ class VisitPage extends Component {
...
-    truth.setState(
-      { [name]: value },
-      function () {
-        // console.log( truth );
-        if ( callback ) { callback( truth ); }
-      }  // This is given no arguments
-    );
+  setClientProperty = (e, data) => {
+    let propertyName = data.name
+    let value = (data.checked != null && data.checked != undefined) ? data.checked : data.value  //This handles both complex values and checked values
```
knod: If you're going to have conditionals, see if `!==` will work. I'm not sure why `checked` would ever be 0, though. That would be very strange, and wrong, but strange and wrong things do happen.

shp54: checked could be `false`, which means the value `false` exists and should be stored. I'm changing `!=` to `!==`, though.

knod: Wow, brain must have been out to lunch. That's a great point. In that case though, would you not simply check that the type of `checked` is a boolean?

src/visitPage.js
```diff
@@ -95,7 +95,7 @@ class VisitPage extends Component {
 
   setClientProperty = (e, data) => {
     let propertyName = data.name
-    let value = (data.checked != null && data.checked != undefined) ? data.checked : data.value  //This handles both complex values and checked values
+    let value = (data.checked !== null && data.checked !== undefined) ? data.checked : data.value  //This handles both complex values and checked values
```
knod: You made an excellent point and I have no idea what I was thinking. With that in mind, though, maybe it'd be best to just check for a boolean.

knod:
Note that this gets rid of the callbacks that we built in for setting state.

PR #107, opened 2017-10-18
Simplifying steps array

shp54:
There's no need to specify default `completed` and `active` properties, because they're recalculated on every render.

### DIFF
--- a/src/visitPage.js
+++ b/src/visitPage.js
@@ -77,17 +77,17 @@ class VisitPage extends Component {
     };  // end this.state {}
 
     this.steps = [
-      { completed: false, active: false, title: 'Current Benefits', form: CurrentBenefitsStep, },
-      { completed: false, active: false, title: 'Household Size', form: HouseholdSizeStep },
-      { completed: false, active: false, title: 'Current Income', form: CurrentIncomeStep },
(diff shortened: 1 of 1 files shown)


src/visitPage.js
```diff
@@ -77,17 +77,17 @@ class VisitPage extends Component {
...
-      { completed: false, active: false, title: 'Current Income', form: CurrentIncomeStep },
-      { completed: false, active: false, title: 'Current Expenses', form: CurrentExpensesStep },
-      { completed: false, active: false, title: 'Future Income', form: FutureIncomeStep },
-      { completed: false, active: false, title: 'Citizenship', form: CitizenshipStep },
-      { completed: false, active: false, title: 'MassHealth', form: HealthStep },
-      { completed: false, active: false, title: 'Confirm Information', form: ConfirmInformation  },
-      // { completed: false, active: false, title: 'SNAP', form: SNAPStep },
-      // { completed: false, active: false, title: 'Housing', form: HousingStep },
-      { completed: false, active: false, title: 'Results', form: ResultsGraph }
+      { title: 'Household Size', form: HouseholdSizeStep },
```
knod: Current Benefits should be the first step. They should remain in the order in which they were previously.

knod: Can't check other functionality till that is corrected.

PR #110, opened 2017-10-18
Fixes #104, adds beta/prototype warning

### LINKED ISSUE #104: Add disclaimer to codebase
@lianilychee to sync w/ Andrew on verbiage

knod:
Not sure how noticeable it is. Do we want to change the color? If so, to what?

### DIFF
--- a/src/BetaWarning.js
+++ b/src/BetaWarning.js
@@ -1,0 +1,15 @@
+import React, { Component } from 'react';
+import { Menu } from 'semantic-ui-react';
+
+const BetaWarning = function ( props ) {
+  // The `style` 'fixes' no space in between those two nodes, but whoever works on
+  // this long term might consider a different UI package because that's not great.
+  return (
+    <Menu.Item>
+      <div style={{ display: 'block' }}><strong>WARNING:</strong>{' '}This tool is a Prototype. It should not be used for official purposes.</div>
+    </Menu.Item>
+  );
+};  // End BetaWarning(<>)
+
+
+export { BetaWarning };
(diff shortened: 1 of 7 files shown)


lianilychee:
Could you add a screenshot, please?

knod:
disclaimer

knod:
@lianilychee Does that screenshot help?

lianilychee:
@knod thank you!  Yes, it does.  I would change to:
>This tool does not contain full functionality, and should not be used with actual client data.

... because a demo for more funding could very well be an official purpose.  BUT, I will punt to Andrew on this one.  I will tag Andrew once he's accepted the repo invite.

anhase:
Let's use: "This tool is a prototype. Do not use with actual client data." Strike "WARNING" from the line. Having the language will suffice.

knod:
The start of both of those work for me. My concern is that "actual client data" sounds like the data is going to be stored somewhere. We're trying to say that this doesn't guarantee results, right? Maybe something like this?

> This tool is a prototype. It does not contain full functionality.

knod:
Or
> This tool is a prototype. It is not fully functional.

anhase:
Good point. We're trying to limit expectations and say that we can't guarantee accuracy. My only hesitation about using the word "functionality" is that we might end up with an app that has full functionality for the benefits that we list (Section 8, SNAP). The reason we're providing the caveat is that there are many more benefits to incorporate down the line, so it's not currently a tool that works for everyone. How about, "This tool is a work in progress. It does not calculate cliff effects for all available public benefits for every user"?

knod:
Even if we have full functionality, we can't guarantee accuracy. I'm not sure anyone can. We want it to basically say that a person shouldn't rely on the results they get with this tool, right?

anhase:
Yes

knod:
How about...?
> This tool is a prototype and should not be used to make financial decisions.

anhase:
Yes, works for me!

knod:
disclaimer-2

Issue #109, opened 2017-10-18
Future earned income value is prefilled with incorrectly altered current value


knod:
### Replicate:
1. Go to 'Current Household Income' form page
2. Enter '200' into monthly earned income value
3. Go forward two steps to Future Household Income

### Expected result:
- Monthly income value should be 200.

### Actual Result:
- Monthly income value is 2.

### Environment:
Chrome v61.0.3163.100
Commit: f64d563

knod:
It looks like the issue is a little stranger than that.

### Replicate:
1. Go to 'Current Household Income' form page
2. Enter '5000' into monthly earned income value
3. Go forward two steps to Future Household Income

### Expected result:
- Monthly income value should be 5000.

### Actual Result:
- Monthly income value is 5.

### Environment:
Chrome v61.0.3163.100
Commit: f64d563

PR #113, opened 2017-10-19
Clean up ensureFuture methods and lift logic to setClientProperty method

shp54:
Putting all the property-setting logic in the same place makes things easier to read and reason about. Also, as a bonus, fixes issue #109!

### DIFF
--- a/src/visitPage.js
+++ b/src/visitPage.js
@@ -96,7 +96,14 @@ class VisitPage extends Component {
   setClientProperty = (e, data) => {
     let propertyName = data.name
     let value = typeof(data.checked) === "boolean" ? data.checked : data.value  //This handles both complex values and checked values
-    this.setState(prevState => ({ client: {...prevState.client, [propertyName]: value }}));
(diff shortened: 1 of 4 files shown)


src/visitPage.js
```diff
@@ -96,7 +96,14 @@ class VisitPage extends Component {
   setClientProperty = (e, data) => {
     let propertyName = data.name
     let value = typeof(data.checked) === "boolean" ? data.checked : data.value  //This handles both complex values and checked values
-    this.setState(prevState => ({ client: {...prevState.client, [propertyName]: value }}));
+    
+    let newClientValues = {[propertyName]: value}
+    if(data.fillFuture){ //If fillFuture is true, values will be propagated to both 'current' and 'future' versions
+      let futurePropertyName = propertyName.replace('current', 'future')
+      newClientValues[futurePropertyName] = value
```
knod: Value that the user has entered in cannot be overwritten

PR #130, opened 2017-10-28
updated federal poverty data to yearly

scapp281:


knod:
Note: This is partly because of the way we've seen percents and monthly values calculated based on 100% yearly values. For example, [this doc](http://www.mass.gov/eohhs/docs/masshealth/deskguides/fpl-deskguide.pdf) looks like is does `Math.ceil( yearlyVal/percent )` for different percent values, etc.

PR #137, opened 2017-10-30
Fixes #136 - changes and fixes 'About' page link, points at README

### LINKED ISSUE #136: Change 'About' page target, fix broken link interface
From what was expressed at the last two meetings, it occurs to me that we might want the 'About' page to point to our repo's README instead of at any particular organization's page. Also, somewhere along the way (probably when trying to switch from react-redirect), the 'About' page link got broken. Those can be taken care of in one go.

Like the other similar issues - changing the logo, etc. - I'm going to put this on the MVP milestone. Feel free to remove it if desired.

knod:
See #136. The actual visual appearance is not different.

### DIFF
--- a/src/Beta.js
+++ b/src/Beta.js
@@ -9,13 +9,7 @@ const BetaWarning = function ( props ) {
   );
 };  // End BetaWarning(<>)
 
-const About = function ( props ) {
-  return (
-    <Menu.Item>
-      <a href='https://github.com/codeforboston/cliff-effects#cliff-effects'>About</a>
-    </Menu.Item>
-  );
-};  // End About(<>)
 
-
-export { BetaWarning, About };
+export {
+	BetaWarning
+};
--- a/src/Beta.js
+++ b/src/Beta.js
@@ -9,5 +9,13 @@ const BetaWarning = function ( props ) {
   );
 };  // End BetaWarning(<>)
 
+const About = function ( props ) {
+  return (
+    <Menu.Item>
+      <a href='https://github.com/codeforboston/cliff-effects#cliff-effects'>About</a>
+    </Menu.Item>
+  );
+};  // End About(<>)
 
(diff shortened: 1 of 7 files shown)


src/Beta.js
```diff
@@ -9,5 +9,13 @@ const BetaWarning = function ( props ) {
   );
 };  // End BetaWarning(<>)
 
+const About = function ( props ) {
+  return (
+    <Menu.Item>
+      <a href='https://github.com/codeforboston/cliff-effects#cliff-effects'>About</a>
```
shp54: You can add `target="_blank"` to the `<a>` tag to open in it in a new window. Not a dealbreaker, but worth nothing.

knod: Good point! Though this may be moot now. Still, we should keep this in mind for the future since we have the 'Feedback' button to consider.

lianilychee: As per discussion on [Issue 136](https://github.com/codeforboston/cliff-effects/issues/136) create blank About Page instead of pointing to repo README.  Content will be populated later.  Since About will not point to an external site, we don't need to open in new tab (in other cases, would agree!)

knod:
I've created a file ('AboutContent.js') for editing 'About' page content that's been simplified in order to make it more clear where the editing should happen. To be honest, we can make 'aboutPage.js' into something much more simple. It's a small amount of work, but it would need to be done. For now, is this sufficient?

knod:
@lianilychee, @anhase, does this seem like it'd work for our current purposes?

anhase:
Seems so!

knod:
@lianilychee: I'm going to go ahead and dismiss the 'changes requested' review since the changes were made and I want to merge this before further merge conflicts become issues. If further changes need to be made just let me know. Another note: the Travis Cl test aren't passing because the tests have not yet been properly integrated. All is well.

PR #140, opened 2017-11-01
snap program

scapp281:


src/programs/state/massachusetts/snap.js
```diff
@@ -1,14 +1,443 @@
...
     }
-}
+  }
+  return isPassGrossIncomeTest;
+};
+
+// INCOME DEDUCTIONS
+const getStandardDeduction = function (client, timeframe) {
+  if (client[timeframe + 'HouseholdSize'] >= 6) {
+    return data.standardDeduction[6];
```
shp54: I think you can use `getMaxIntKey()` from `helpers/get-gov-data.js` here (you'll have to change it to export `getMaxIntKey` though.)

src/programs/state/massachusetts/snap.js
```diff
@@ -1,14 +1,443 @@
...
+//   'currentRentShareMonthly',
+//   'currentContractRentMonthly',
+//   'currentRentMonthly'
+//   */
+// ];
+
+const getAllowance = function(client, timeframe, data, baseRate ) {
+  if ( client[timeframe + 'HouseholdSize'] > 8  ) {
+    return data[8] + (client[timeframe + 'HouseholdSize'] - 8) * baseRate;
+  } else {
```
shp54: `getMaxIntKey()` should probably be an exported method, since it seems that `baseRate` changes a lot. @knod I want to hear your thoughts on how to use the helper methods here

knod: @shp54: He can actually just use `getMonthlyLimitBySize()`. It's exactly what it's built for. In fact, it would be perfect for both of these. It's part of why I didn't export `getMaxIntKey()` - the other functions are already taking care of what needs to be done. If we find we have more need of `getMaxIntKey()`, we'll set up a file for more utility-type stuff and export it there (it's not like it really has much to do with getting government data other than being useful to the process).

Issue #148, opened 2017-11-01
Switch 'Result' page to table


knod:
Results page should show table shown by [this mockup, updated 10/31](https://docs.google.com/presentation/d/1hPZAPpPhdwpbn2Pv3LbjyXE3bekn_RG-CS8H0NT8rUM/edit#slide=id.p). It should include every program for which we're doing calculations.

For now KEEP THE GRAPH on that page too, just to aid in development. Also, we may want to consider keeping the graph, though hiding it, and allowing a user to reveal it if they so desire. The graph really does give a good idea of the progression of a program as income changes.

lianilychee:
From [Issue 46](https://github.com/codeforboston/cliff-effects/issues/46)

> @WillMcIntosh: The [react-json-table](https://github.com/arqex/react-json-table) package looks promising. We just need to put our data into json format to be rendered by react. (It's currently not, to my knowledge)

knod:
We did get a table. We're keeping the Results page till we figure out what kind of visualizations the project's clients are looking for.

PR #149, opened 2017-11-01
Addresses #148, results as table

knod:
It doesn't look identical to the mock-up, but is it close enough to what we're looking for? The separate tables thing is because it was the simplest way to achieve the type of definition it looked like you wanted. We can combine them into one table and see if someone can get into the style manipulation side of things. If someone is into that, let me know and we can explore how to get you/us into that with React. It's just going to take a fair amount of fiddling.

table-v1


anhase:
Looks good to me!

PR #157, opened 2017-11-07
Strengthens and clarifies questions about shelter, updates related files

knod:
A lot of our questions have toggles in a way that is confusing. I'm converting some questions to behave slightly differently. Also, the way some of these properties are stored and used can be made more efficient now that we know a little more about how we need to use them.


knod:
Welp, this has gone pear-shaped with the other merges, so I'm redoing it. Keeping this here for now.

knod:
This should be ready again. Whew.

PR #161, opened 2017-11-07
Fixes result table income being monthly instead of yearly

knod:


knod:
Also, the calculations should probably be abstracted to somewhere so they can be tested independently.

Issue #167, opened 2017-11-10
Organize Project Structure


ManickYoj:
Currently, our project structure has some issues.

* Some code paths (maybe just `src/.github`) are non-functional because of their file structure placement.
* Some files are unused or contain dead code.
* We lack consistent casing conventions. This makes it harder to identify the contents of a file.
* We are not following any of the common conventions for the folder structure of React projects. This makes files hard to find, understanding the codebase hard, and adding to the codebase more difficult.

I'd like to remedy this. Here are my proposals from least to most controversial:

* Code paths that don't work should work. That means moving files to the right places to function, eg. `src/.github` to `/.github`.
* We should not have unused files or files containing dead code. If we still need them, that's what branches are for. We should not be accepting PRs with dead code. If we have existing dead code, we should get rid of it unless a person will be working with that dead code in the immediate future.
* Casing should follow React convention. A file containing a React component (it should contain only one) should be named with PascalCase identically to the exported component. Files that are not react specific should use camelCase. More controversially, all React files should be suffixed with `.jsx` to indicate that they are using React's JSX language. This will help identify their file contents to code authors at a glance, and assist text editors in properly syntax highlighting the files.
* We should pick a folder structure and stick to it. I propose the '[one-folder-per-concern](https://daveceddia.com/react-project-structure/)' structure. It is not the most scalable structure out there, but it is likely the simplest to add to and maintain for small scale projects, which we are.

ManickYoj:
I'm going to start hacking away at this unless anyone raises strong objections. Insofar as possible, I'll try to group my change-sets independently so that only the controversial changes need be debated.

knod:
Glad to see this happening! My two cents:
- Dead code: agreed, though I do wonder if some of the dead code is still meant to be reintegrated. Regardless, we can checkout old commits and find whatever we're looking for.
- Casing conventions: Yes please! I also suggest that files that export a single class start with capital letters: SomethingUseful.js. Thoughts?
- .jsx: Not as sure about this one. I've not used it before. What are its potential downsides?
- Folder structure: that link leads to a page that doesn't seem to propose a one-folder-one-concern solution, but it does have an initial folder structure, and some guiding principles, that I like. I think as long as our 'containers' and 'stylers' (or whatever) folders can be broken up into purposes like 'forms'/'formPages' (or would it be 'form-pages'?), etc. then it looks great. Followup questions:
  - Do our 'programs' and 'data' folders stay where they are?
[...]

PR #173, opened 2017-11-10
Delete unused login page

ManickYoj:
We have a `loginPage.js` file. We do not use it. We do not, insofar as I know, plan to use it in the foreseeable lifespan of this project. #nuke-it-from-orbit-unless-someone-disagrees

One piece of #167 project organization issue.


ManickYoj:
Thanks for resolving the conflict @knod ! Should be good to go.

PR #174, opened 2017-11-10
Relocate pages to `src/containers`

ManickYoj:
Previously, all of our pages were located within the root of `/src`. As a piece of the #167 project organization issue, pages should be located in `/src/containers` as they are non-reusable, stateful components (ie. a container, for our purposes).

This PR moves all active pages to `/src/containers`. It may eventually be useful to locate them in a `src/containers/pages` directory, but I don't think we're there yet.

This PR also deletes trailing whitespace and dead code within the pages.


knod:
Not a request for change, but this merge is going to wait on the earlier PR that hasn't yet been approved because there's a potential for causing conflicts there too. Since there are already ones here that will have to be worked out, I'd rather deal with all conflicts on this PR.

lianilychee:
#157 has been merged.

PR #177, opened 2017-11-10
Updates package.json for peer depenency warnings and errors

knod:


knod:
We're going to try this again with an updated version of node.

PR #181, opened 2017-11-11
Setup for new form/client value management

knod:
It has transformers, but I'm open for discussion and I think it's a start.


src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,109 @@
+/** Handles the core of form values, including:
+ * - Storing both a current and future value
```
ManickYoj: Hmm, when we were discussing this, I thought we were talking specifically about the concept of an `Income`, which in our version of the world is always represented with a current and future value. I'm not sure if this concept is as universal when applied to something as a generic as a form value.

If this refers to the difference between the input's current value vs. some previous value, then that seems like a different concept, better handled by our view layer (React) than by our data models.

If we're storing their last, fully valid input before they moved on to another input, maybe that's a case for this. I'm not positive what our objective is here.

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,109 @@
...
+ * @param {function} [makeValid] - returns a valid value no
+ * matter what. Will be given the new value, the current value,
+ * the string 'current' or 'future', and the `FormValue` instance.
+ * 
+ * @returns FormValue
+ * @todo Finish description of FormValue return value
+ */
+class FormValue {
+
+  constructor ( name, val = null, makeValid ) {
```
ManickYoj: If we're going to make one default explicit in the constructor, I would strongly suggest we make all defaults explicit in the constructor: IE if `val=null`, then we should have `makeValid = (newVal) => newVal` rather than mix and match. Otherwise, the next person to read the code may well expect that makeValid has no default because they expect the defaults to be explicit based on `val`.

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,109 @@
...
+
+    var fVal = this;
```
ManickYoj: Why not just use `this` in the following variable assignments?

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,109 @@
...
+
+  }  // End FormValue.constructor()
+
+
+  setCurrent ( val ) {
+
+    var fVal = this;
+
+    // Transformer needed for parseInt and such
+    fVal.current = fVal.makeValid( val, fVal.current, 'current', fVal );
```
ManickYoj: What happens if makeValid cannot transform the value? I suggest we explicitly raise an error. Certainly, wee should not proceed with execution. Also, if makeValid is just casting values I wonder if there's not a better solution. Ordinarily it would be a strongly typed database, but I imagine there are other solutions here.

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,109 @@
...
+    if ( !fVal._futureWasChanged ) {
+      fVal.setFuture( val, true );
+    }
+
+    return fVal;  // Return valid value instead?
+
+  }  // End FormValue.setCurrent()
+
+
+  setFuture ( val, propagating ) {
```
ManickYoj: In what scenario would we want use the propagating flag? Also, what does it semantically mean to set a future value as propagating? Where would it propagate to?

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,109 @@
...
+      fVal._futureWasChanged = !propagating;
+    }
+    fVal.future = fVal.makeValid( val, fVal.future, 'future', fVal );
+    
+    return fVal;  // Return valid value instead?
+
+  }  // End FormValue.setFuture()
+
+
+  refresh ( timeframe ) {
```
ManickYoj: I have no idea what this function is supposed to do. I strongly suggest a doc-string, if we need this function at all.

src/helpers/client/FormValue.js
knod: There are a lot of future/past/present concepts in there, so I'll try to explain the situation. *All* our form values have a theoretical current and projected/imagined future value, just like income. This, again, depends partially on how we want to think of this project moving forward. Theoretically, in future you'd be able to say, 'What if I got pregnant? What would be my benefit situation in that case?'. There would eventually be two stages to the form - one for all current values, and one for all projected/future values. Does that clarify a bit?

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,109 @@
...
+ * All values stored in a FormValue instance should be valid
+ * 
+ * There must be a better way to write that.
+ * 
+ * @class
+ * @param {string} name - the name of the client property
+ * @param {any} [val] - the initial value
+ * @param {function} [makeValid] - returns a valid value no
+ * matter what. Will be given the new value, the current value,
+ * the string 'current' or 'future', and the `FormValue` instance.
```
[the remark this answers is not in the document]
knod: If this is a detailed reference guide for jsdoc, then this is exactly what we need. I've just never worked with jsdoc before. Is that what this is, or is dev docs its own thing?

src/helpers/client/FormValue.js
knod: Fair point. I'd rather put the defaults in the body of the function then, so I'll do that.

src/helpers/client/FormValue.js
knod: Managing `this` is a whole discussion we can have sometime. Let's see if we can clear some distractions this Tuesday and get into it.

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,109 @@
...
+ * @todo Finish description of FormValue return value
+ */
+class FormValue {
+
+  constructor ( name, val = null, makeValid ) {
+
+    var fVal = this;
+    
+    fVal.name = name;
+    fVal._futureWasChanged = false;  // Handled internally
```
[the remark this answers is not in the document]
knod: It's something I'm thinking about. Again, non-explicit getters and setters, and code in general, are things I tend to avoid. The getter solution was very elegant, though. Maybe I need to see it in action with a few of our other, more sticky, cases to see how well it handles there.

src/helpers/client/FormValue.js
knod: That's the point of `makeValid()`. It *always* returns a valid value no matter what. As for what it's doing to transform the value, it's not just casting it. In some cases it's converting a string to a number, or a string to a boolean value. Other times, it's checking against values that it is dependent on. For example: household size is changed to 5, then number of dependents is changed to 4, then household size is changed to 3. Number of dependents now needs to change to something below 3.

knod: Errors in input, as we said, should be dealt with by components. The client object properties should always be valid. Maybe I'm just not seeing some cases, though. Can you give some examples?

src/helpers/client/FormValue.js
knod: This flag is to tell `setFuture()` whether it is being called because of a user change or being called internally by something like `setCurrent()`. It is being *propagated to*.

src/helpers/client/FormValue.js
knod: I agree about the doc-string if this function survives. As I hoped to say in the comments but apparently omitted, when one value relies/is dependent on other values, it may need updating. The name really should have been `update`, but c'est la vie, and it's a use case that came up through experimentation with trying to implement the class. This is actually part of the getter/setter debate. If we did this with a prototypical getter, or if we switched to an explicit getter for that matter, then this function wouldn't need to exist.

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,109 @@
...
+ * @class
+ * @param {string} name - the name of the client property
+ * @param {any} [val] - the initial value
+ * @param {function} [makeValid] - returns a valid value no
+ * matter what. Will be given the new value, the current value,
+ * the string 'current' or 'future', and the `FormValue` instance.
+ * 
+ * @returns FormValue
+ * @todo Finish description of FormValue return value
+ */
```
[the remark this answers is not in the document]
knod: Yeah, not being even passingly familiar with jsdoc has not made this part of the journey super easy.

src/helpers/client/FormValue.js
knod: Ah, I did include a comment, but it was in the commit message.

src/helpers/client/FormValue.js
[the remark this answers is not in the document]
knod: Not done yet, not sure why it's marked as 'outdated'.

src/helpers/client/FormValue.js
knod: Hmm, how do we go about this - transform in both getters and setters?

src/helpers/client/FormValue.js
ManickYoj: Fair enough.

src/helpers/client/FormValue.js
ManickYoj: Sure. Most of the time, people save to var when passing to a callback though - it's just a bit odd when we're not doing anything with it.

src/helpers/client/FormValue.js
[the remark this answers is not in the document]
ManickYoj: I accept that the implicit behavior of propagation could raise a risk of black-boxing the behavior. I think that will be the case with either solution. If you'd strongly prefer the explicit state, I can live with it.

src/helpers/client/FormValue.js
ManickYoj: If it's mostly used for enforcing min/max values, like in your answer, that's pretty reasonable, though we may consider just using an optional `bounds` param. Then we can clamp the value to be within those bounds as you're suggesting.

For eg. string to boolean though, if the valid answers are `yes`/`no` and you write `'bavadf'`, or `'5'`what are we supposed to do with that? What about string to number (probably a more common case) with a bogus input value? I think most of these can be enforced by using the right input type on the front end. Then we're down to if someone were to console in values in which case, I think a validator will do just fine for us since there's no reason for us to try to convert bogus values. In that case, what cases *are* we concerned about?

I guess my concern here is just that it's really hard to identify all the possible edge cases we'd need to convert, but it's relatively simple to reject any except for those that match a very specific format, then let the user re-enter them correctly.

Regardless, we can put it out there and see how it works in action.

src/helpers/client/FormValue.js
ManickYoj: Hmm. Wouldn't it make more sense to flag setCurrent to set the value of `_future` directly? Why explicitly expose the `propogating` flag to the outside world to poke at?

src/helpers/client/FormValue.js
ManickYoj: That helps a lot. Thank you.

I am curious about the *all* though - we have non-temporal form data: when a client enters their name, or when they select that they use SNAP benefits. These are temporally static form values. Is there a more scoped name we could give to `FormValue` to indicate that it only refers to the subset of data we track that has a past and future state?

ManickYoj:
Perhaps predictably, I prefer the implementation of this in #189 which, to me, reads much more clearly and succinctly. I'd prefer to move through and approve that one, if possible.

src/helpers/client/FormValue.js
knod: Good point, though their name isn't one of the questions in the form. Any ideas for a better name?

knod:
We'll have to discuss issues mentioned in the latest comments in #189 before making decisions.

src/helpers/client/FormValue.js
knod: If we go with this option, maybe you could put together some demo code showing what you envision.

knod:
Retiring this considering new issues on the table, though parts of this may be used.

PR #182, opened 2017-11-11
Project folder cleanup: helpers, src root naming, and tests

ManickYoj:
I got a bit carried away here. Really 3 PRs in one related to PR #167 project organization issue.

1. Moves tests to a `src/tests` directory. Would have moved them to a top-level `/tests` dir, but react-scripts forbids it. Not hyped about `src/tests` but it maintains one folder per concern, technically.
2. Moves helpers to the more canonically named `src/utils` directory. Normalizes the file names and deletes unused files. Notably, two of the files that remained were imported only within dead code in another file. If we don't need them, it would be nice to be able to prune these two.
3. Normalizes the names of all of those components that remain (for now) in the `/src` root.

Tests passing and nothing looks amiss on the front end or console.


src/config/snap.js
```diff
@@ -1,3 +1,4 @@
+// REVIEW: This file appears to be currently unused. Candidate for removal
```
knod: It's not currently unused. 'snap.js' uses it. It's just in the wrong place. The data needs to be re-named to be more descriptive, and objects need to be moved to the data folders with some sensible file naming. Then snap.js needs to be updated to use the new locations and names.

src/containers/ClientDetailPage.js
```diff
@@ -15,8 +15,12 @@ import {
 import { Link } from 'react-router-dom';
 
 import { MainMenu } from '../MainMenu';
-import FixedMenu from '../fixedMenu';
-import { clientList } from '../clientList';
+import FixedMenu from '../FixedMenu';
+
+// REVIEW: As we approach MVP, we probably don't want to be
```
knod: Not for this PR (enough going on already), but in one of the next rounds, how about we just move the 'data' for one client into here or something and get rid of 'DummyClients'?

src/programs/state/massachusetts/housing.js
```diff
@@ -3,11 +3,14 @@
 import { CHILD_CARE_EXPENSES } from '../../../data/state/massachusetts/name-cores';
 
 // UTILITIES
-import { toCashflow, sumCashflow, getGrossUnearnedIncomeMonthly } from '../../../helpers/cashflow';
-import { sum } from '../../../helpers/math';
+import { sum } from '../../../utils/math';
+import { Result } from '../../../utils/Result';
```
knod: 1. This comment doesn't matter because we're going to change what calculations return.
2. `Result` seems to be to be in a different category than a utility and I'm not sure this way of writing it clarifies anything about its use in this context.

src/programs/state/massachusetts/snap.js
```diff
@@ -421,13 +425,14 @@ export { getSNAPBenefits };
 
 
 
-
-
-
-
+// REVIEW: Can this dead code be removed?
```
knod: No, the code is actually being changed to use 'getGovData.js', etc. We may end up dong away with 'helperFunctions.js', though. 'getGovData.js' will also be used to replace other functionality in here. We've found it to be a common data pattern.

src/programs/state/massachusetts/template.js
```diff
@@ -17,10 +17,10 @@
 // import { SOME_DATA } from '../../../data/state/massachusetts/2018/SOME_DATA.js';
 
 // UTILITIES
-// import { roundMoney } from '../../../helpers/math';
+// import { roundMoney } from '../../../utils/math';
```
knod: Just fyi, we'll probably be able to remove this after the client form values stuff has been worked out. Then again, this template would do well to be re-worked. For that matter, I've been debating just having a templates folder. Some of the same questions as the location of the test files come up. On the third (fourth?) hand, there may be little point to templates at this juncture unless we're trying to help our possible future collaborators (or explore some kind of volunteer collaboration format/architecture).

ManickYoj:
Alright - I can respond to the line comments independently, but I can't respond to the review comments directly, so I'll get them all here. I just need a firm opinion on each to move forward.

> Test location: from what I'd been told, the convention with Jest is to put tests next to the files that they test. Is that not the case?

Jest doesn't discriminate, so long as tests are either in a dir named `__tests__` or have `.test.js` as a suffix. I thought about both approaches. Because tests are a concern, just the same as styling or markup, I wanted to maintain our directory per concern approach. If you'd strongly prefer tests alongside their subjects. We may do that.

> I do like 'utils' better (it's what I choose for my own projects) and I like the move in general. I'm not sure all of these are 'utils' exactly, though. For example, I wouldn't call the Result class a util. That's more about deciding where those others go, not about choosing the file name, but it is something we should work out.

`Results` had lived in helpers. Helpers and utils are largely synonymous to me. In my work (using rails), we have what is effectively a non-view, non-model related logic folder for helper objects, named `objects` as per the ruby on rails convention. I figure for us, that is `utils`, but it could be generalized to `logic`. `utils` is certainly more canonical.

> 'Beta.js' is meant to contain beta version/prototype stuff, not just the warning. For example, a 'Feedback' button. Now that I think about it, though, the recent discussions about how feedback is going to work imply that there won't be a 'Feedback' button, so maybe that's not the plan we want. Eh, we'll see how it pans out I guess.

If that's the case, I would suggest separate files `BetaWarning` and `BetaFeedback` since each is a unique React component. They're not really related and don't have to appear together either spacially or conceptually, so I don't think we need to couple them. I'm in favor of following convention here: one component per file, with the filename matching the component.

src/config/snap.js
ManickYoj: This is my mistake - you're correct. It certainly is used. Not sure why I left this comment. I did update the import as part of this PR, as seen farther down the page.

I will remove this line comment.

src/containers/ClientDetailPage.js
ManickYoj: Hmm - I'd rather keep it separate: it makes it easier to pull out the code cleanly when we no longer need it, and importing `dummyClients` is a nice red flag that this isn't production ready.

src/programs/state/massachusetts/housing.js
ManickYoj: Let me know what you want done with `Result`. I don't think it's enough of a special snowflake to get special treatment unless it truly represents a different category of logic than other utilities.

src/programs/state/massachusetts/snap.js
ManickYoj: Okay, sounds good. Is there an issue yet for this change? If not, can we create one so that we've got this change marked on the horizon?

src/programs/state/massachusetts/template.js
ManickYoj: Okay - what I'm hearing re: this PR is change nothing. If there's a work item here, I'd suggest populating a separate issue.

knod:
> Jest

Do whatever convention is and, beyond that, whatever path your heart follows!

> `Results`

The name `utils` is fine. I'm saying not everything in there should be in there. We'll figure out how nuke 'Results.js', for example.

> I would suggest separate files `BetaWarning` and `BetaFeedback`

Sounds good!

src/containers/ClientDetailPage.js
knod: Makes sense.

src/programs/state/massachusetts/housing.js
knod: We're gonna nuke it, but it'll be a different issue. This is just part of my journey of self discovery.

src/programs/state/massachusetts/snap.js
knod: It is already a pull request as we speak, I believe!

knod: But we can also make it an issue.

src/programs/state/massachusetts/template.js
knod: Good idea.

src/programs/state/massachusetts/housing.js
ManickYoj: :snowman_with_snow:

Issue #172, opened 2017-11-10
Footer Clean-Up


ManickYoj:
The footer currently looks like this.

![selection_001](https://user-images.githubusercontent.com/2252820/32644059-4272f800-c5ad-11e7-8a15-5a47095d1775.png)

The links under 'About' and 'Services' are duds that do not work. I do not believe we have any intention of making them work. Assuming that is so, let's remove them, and simply leave 'Cliff Effects Tool'.

We probably don't need the footer to take up quite so much space (padding) either, considering its lack of utility. Once the content is changed, make some reasonable decisions to make it look nice.

knod:
Regardless of whether we remove other things from the footer, it occurs to me that we can free up some room at the top by adding the footer to other pages and moving the 'About' link to it. Might also be a good place for the 'Feedback' button mentioned in #133.

knod:
Any opinions on keeping the 'About' page link in the footer?

PR #183, opened 2017-11-12
Work for Issue #172

dtsal3:
Removed the two left-most columns in the footer of the page as well as limited the padding on the footer

I was unsure how people want to proceed with whether or not to put the feedback option in the footer so I didn't touch any of that potential work


ManickYoj:
@dtsal3 - This was a great start but you uncovered the tip of the iceberg here. The footer and header weren't reused between pages, so your changes only hit the home page. You also uncovered that the footer wasn't properly stuck to the bottom of the screen. Starting from your changes, I took care of those and a few more along the way in #191. That's all buttoned up, so you're good to go here. Thanks for kicking this off!

PR #189, opened 2017-11-13
Form values mock-up using prototypicla getters and setters

knod:
...that are able to maintain valid values if the form value is dependent on other form values.


src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,89 @@
...
+ * @returns {FormValue} formVal
+ * @todo Finish description of FormValue return value (tbd)
+ */
+class FormValue {
+
+  constructor ( name, current, makeValid ) {
+
+    // Set defaults
+    if ( current === undefined ) { current = null; }
+    this._makeValid = makeValid || function ( newVal ) { return newVal; };
```
ManickYoj: This default implementation differs from the one you've got in your docstring, and in your function invocations in lines 44, 52, 58, and 66. I think it probably needs to change.

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,89 @@
...
+ * 
+ * All values stored in a FormValue instance should be valid
+ * 
+ * There must be a better way to write that.
+ * @todo Should `FormValue`s contain a reference to their
+ *     client object?
+ * 
+ * @constructor
+ * @param {string} name - the name of the client property
+ * @param {any} [val] - the initial value
```
ManickYoj: `[val]` -> `[current]`

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,89 @@
...
+ * @param {function(*, *, string, FormValue): *} [makeValid] - 
+ *     returns a valid value no matter what. Will be given the
+ *     new value, the current value, the string 'current' or
+ *     'future', and the `FormValue` instance.
+ * 
+ * @returns {FormValue} formVal
+ * @todo Finish description of FormValue return value (tbd)
+ */
```
ManickYoj: Great doc-block.

My only suggestion is that you indicate what the defaults are for those params that have defaults.

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,89 @@
+/** Handles the core of form values, including:
+ * - Storing both a current and future value
+ * - Tracking when the user manipulates a future value
+ * - Propagating current values to future values if they
+ *     haven't been manipulated by the user
+ * - Making this instance serializable
+ * - Making sure only valid values are saved.
```
ManickYoj: :+1: I appreciate this intro to the intent of the class

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,89 @@
...
+ * 
+ * All values stored in a FormValue instance should be valid
+ * 
+ * There must be a better way to write that.
+ * @todo Should `FormValue`s contain a reference to their
+ *     client object?
```
ManickYoj: If this were a database model, the answer would be yes. I think that's effectively what we're simulating, so probably.

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,89 @@
...
+  set current ( value ) {
+    this._current = this._makeValid( value, this._current, 'current', this );
+    return this._current;
+  }
+  
+
+  /** Getter makes sure that if another value affects this
+   *     one, when that value changes this value will update itself. */
+  get current () {
+    this._current = this._makeValid( this._current, this._current, 'current', this );
```
ManickYoj: You shouldn't need to make anything valid with the getters here or in line 66 if the setters are implemented correctly. That's our guarantee, right? Anything saved into the instance must be valid.

src/helpers/client/FormValue.js
```diff
@@ -0,0 +1,89 @@
...
+    // Set and 'set' the starting values
+    this._future = current;
+    this._current = current;
+
+  }  // End FormValue.constructor()
+
+  
+  set current ( value ) {
+    this._current = this._makeValid( value, this._current, 'current', this );
+    return this._current;
```
ManickYoj: `return` values in setters are ignored, I believe (check me though, if you're not sure). Default behavior is that the setter will always return the new value. I would suggest removing this and line 59 for clarity/consistency with most JS setters in the world.

knod: `return` values in setters seem to return the value that was sent, not the value that was eventually set.

knod: Check the first console log here: https://codepen.io/knod/pen/XzRedx?editors=0010

src/helpers/client/FormValue.js
knod: Again, this is about dependent values. Dependent values must update themselves based on other values in client. Either something else needs to set them again, or they need to check themselves when they are being gotten.

knod:
We may have a slightly different question to answer at the moment - React is functional in nature. With values that are dependent on other values, where do the dependent values get their needed information? How do they keep up to date with the most recent version of the client object?

knod:
Retiring this considering new issues on the table, though parts of this may be used.

PR #191, opened 2017-11-14
Footer and Header Rework

### LINKED ISSUE #172: Footer Clean-Up

ManickYoj:
Builds on the changes started in #183 (thanks @dtsal3!) to resolve #172 re: excess components in footer.

#### Context
#183 removed the excess components, but, because our footer was separately implemented for each page, the work was only complete on the home page. It also revealed a separate issue: that the footer was not properly sticking to the bottom of the page, so it was possible to have white-space float underneath the footer on a sufficiently large screen.

#### Header/Footer Rework
- Made footer and header reusable components s.t. they are not duplicated on each page
- Extended removal of excess, unused components from footer to every page
- Made footer stick to bottom of page w/o gap
- Deleted 'simpleMenu' as it was no longer used
- Style tweaks on ClientIntake page to align header with content
- Style tweaks on Visit page to account for removal of custom header - adds page header and back link. Also put content in container to match every other page in the app.


dtsal3:
Awesome! I noticed some of these things from #172 and was working on them but forgot to comment. Thanks for cleaning it up!

Issue #151, opened 2017-11-02
Return sends you back one page


sashamaryl:
﻿

### Images:
None

### Replicate:
1. Place cursor in any entry field on any page. (Does not matter if it has date or not.)
2. Hit return.

### Result I'd Expect:
 - Data would be "entered" and cursor would go to the next field.

### Actual Result:
 - User is returned to previous page. (No data is lost in any field.)

### Environment:
Operating system and version: 2008 macbook running Lubuntu 16.04
Browser type and version: Chromium, 62.0.3202.62 (Official Build)

App version or commit id at the time:
eb8bd79

donblair30:
Thanks Sasha!  I'll plan to take a crack at this.

knod:
I believe the issue is a normal html one. I think it'd be best to tackle this by looking at the html `tab-index` attribute.

knod:
Fixed by #215

PR #198, opened 2017-11-15
return entry behavior -- issue #151

sashamaryl:
stopping the return-jump-back-a-page issue #151 -- correction made line 132 in VisitPage.js

### DIFF
--- a/src/containers/VisitPage.js
+++ b/src/containers/VisitPage.js
@@ -125,20 +125,20 @@ class VisitPage extends Component {
     this.setState(prevState => ({
       currentStep: prevState.currentStep + 1
     }));
-  };
+  }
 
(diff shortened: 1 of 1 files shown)


src/containers/VisitPage.js
```diff
@@ -125,20 +125,20 @@ class VisitPage extends Component {
     this.setState(prevState => ({
       currentStep: prevState.currentStep + 1
     }));
-  };
+  }
 
   previousStep = () => {
     this.setState(prevState => ({
-      currentStep: prevState.currentStep - 1
+      currentStep: prevState.currentStep
```
knod: This breaks the 'Previous' buttons. I think we need to look into the tab-index of the elements instead.

PR #203, opened 2017-11-16
Moves snap data to more logical places

knod:


knod:
Dang, forgot to all-caps the snap data. Will do that as a next step if there's a chance.

Issue #185, opened 2017-11-12
Home page redesign


lianilychee:
Mockup:
![image](https://user-images.githubusercontent.com/6895034/32857458-e0864f76-ca15-11e7-9cd7-f1ffcbaa542f.png)

Actions:
- [x] Hide the header on the home page
- [x] Make [this SVG](https://vectr.com/lianilychee/aFpUTVlyM.svg?width=1920&height=1080&select=aFpUTVlyMpage0) the background image (you will have to right-click save-as the image).  [StackOverflow on how to make a background image in React](https://stackoverflow.com/questions/29322973/whats-the-best-way-to-add-a-full-screen-background-image-in-react-native)
- [x] Change the header + footer color to #00aaa1
- [x] Add title + subtitle.  Text color is #efefef
- [x] Build buttons.  Background color is #ffffffc8.  Text color is the original teal (#00b9b0)
- [x] Add disclaimer.  Text color is #efefef 

This issue is merged with #200 

Defunct as of 11/15
~- [ ] Remove giant CfB logo~
~- [ ] Remove "intake a new client to begin" and associated button~
~- [ ] Merge New Client Intake page content with Home Page~
~- [ ] Remove Client Avatar URL field~
~- [ ] Remove New Client Intake page~
~- [ ] Modify header:  "Home" -> "Intake New Client"~

Defunct as of 11/13
~Landing page needs to be redesigned such that:~
~- Content fills height perfectly, so no scrolling~
~- The awkward whitespace is filled (perhaps with the 'about' content?~
~- The top teal section is not so aggressively large (perhaps change the logo to match the one in left of nav bar, and scale down in size)~

lianilychee:
Thoughts from @sashamaryl 
>remove white bar maybe open up between logo and title
>possibly make whole middle white & reverse out logo and title etc

lianilychee:
Sorry @scapp281 - spec change, just fyi.

knod:
I suggest just getting rid of the header on the home page.

PR #211, opened 2017-11-17
Splash page -- #185

lianilychee:
Pair-programmed with @ManickYoj to change home page to splash page.

Work done:
- removed the header -- this part is a little hacky
- set the svg as full-screen background
- added title text and buttons

@knod question:  The "Get Started" button currently links to the Client Intake pg, which we have decided to bypass.  What, then, am I supposed to put in HomePage.js Line39 to reroute properly?


knod:
@lianilychee: I'll take a look.

knod:
Note that your Travis build is failing, so we may want to check what that's about. Other things that may belong in this PR are removing the ClientIntakePage.js and ClientDetailPage.js, as well as the menu options that link to them on other pages.

knod:
@lianilychee: The fast and dirty way is to change `<Link to="/intake">` to `<Link to="/visit/54321/1">`. Do you want to start off with that?

src/components/Header.js
```diff
@@ -7,7 +10,9 @@ import {
 import { MainMenu } from '../MainMenu';
 import FixedMenu from '../FixedMenu';
 
-const Header = ({fix}) => {
+const Header = ({fix, location: location}) => {
+  console.log(location)
```
knod: Is this console log still needed?

src/components/Header.js
```diff
@@ -7,7 +10,9 @@ import {
...
+  if (location.pathname === '/') return (<div></div>)
```
knod: Where is `Header` being used that this is needed? Also, we should probably not name this `Header` since it has a name conflict with React's `Header`. Never noticed that before.

src/index.css
```diff
@@ -10,13 +10,42 @@ form .field-aligner {
 	max-width: 80%;
 }
 
-#root, #App {
+#App {
   height: 100%;
   display: flex;
```
knod: I don't think this ever got addressed, but we may want to find out which versions of what browsers our beta tests are using. `flex` is relatively new and won't work in some.

lianilychee: IIRC, Andrew said that he would send a notice to the folks we are demoing to, requiring fresh Chrome installs before we demo / test.  So, addressed on a process level :)

src/components/Header.js
lianilychee: Yes, 'tis new.  It's a hacky way to remove the header but only for the home / splash page.  I believe @ManickYoj will be working on a patch.

src/index.css
knod: lol, that's one way to do it!

lianilychee:
Tests pass locally, so should not be a blocker for merge.

@ManickYoj posits that tests may be failing remotely because Travis has not been updated to latest Node, and is happy to fix.

knod:
@lianilychee Do you see the Travis tests failing on github? Have you tried doing `npm install` again since when we checked up the 'household' branch? If you didn't do it again since checking out that branch, your packages are not up to date and you may be getting a false positive.

src/index.css
ManickYoj: Should work on about [85% of the browsers people use](https://caniuse.com/#feat=flexbox). Not ideal, not the end of the world either. I like flexbox because it makes otherwise hacky solutions not hacky, but it being a relatively new feature is a downside.

src/index.css
```diff
@@ -10,13 +10,42 @@ form .field-aligner {
...
+#root {
+  height:  100%;
+}
+
+#HashRouter {
+  flex: 1 0;
+}
+
+#HomePage {
+  background-image: url(./images/splash.svg);
```
knod: Our footer color is #00b6ad, but the decorative arrow color is #01aaa0. I think it may look better if they match.

knod: Also, when it's far from the edge, the shape looks a bit strange.

src/index.css
knod: Yeah. My concern is because I've specifically heard that organizations with little funding tend to have outdated systems and tend to rely on older technology, but maybe that's not the case here.

src/index.css
lianilychee: Thanks for the color definition!  For some reason, my eyedropper wasn't quite doing the right thing @_@

knod:
@lianilychee: I made a PR to your branch which converts stuff to css. It might help with readability, etc.

knod:
> tests may be failing remotely because Travis has not been updated to latest Node, and is happy to fix.

The PR's I've made recently with the updated version of Node haven't had trouble with the tests, so it might be something else.

PR #216, opened 2017-11-19
Fixes possible negative number issue in housing.js

knod:
This might be what turned the graph upside-down.


knod:
This one's small and simple, and people seem a bit busy. Mea culpa if something goes wrong.

PR #217, opened 2017-11-19
Deletes unused files

knod:


knod:
Sorry, folks, two days is long enough for me, I'm calling this one.

Issue #186, opened 2017-11-12
Make disclaimer more evident


lianilychee:
![image](https://user-images.githubusercontent.com/6895034/32703972-03099a0e-c7cc-11e7-8ff4-61288a07f61c.png)

~Consider bolding, or underlining text?~

Please bold the disclaimer, "This tool is a prototype and etc. etc."

knod:
Fixed by #222

PR #222, opened 2017-11-20
fixes issue 186 -- makes disclaimer more evident

sashamaryl:
bolds disclaimer in header


knod:
Fixes #186. Not sure if the comment will close it.

PR #223, opened 2017-11-20
Addresses #208, individual member data for household page

### LINKED ISSUE #208: Replace household form section with individual member data collection
- [x] Start with one member
- [x] Allow the addition of more members
- [x] Allow the deletion of members
- [x] For each member, collect:
   - [x] Their role. Ways to go about it:
      - Dropdown to choose 'Head of household', 'Spouse of head of household', or 'Member'.
      - Just provide the ability to select one member as a 'head' and one member as a 'spouse'.
      - Make the first member the 'head' automatically. Provide an option to select the 'spouse'. Don't allow the first member to be deleted.
   - [x] Age (number input).
   - [x] If they have a disability or not (checkbox).
   - [x] Are there more? Will need to double check the current questions. [Edit: this will be it for now]
- [x] Integrate new system with other code that uses those values:
   - [x] Expenses page
   - [x] housing.js
   - [x] snap.js

knod:
Fixes other files to work with the new data system.


sashamaryl:
Looks like everything is working! And the form structure is much easier to see. 
It can still be made clearer and easier, though!
- “not allowed” icon is confusing. We should differentiate the undeletable row with space, color, etc. instead. 
- the repetition in the first 
-

knod:
@sashamaryl: I hear those, and they sound like interesting ideas. I think you should draw up some ideas of what those should look like and post them in the associated issue, #208. I'm going to merge this, since it's functional, and the UI can be played with later once we've got some solid pictures of what our options would look like.

Issue #166, opened 2017-11-08
Build client/form values object


knod:
A complex question, but one we have some leads on that I'm going to use to begin work.

When managing client data, we want to:
- Propagate current values to future values as long as the future values haven't been 'touched'/'changed'.
- Make sure our client object values are always valid.

When managing user input, we want to:
- Not change values as a user is typing them.
- Validate the values the user is entering.
- Format output correctly when needed.

We've determined that:
- All input and display stuff will be taken care of by components with helper functions.
- We could use a base 'futurizable' class to propagate values to future data, store whether a future input was touched, and make the values serializable.
- We could build other classes on top of that if needed (though we may start with plain old functions for now until we know what we need to abstract).
- We do not want setters and getters that hide what's really going on.

An image from our brainstorming:
![form-values](https://user-images.githubusercontent.com/4466164/32575917-5f6bd63a-c4a3-11e7-91a1-382cdb4dd620.jpg)

Here was an initial simple demo (contains logic for maintaining and validating user data as it's typed) where values are stored in a client object and plain old normalizing/transforming functions are used: https://codepen.io/knod/pen/mqPLEB?editors=0010

knod:
This is the branch I'm doing work on: https://github.com/knod/cliff-effects-1/tree/form-values. Fair warning - don't rely on it for code, but feel free to discuss.

knod:
I think at this point we've settled on our model and how we're going to handle values. I call this one cooked.

PR #232, opened 2017-11-28
[PRESSING 1] Gets client values started, converts 'Current Benefits', #166

knod:
Somewhat urgent. These PR's need to move forward in order to not have a million conflicts. This is the first of several PR's. I'm converting our code to use the new client object. I'm going to try to break this out into categories, because reviewing one PR with all the changes will be a monumental task, so I'm trying to do it a chunk at a time, but these will need to be taken care of promptly.


src/utils/objectManipulation.js
```diff
@@ -0,0 +1,15 @@
...
+// I can't figure out how to use lodash's
+// `cloneDeep`, so I'm resorting to this.
+const cloneDeep = function ( obj ) {
+  return JSON.parse( JSON.stringify( obj ) );
+};
+
+
+export {
+  cloneDeep
+};
```
shp54: If you're using lodash's `cloneDeep`, can this be safely deleted?

src/utils/valueFixers.js
```diff
@@ -0,0 +1,73 @@
...
+  climateControl: returnSame,
+  nonHeatElectricity: returnSame,
+  phone: returnSame,
+  fuelAssistance: returnSame,  // to bool
+  otherExpenses: returnSame,  // money
+
+};  // end valueFixers
+
+
+export { valueFixers };
```
shp54: This could export a function that accesses the field in `valueFixers` or returns a default fixer (like `returnSame`). I think this would clean up a lot of repetition.

src/utils/objectManipulation.js
knod: This can! I thought I did! Thanks for picking up on that.

src/utils/valueFixers.js
knod: @shp54: Is this a blocking change? If not, I'd like to merge this and then discuss other options with you in person.

shp54: Not a blocking change, so I'll approve.

PR #233, opened 2017-11-28
[PRESSING 2] Converts household vals, addresses #166

knod:
The second of these. Merge #232 first.

### DIFF
--- a/src/containers/VisitPage.js
+++ b/src/containers/VisitPage.js
@@ -15,7 +15,8 @@ import { getHousingBenefit } from '../programs/state/massachusetts/housing';
 
 // Object Manipulation
 import { setNestedProperty } from '../utils/setNestedProperty';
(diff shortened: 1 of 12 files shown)


src/containers/VisitPage.js
```diff
@@ -82,7 +89,37 @@ class VisitPage extends Component {
...
+  changeClient = (evnt, { route, name, value, checked, time }) => {
+
+    route = route || name;
+
+    var val = value;
+    if ( typeof checked === 'boolean' ) { val = checked; }
+
+    // GOAL 1. Clone, not reference
+    var client      = cloneDeep( this.state.client ),
+        userChanged = {...this.state.userChanged},  // only 1 deep
```
shp54: Not a dealbreaker, but you can use `clone` from lodash.

PR #236, opened 2017-11-29
Splash header inquest

sashamaryl:
this is meant to continue pull request #185   
test of travis


src/containers/HomePage.js
```diff
@@ -1,47 +1,73 @@
...
+            marginBottom: '-15px'}}>
+              Cliff Effects Tool
+            </h1>
+
+            <h2 style={{
+            fontSize: '2em',
+            color: '#efefef',
+            textAlign: 'right',
+            marginTop: '-15px',
+            marginRight: '45px'}}>
```
ManickYoj: Generally prefer styling in the CSS file, but that needn't block this from going out.

src/components/Header.js
```diff
@@ -7,7 +10,8 @@ import {
 import { MainMenu } from '../MainMenu';
 import FixedMenu from '../FixedMenu';
 
-const Header = ({fix}) => {
+const Header = ({fix, location}) => {
+  if (location.pathname === '/') return (<div></div>)
```
ManickYoj: This is hax. I wrote it, so the shame is mine, but there ought to be a fast follow on issue here to find a better way to not display the header on the home page.

PR #241, opened 2017-11-29
Start writing test descriptions for cashflow

donblair30:
Write a few test descriptions. 
not a lot, but I want to commit what I have.

### DIFF
--- a/src/test/utils/cashflow.test.js
+++ b/src/test/utils/cashflow.test.js
@@ -55,12 +55,6 @@ test('Current SSDI cashflow 2', () => {
 // Use toCashFlow to fetch each current property on the test client, confirming that 
 // it's the value we expect. 
 
-// Use toCashFlow to fetch an invalid current property on the test client, confirming that 
-// it's invalid. 
-
 // Use toCashFlow to fetch each future property on the test client, confirming that 
 // it's the value we expect. 
-
-// Use toCashFlow to fetch an invalid future property on the test client, confirming that 
-// it's invalid. 
 
--- a/src/test/utils/cashflow.test.js
+++ b/src/test/utils/cashflow.test.js
@@ -44,4 +44,23 @@ test('Current SSDI cashflow 2', () => {
 	expect(cashflow.toCashflow(testClientList[0], 'current', 'SSDI')).toBe(99);
 });
 
+/** 
+* Test toCashFlow method.
+* For now, there are just test descriptions. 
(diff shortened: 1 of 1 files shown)


src/test/utils/cashflow.test.js
```diff
@@ -44,4 +44,23 @@ test('Current SSDI cashflow 2', () => {
...
+* For now, there are just test descriptions. 
+* These cannot be implemented until the client object is ready. 
+*/
+
+// Create a test client, setting new values for every prop for current values. 
+
+// Use toCashFlow to fetch each current property on the test client, confirming that 
+// it's the value we expect. 
+
+// Use toCashFlow to fetch an invalid current property on the test client, confirming that
```
knod: The comments do look like comments, which is great! Is there a reason for the tests of invalid values, though, since client values should always be valid?

src/test/utils/cashflow.test.js
```diff
@@ -44,4 +44,23 @@ test('Current SSDI cashflow 2', () => {
...
+// Use toCashFlow to fetch each current property on the test client, confirming that 
+// it's the value we expect. 
+
+// Use toCashFlow to fetch an invalid current property on the test client, confirming that 
+// it's invalid. 
+
+// Use toCashFlow to fetch each future property on the test client, confirming that 
+// it's the value we expect. 
+
+// Use toCashFlow to fetch an invalid future property on the test client, confirming that
```
knod: Same comment here

src/test/utils/cashflow.test.js
donblair30: For starters, thank you for reviewing and replying with comments!  Discussion is good. 
Here's what I'm thinking.  Is it helpful to flag an invalid property in case someone misspells it when passing to this routine?   I'm trying to anticipate future coding errors.  But I'll defer to you, whether you think this is helpful.

knod: Client values must always be valid, but if they're not toCashFlow is not the place that that should be dealt with. We can talk about it more detail, though, if you're still concerned.

donblair30:
I just removed the test descriptions in question and pushed the commit.  Looks like github is smart enough to automatically include that in this pull request without any effort on my part.  This should be good to go.  Although the change is trivial, this is helpful to get my feet wet on making changes to pull requests.

knod:
Accept now with room for discussion later!

PR #252, opened 2017-11-30
Fixed menu in <Header>, no `withRouter()`, fixes #239

### LINKED ISSUE #239: Stabilize Conditional Logic for Header
Currently the `Header` file has a [hacky line](https://github.com/codeforboston/cliff-effects/blob/dev/src/components/Header.js#L14) that makes it only display if the current path is '/'. This is pretty fragile because changing the path can cause the header to display on the home page.

Action item here is to implement a more robust toggle for the header visibility.

knod:


knod:
Fixed by #259!

PR #257, opened 2017-11-30
Refactor Header toggle to use Route

### LINKED ISSUE #239: Stabilize Conditional Logic for Header

wacii:
Use React Router to match path when determining whether to show the `<Header>` component. Also moves the `<Header>` related state into the `<Header>` component rather than maintain it in `<App>`. Less re-rendering that way.

Hi. Was looking into Code for Boston and some of the code y'all have written, and thought I'd do something useful while I was here... and steal @knod's issue~

Closes #239.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,10 +4,6 @@ import {
   Route,
 } from 'react-router-dom'
 
-import {
-  Visibility,
-} from 'semantic-ui-react';
-
 import HomePage from './containers/HomePage'
 import AboutPage from './containers/AboutPage'
 import VisitPage from './containers/VisitPage'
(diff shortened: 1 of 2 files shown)


wacii:
Hum, well that's no good. The `+` means much the same thing it does in regex, one or more instances. So here `path='/:rest+'` means match one or more parameters. So `/foo` or `/foo/baz` but not `/`, since there is no parameter to be matched.

Perhaps that was a little too cute. You could also do...

```js
<Route exact path="/" children={({match}) => !match && <Header />} />
```

Which reads a little more naturally. Although you'll end up creating a bunch of functions as written.

wacii:
Ah! I didn't notice this had already been fixed. And twice! What I get for not taking the time to familiarize myself...

knod:
I didn't notice that there was an extra PR out on it either, but I like a lot of what your version is doing. '+' makes sense now, but what does ':rest' do? Is that a keyword in jsx?

wacii:
`':rest'` isn't special, just an arbitrary parameter name. I named it for what it was doing, capturing the rest of the parameters.

Normally parameters are used like `'/users/:userId'` to capture identifying information for some resource, but they can capture anything. React Router uses [path-to-regexp](https://github.com/pillarjs/path-to-regexp) (Express also uses this) if you wanted to read more about what you can do with the paths.

PR #271, opened 2017-12-03
Extract formHelpers to separate files

wacii:
Thought I'd tidy up as I looked through the code. In addition to extracting the code into separate files I...
* updated JSDoc
* normalized styling a bit
* extracted css styles so you're not creating objects each render
* removed unused variables
* addressed some of the simpler, refactor-y todos

One notable change is that `<CashFlowInput>` was removed. Following one of the todos and having the parent do some of the work made the component unnecessary.

I suggest removing or reducing the use of JSDoc, at least for components. It is not as expressive as PropTypes for describing props, and the components themselves are only really meaningful when consumed by React. That and the JSDoc usage was very inconsistent. If there was information provided it tended to be out of date.

I would be willing to do this, leave the summaries and replace the rest of the JSDoc with PropTypes, if you're interested.


knod:
We may be able to deal with the cashflow thing by abstracting further, since I'm finding a similar functionality is needed for another component, so we may be able to get rid of that component in the end anyway, but it'll be an adjustment regardless.

knod:
@wacii: Do you think we could catch each other on Slack sometime and have a real-time focused conversation about the different issues? What's your username there?

wacii:
Sure. It's wacii.

wacii:
But if you're actively developing the code, not sure you'd want to let refactoring get in the way. It was mostly just shuffling code around and doing some light pruning.

knod:
Thanks for clarifying! I actually like some of what you've done there, so it'd be great to talk about. There are also other questions I have if you're up for them. I'm on most days during a fair amount of the day. When can I find you on?

wacii:
@knod I'm on now. And I can be on most nights around this time.

Issue #262, opened 2017-12-01
Explore only sending valid values


knod:
Explore the possibility of never sending invalid values to be set in the client object. It should really only be an issue for cash-flow values.

Plan: Keep one variable that contains the current input element, the current value, and whether it's valid or not. Only set the value onto the client object when it's valid. Otherwise, style the input differently and, when the user leaves that input, restore it to a valid value.

Empty number fields should be a valid value (0), but if we want to actually display 0's, that value should also not be displayed till the user leaves the field. That may be tricky.

PR #272, opened 2017-12-03
Addresses #262, only send valid input values

knod:
Not fixed yet, still have to do 'Household'.

### DIFF
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -15,6 +15,8 @@ import {
 
 // UTILITIES
 import { roundMoney, toMonthlyAmount } from '../utils/math';
+import { isPositiveNumber } from '../utils/validators';
+import { toMoneyStr } from '../utils/prettifiers';
 
 
 // ========================================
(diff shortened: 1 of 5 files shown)


wacii:
Some comments on `<CashFlowInput>`

A lot of what has been hard-coded into `<CashFlowInput>` could be extracted as props. You could provide `isPositiveNumber()` as `props.validate()`, `toMoneyStr()` as `props.format()`, and
```js
(evnt, value) => {
  var monthly = toMonthlyAmount[ interval ]( evnt, value );
  var obj = { name: generic , value: monthly };
  store( evnt, obj );
}
```
as `props.updateClient()`. Even `'number'` could be abstracted as `props.type`.

I don't think you need to keep track of focused state. Just set `focusedValue` to `toMoneyStr(value)` when the component is instantiated and `onBlur()`.

You don't need to store `value` in state. You never update it.

You don't need to use a function in `setState()`. The only reason to do this is if you are accessing the previous state.

wacii:
@knod Never mind about not needing to keep track of focus. I forgot the inputs were linked.

PR #273, opened 2017-12-04
Abstraction for only returning valid values

knod:
For values in number fields - income and, soon, age.

### DIFF
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -345,16 +345,10 @@ const IntervalColumnHeadings = function ({ type }) {
 };  // End IntervalColumnHeadings{} Component
 
 
-
-/** @todo description
- *
- * @function
- * @param {object} props
- * @property {object} props.__ - explanation
- *
- * @returns Component
+/** 
+ * @todo description
  */
-class ManagedInput extends Component {
+class ManagedNumberField extends Component {
   constructor ( props ) {
     super( props );
     // Need updating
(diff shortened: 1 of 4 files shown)


src/forms/formHelpers.js
```diff
@@ -443,50 +514,54 @@ class CashFlowInput extends Component {
 */
 const CashFlowRow = function ({ generic, timeState, setClientProperty, children, labelInfo, type, time }) {
 
+  var updateClient = function ( evnt, value, interval ) {
+    var monthly = toMonthlyAmount[ interval ]( evnt, value ),
+        obj     = { name: generic, value: monthly };
+    setClientProperty( evnt, obj );
+  };
```
wacii: You could get around passing interval to the `<ManagedInput>` if you wanted it to be extra general.

```js
const updateClient = (interval) => (
  (evnt, value) => {
    const monthly = toMonthlyAmount[ interval ]( evnt, value );
    const obj = { name: generic, value: monthly };
    setClientProperty( evnt, obj );
  }
);
```
Then when you are defining your inputs:

```js
<ManagedInput {...props} store={updateClient('weekly')} />
```

src/forms/formHelpers.js
```diff
@@ -443,50 +514,54 @@ class CashFlowInput extends Component {
...
+   * @var
+   *
+   * @todo Add some kind of UI indication when it's the same as the 'current'
+   *     value. What if some of the row's values are the same and some are
+   *     different?
+   */
+  var baseVal   = timeState[ generic ],
+      classes   = [ time, type, 'cashflow-column' ],
+      baseProps = {
+        generic:  generic,
```
wacii: `<ManagedInput>` doesn't actually use `generic` anymore.

src/forms/formHelpers.js
```diff
@@ -347,89 +347,160 @@ const IntervalColumnHeadings = function ({ type }) {
...
+// * @todo Make this more generic and make the caller
+// * handle more of the specifics.
+// *
+// * @function
+// * @param {object} props
+// * @property {object} props.__ - explanation
+// *
+// * @returns Component
+// */
+class CashFlowInput extends Component {
```
wacii: If this isn't being used anymore, I think it should be removed. If you want to reference it, you can always reach back in git's history.

src/forms/formHelpers.js
```diff
@@ -347,89 +347,160 @@ const IntervalColumnHeadings = function ({ type }) {
...
-* @returns Component
-*/
-class CashFlowInput extends Component {
+ *
+ * @function
+ * @param {object} props
+ * @property {object} props.__ - explanation
+ *
+ * @returns Component
+ */
```
wacii: This information isn't correct, so I'd remove it. Maybe just leave:
```js
/**
 *  @todo description
 */
```

src/forms/formHelpers.js
```diff
@@ -347,89 +347,160 @@ const IntervalColumnHeadings = function ({ type }) {
...
         value     = { value }
-        name      = { this.name }
-        className = { this.className }
-        style     = {{ width: '7em', display: 'inline-block' }}
+        name      = { name }
+        className = { className }
         onChange  = { this.handleChange }
         onFocus   = { this.handleFocus }
         onBlur    = { this.handleBlur }
         type      = { 'number' } />
```
wacii: `type` could be passed in as a prop. Otherwise I'd consider renaming the component to reflect that it is for numbers only.

src/forms/formHelpers.js
knod: Yeah, leaving that was definitely not intentional.

src/forms/formHelpers.js
knod: It's an interesting solution and I'll have to think about it. I find that kind of function nesting more complex to hold in my mind when coding, so I tend not to use it where possible. There's at least one place in the code base where do that and I'm not super happy with it.

src/forms/formHelpers.js
knod: Renamed for now, though worth considering the prop at some later date.

Issue #243, opened 2017-11-29
Clearly express the valid values for each client property


knod:
Perhaps on the client default object.

wacii:
Once more inputs are validated, you might maintain a file for the validations like 'src/utils/valueFixers.js'. Code is less likely to become out of date than comments.

knod:
@wacii There's a file for that ('src/utils/validators.js'), but validation is happening in components - no invalid values will be passed to the client object. Part of why it's handled there is to allow easier feedback to the user about invalid values. We can consider using a 'valueValidators.js' kind of thing, but it seems like it may add unnecessary complexity/abstraction. Maybe seeing it in action would be different.

knod:
Fixed by #277

PR #277, opened 2017-12-04
Notes on client values and what they're expected to be

knod:
It's meant to be prettier to look at too, for easier consumption. Do take a look at the file itself to see how it really looks before hating it. Feel free to hate it after that.


knod:
Fixes #243

PR #282, opened 2017-12-05
Making the site more tablet friendly

wacii:
8649e6a
The next/previous buttons just kept shrinking as the screen did, making them much too small on anything but a computer.

a3a7888
The steps are too large for smaller screens, regardless of how they are oriented. And as they are not strictly necessary, I simply hid them on tablet and mobile. There is still a range close to the cutoff where the steps don't look great, but it is an improvement. (The cutoff for tablet could be changed to address this issue.)

Some alternatives to simply hiding the steps might be replacing them with a dropdown or hiding all but the current step's text.

~The alerts sidebar also needs to be updated, but there were a couple of things I wanted to ask first. Is the `<Reveal>` in `<BenefitCard>` necessary? The information the `<Reveal>` displays now doesn't seem to be super helpful and hover effects are pretty useless on touch devices. Would it be alright to move the `<AlertSidebar>` above the `<FormSection>` on smaller screens? Otherwise I would suggest shortening the header 'Warnings and Alerts' to just 'Alerts', as the current header tends to overflow on smaller screens.~


wacii:
Forgot to update the snapshots...

knod:
The alerts are being/have been? removed. There may be other alternatives to the steps, but maybe we can talk about those tonight. Feel free to update the snapshots and push again.

src/forms/formHelpers.js
```diff
@@ -23,29 +22,6 @@ import { toMoneyStr } from '../utils/prettifiers';
...
-* @function
-* @param {object} props
-* @property {object} props.func - A function that is run when the
-* button is clicked.
-* @property {string} props.name - The text to be displayed on
-* the button.
-*
-* @returns Component
-*/
-const BottomButton = function(props){
```
knod: This was here because of the 'Save' button that used to exist. I'm not sure if we're now going to replace it with a 'Finish'/'Erase data' button, so it may be worth keeping the button separate, or abstracting (as I started to do in a different PR, though if we do that, it'd probably be worth doing almost what you did here).

wacii:
> The alerts are being/have been? removed.

Well shoot. I do kind of remember that.

Snapshots updated.

src/forms/formHelpers.js
wacii: Hmm. I got rid of the `<Grid>` so there was nothing in `<BottomButton>` other than a `<Button>` with a few extra props. But if its used elsewhere...

PR #284, opened 2017-12-05
Improvements to number input validation handling component

knod:
Adds a description, which I think is warranted

### DIFF
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -347,6 +347,15 @@ const IntervalColumnHeadings = function ({ type }) {
 
 /** 
  * @todo description
+ * 
+ * @param {Object} props
+ * @param {number || string} props.value - Valid client value
+ * @param {string} props.name - For HTML name property
+ * @param {string} props.className - HTML class names
+ * @param {*} [props.otherData] - Sent back to `store()`
+ * @param {function} props.format - Given `value`. Must return what you want shown in the number field. Not sure how to jsdoc this.
+ * @param {function} props.validate - Given `value`. Must return boolean.
+ * @param {function} props.store - Given an event, `value`, [`otherData`]
  */
 class ManagedNumberField extends Component {
   constructor ( props ) {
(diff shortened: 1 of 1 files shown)


src/forms/formHelpers.js
```diff
@@ -347,6 +347,15 @@ const IntervalColumnHeadings = function ({ type }) {
 
 /** 
  * @todo description
+ * 
+ * @param {Object} props
+ * @param {number || string} props.value - Valid client value
+ * @param {string} props.name - For HTML name property
+ * @param {string} props.className - HTML class names
+ * @param {*} [props.otherData] - Sent back to `store()`
+ * @param {function} props.format - Given `value`. Must return what you want shown in the number field. Not sure how to jsdoc this.
```
wacii: With the `@callback` tag I think. http://usejsdoc.org/tags-callback.html

src/forms/formHelpers.js
```diff
@@ -377,7 +386,7 @@ class ManagedNumberField extends Component {
         valid = this.props.validate( value );
 
     if ( valid ) {
-      this.props.store( evnt, value, this.props.interval );
+      this.props.store( evnt, inputProps, this.props.otherData );
```
wacii: Not sure why you're passing `inputProps` here instead of `value`. But then again I don't understand why Semantic UI includes these in the first place, much less merged with the value (which is already available on the event). So there's probably something I'm missing here.

shp54: `evnt` isn't a representation of an input event. It's sort of like a Redux action. (One of the other files has where this is used - will update the comment when I find it.)

knod: I think the `inputProps` are the props that you give to the React component, not just what's on the element. I could be wrong. I saw no reason not to pass along the whole thing. Is this a blocking issue?

src/forms/formHelpers.js
knod: Thank you for the info, that's a great piece of knowledge! I think for now, given time constraints, I'll put a todo in there about it and leave it for a later date.

src/forms/formHelpers.js
wacii: No, its not a big deal.

`inputProps` are the props of `<Input>` merged with `evnt.target.value`. [source](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Input/Input.js#L135-L138)

knod: Thanks for the further info! Looks like it at least includes the Component props.

Issue #134, opened 2017-10-30
Changes needed for tablet use?


knod:
This answer is needed fairly urgently. I recall something about clients using tablets to fill out the form. Is that still the case? If so, what browsers will they have access to? If not Chrome, we'll need to really dig into making this compatible with other browsers. Also, do we need the page to be fluid? It could probably be used with a tablet in landscape modes for some sizes of tablets without cutting off part of the page. Not sure about portrait mode.

lianilychee:
cc @anhase - thoughts?

Michelle, that's an excellent thought, thanks for remembering!

I think we should keep to our priority stack, of which rendering SNAP and Section 8 benefit amounts in results pg were first and foremost.

I can run tests with a tablet first, before we decide criticality regarding browser-size reactivity.  re: browser compatibility: my first hunch is that this is out-of-scope, but tell me @knod - how complex do you foresee that being, on the following scale?  1 is easy.  2 is medium.  4 is hard.  8 is "so-complex-we're-gonna-need-multiple-devs"

knod:
Hehe, interesting way of putting it. Without some preliminary research, I'm not comfortable giving any numbers at all. Looking into it should be a 1-2, though.

lianilychee:
![image](https://user-images.githubusercontent.com/6895034/32202302-19482476-bdb2-11e7-8b88-d78f89cd9c3b.png)
 iPad, chrome, landscape orientation.

knod:
@lianilychee Thanks for testing it out!

knod:
Worth another go?

ManickYoj:
@knod I think this was resolved by @wacii on #282 . Can we consider this resolved? It's not perfect, but it's usable, which is a massive step.

wacii:
I'm doing some more work on this.

wacii:
I addressed all the problems I found. There are likely more out there, but only so many I can find using a single device.

knod:
*Now* fixed by merge of #292.

PR #292, opened 2017-12-06
Some more design changes

wacii:
Took a bit of tinkering and backtracking to get something that works, so I ended up with one big commit. I can go back and split things up, but I wanted to run the changes by y'all first.

% heights don't play well with Flexbox on Mobile Safari. So once Flexbox starts being used, its nothing but Flexbox. So much Flexbox.

Removed the min-heights on `<FormPartsContainer>`. Safari didn't like those either. Instead I just grew the container so that its bottom was more or less flush with the footer.

Shortened some of the step text. 'Current Income' => 'Income'

The steps and forms now take up the full width of the container.

The beta warning is shortened on tablets to 'This tool is a prototype'.

see #134

### DIFF
--- a/src/BetaWarning.js
+++ b/src/BetaWarning.js
@@ -1,10 +1,15 @@
 import React from 'react';
-import { Menu } from 'semantic-ui-react';
+import { Menu, Responsive } from 'semantic-ui-react';
 
 const BetaWarning = function ( props ) {
   return (
     <Menu.Item>
-      <div><strong>This tool is a prototype and should not be used to make financial decisions.</strong></div>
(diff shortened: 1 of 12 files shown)


wacii:
The first commit is still kinda big, but I pulled out the easier changes and added one to improve the Household form. Overflow is still a problem there, but by removing a seemingly unnecessary `max-width`, I got it to work on my iPad 2. Whether it will need work depends on which devices we will be supporting.

knod:
It looks good to me, but I'd like @lianilychee to have eyes on it before it goes through.

PR #294, opened 2017-12-07
Valid member age, feedback to user, fixes #262

### LINKED ISSUE #262: Explore only sending valid values

knod:
Try putting negative numbers or decimals into age and you'll see.

### DIFF
--- a/src/forms/Household.js
+++ b/src/forms/Household.js
@@ -6,18 +6,19 @@ import {
   Dropdown,
   Header,
   Checkbox,
-  Icon,
-  Input
+  Icon
 } from 'semantic-ui-react';
 
 // PROJECT COMPONENTS
 import {
   FormPartsContainer,
-  ColumnHeading
+  ColumnHeading,
+  ManagedNumberField
 } from './formHelpers';
 
 // COMPONENT HELPER FUNCTIONS
 import { getTimeSetter } from '../utils/getTimeSetter';
+import { isPositiveWholeNumber } from '../utils/validators';
 
 // OBJECT MANIPULATION
 import { cloneDeep } from 'lodash';
@@ -219,12 +220,13 @@ const MemberField = function ({ household, time, setHousehold, setClientProperty
       </Columns.Two>
 
       <Columns.Three>
-        <Input
-          className = {time + '-member-age ' + time}
-          onChange  = {onMemberChange}
+        <ManagedNumberField
           value     = {member.m_age}
           name      = {'m_age'}
-          type      = {'number'} step = {'1'} min = {'0'} />
+          className = {time + ' member-age ' + time}
+          validate  = {isPositiveWholeNumber}
+          format    = {function ( value ) { return value; }}
+          store     = {onMemberChange} />
       </Columns.Three>
 
       <Columns.Four>
--- a/src/utils/validators.js
+++ b/src/utils/validators.js
@@ -13,8 +13,9 @@ const isPositiveNumber = function ( str ) {
 
 
 /** Returns true if a string represents a positive integer */
+/** @todo Change name to 'isWholeNumber'. */
 const isPositiveWholeNumber = function ( str ) {
-  return /^[0-9]*$/.test(str);
+  return str !== '' && /^[0-9]*$/.test(str);
 };
 
 


src/forms/Household.js
```diff
@@ -219,12 +220,13 @@ const MemberField = function ({ household, time, setHousehold, setClientProperty
       </Columns.Two>
 
       <Columns.Three>
-        <Input
-          className = {time + '-member-age ' + time}
-          onChange  = {onMemberChange}
+        <ManagedNumberField
           value     = {member.m_age}
           name      = {'m_age'}
-          type      = {'number'} step = {'1'} min = {'0'} />
```
wacii: I would suggest keeping `step` and `min` and changing `<ManagedNumberInput>` to pass through extra props like these.

src/forms/Household.js
```diff
@@ -219,12 +220,13 @@ const MemberField = function ({ household, time, setHousehold, setClientProperty
...
+          className = {time + ' member-age ' + time}
+          validate  = {isPositiveWholeNumber}
+          format    = {function ( value ) { return value; }}
```
wacii: The formatter will accept `023`. Super minor.

src/utils/validators.js
```diff
@@ -13,8 +13,9 @@ const isPositiveNumber = function ( str ) {
 
 
 /** Returns true if a string represents a positive integer */
+/** @todo Change name to 'isWholeNumber'. */
 const isPositiveWholeNumber = function ( str ) {
-  return /^[0-9]*$/.test(str);
+  return str !== '' && /^[0-9]*$/.test(str);
```
wacii: Could use `/^\d+$/.test(str)`

src/utils/validators.js
```diff
@@ -13,8 +13,9 @@ const isPositiveNumber = function ( str ) {
...
 };
```
wacii: Minor performance thing, but there's no reason to recreate the RegExp object every time you invoke the function.

src/forms/Household.js
knod: There are no arrow buttons, so currently no point in `step` and `min`, since those attributes don't control anything about what happens when people type in values. I'm not sure it's worth future proofing these.

src/forms/Household.js
knod: I'm ok with it for now unless someone else objects to it. It's good to keep in mind, though, and maybe we'll deal with it when beta testing is going on.

src/utils/validators.js
knod: That would work too, but I find this more readable unless there's a particular advantage to `\d`.

src/utils/validators.js
knod: True! We should keep this in mind as an improvement. If you want to make a PR for it, I'd be into that. If not, we'll try to catch it later. Wish we could have a secondary list of issues that would stay out of sight without having to use search filters.

knod: Maybe an issue with just a list of minor improvements.

src/forms/Household.js
wacii: Some browsers will add an alternative way of interacting with number inputs. I know in Chrome you can use the arrow keys to step the value up and down.

Minor though.

src/utils/validators.js
wacii: No, there isn't. I probably should have left that out, as I was mostly suggesting using `+` instead of `*` to avoid `str !== ''`.

src/utils/validators.js
wacii: You could use a project.

src/forms/Household.js
knod: Ah, I didn't even consider arrow keys. That's a good point then. This should definitely go on the 'small improvements' list!

src/utils/validators.js
knod: Oh, got it. Yeah, totally missed the `+`. That would probably be a better choice, though I did like how explicit `str !=== ''` is. Another for the list, though.

src/utils/validators.js
knod: Should definitely look into that (they're still a new feature to me).

PR #298, opened 2017-12-08
Sample client data and expected benefit values

knod:
For now they're in one file. May break them up later to make it easier for individuals to add more, but I'm not sure how much more data we're going to get and right now this file has the columns listed at the top, which I think will be useful.

### DIFF
--- a/src/test/utils/sampleClients.js
+++ b/src/test/utils/sampleClients.js
@@ -9,59 +9,59 @@ const CLIENT_ROW_N = {
 
   current: {
     // Current programs
-    hasSnap:                        // COL-T      boolean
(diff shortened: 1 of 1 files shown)


knod:
@nickbreid: It's actually a clarification that I just got back today and found out that only ones that say 'Section 8' are actually Section 8. Ideally, I'd love if those were fixed. On the other hand, it's not super relevant because this data, as it stands, is not sufficient for testing Section 8 calculations, and Section 8 doesn't have much of an effect on SNAP calculations. If we're just testing SNAP, Section 8 data is just flavor text.

Do you feel up for making a quick github edit of the files to reflect the change in our understanding?

nickbreid:
@knod: Done.

Issue #263, opened 2017-12-01
Convert SNAP and Section 8 calculations to using a specific timeframe


knod:
They need to either return values for current benefits or for future benefits. They currently return only future benefits and code that wants current benefits have to hack.

PR #301, opened 2017-12-09
Less timeframe hacking, #263

knod:
Section 8 is going to need the whole `client` object, and I'd rather the function signature for benefit calculations be the same, so we'll never be able to pass in just the `current` or `future` client, but we can get closer to that. Something that's a bit less fragile.

Next is to convert 'housing.js', then to clean up the temporary measures. Just want a fresh eye on the basic premise.


knod:
@wacii: Agreed, that's one of the steps on the wishlist of changes. A few of the cash flow functions were there to make sure that benefit calculations could always get valid values. Now that we have solid valid values, there are ones that can be completely removed.

PR #303, opened 2017-12-09
Award housing benefits only if selected

### LINKED ISSUE #281: 'Inactive' benefit programs still being used for table calculations
### Images:
too-many-benefits

### Replicate:
1. Start form
2. Go to 'Predictions' page

### Result I'd Expect:
- Table to show all 0's

### Actual Result:
- Table shows net total as 2304

### Environment:
Browser type and version: Chrome v61.0.3163.100
App version or commit id at the time: c4c643feb6eca31e8afef86ed5538b6529438456

### Other Notes
The table calculations probably need to be done conditionally too.

wacii:
Added more conditionals to `<BenefitsTable>` calculations. Should probably be done in `getSNAPBenefits()` and `getHousing()`, but I wasn't sure about the result object being returned. The extra fields don't seem to be used. Can I just remove them?

Closes #281

### DIFF
--- a/src/forms/BenefitsTable.js
+++ b/src/forms/BenefitsTable.js
@@ -34,11 +34,11 @@ const BenefitsTable = function ( props ) {
 
 
 
-  var SNAPBenefitCurrent  = currentClient.current.hasSnap ? Math.round( getSNAPBenefits( props.client, 'current' ).benefitValue * 12 ) : 0,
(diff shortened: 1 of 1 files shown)


knod:
@wacii: One of the steps I'd like to do is removing the Results object completely and just returning number values. Ideally it'd be done when other people aren't messing with the files.

PR #307, opened 2017-12-10
Removes ensuring no undefined money values

knod:
Gets rid of `toCashflow()`, though the new `sumProps()` may not be in the right place.


knod:
Newp, not working as expected. My bad.

Issue #289, opened 2017-12-06
Scrolling bug due to header menus


wacii:
Set the window height to ~950px, so the page doesn't quite fit in the screen, then try and scroll down. The page will briefly scroll before jerking back up to the top. This is due to a page height difference introduced by `<MainMenu>` and `<FixedMenu>`.

This can be fixed by making the page heights equal (and smoothing out the transition), but I would instead suggest removing the fixed menu. The pages are never long enough to really need it and the menu transition is kind of choppy on my iPad anyways.

wacii:
The design has been updated, so the stutter no longer appears at ~950px for most of the form pages. Still easy to reproduce though; decrease the window height until the vertical scroll bar just appears, then try and scroll.

ManickYoj:
I agree with @wacii that we have no pages that really need a fancy fixed menu transition. Better to just have a smooth transition. I'll change this presently.

PR #310, opened 2017-12-10
Ask user for confirmation to leave page

### LINKED ISSUE #296: Add warning about loss of data when page is being refreshed
We used to have this way back near the very beginning, so if it doesn't come to us straight away, we can check out the files from the earliest commits.

wacii:
Closes #296.

### DIFF
--- a/src/components/ConfirmLeave.js
+++ b/src/components/ConfirmLeave.js
@@ -13,34 +13,25 @@ class ConfirmLeave extends React.Component {
     when: true
   };
 
-  addListener = () =>
-    (this.listener = window.addEventListener(
-      'beforeunload',
-      event => (event.returnValue = this.props.message || '')
-    ));
-
-  removeListener = () => {
-    if (this.listener !== undefined)
-      window.removeEventListener('beforeunload', this.listener);
-  };
+  handleBeforeUnload = event => (event.returnValue = this.props.message || '');
 
   componentDidMount() {
     if (this.props.when) {
-      this.addListener();
+      window.addEventListener('beforeunload', this.handleBeforeUnload);
     }
   }
 
   componentWillUpdate(nextProps) {
     if (nextProps.when && !this.props.when) {
-      this.addListener();
+      window.addEventListener('beforeunload', this.handleBeforeUnload);
     }
     if (!nextProps.when && this.props.when) {
-      this.removeListener();
+      window.removeEventListener('beforeunload', this.handleBeforeUnload);
     }
   }
 
   componentWillUnmount() {
-    this.removeListener();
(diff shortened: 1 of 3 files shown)


src/components/ConfirmLeave.js
```diff
@@ -0,0 +1,53 @@
...
+ * Ask the user for confirmation to leave the page.
+ * @extends React.Component
+ *
+ * @param props {object}
+ * @param props.when {boolean} - Whether the component should block
+ * @param props.message {string} - The message the browser shows, maybe
+ */
+class ConfirmLeave extends React.Component {
+  static defaultProps = {
+    message: 'Are you sure you want to leave the page with unsaved changes?',
```
knod: Since they can't actually save any changes, maybe something like, 'This action will erase all current data. Are you sure you want to do this?' would be more appropriate. Not a blocking note, but one I want to mark down for a next step.

wacii: Sure. I'll change the other instances as well.

knod: Maybe this belongs in VisitPage.js.

knod:
Also, when I go to the home page after going to the form and try to reload, I get that same outdated warning message about reloading. This is after a couple of attempts to reload without cache.

wacii:
So the listener wasn't being removed when the component was unmounted. I fixed that.

Most browsers are going to ignore whatever message you give them. `beforeunload` can be easily abused, so browsers generally give you very little control over what you can do in response. I'll see if there's anything else we can do though.

wacii:
Nothing I tried works. At least not in Chrome. I think its the built-in message or nothing.

wacii:
Simplified the React-related code and relied more heavily on the very strange `beforeunload` API.

Either not setting `event.returnValue` or returning `undefined` or `null` is supposed to not show the event. So I relied on that to enforce `props.when` instead of adding and removing the listener.

So maybe an improvement?

knod:
Not sure about which is more robust, but it is prettier.

The 'Changes may not be saved' message is the default that we can't change? If so, it it at least possible to re-write the 'Reset' button to bring up our own message and then, if confirmed, reload without the warning?

wacii:
@knod I found no way to affect the default confirmation window (in Chrome).

I'll give the `Reset` button its own confirmation modal.

knod:
Other than a little hop at certain window heights after the reset button (which I think might be related to #289, the fixed header), it looks good to me!

wacii:
Which component is that? `<Prompt>`?

knod:
@wacii: I assume it'd be `<ConfirmLeave>`, but maybe it'd be `<Prompt>`. I'm not sure which would be more appropriate.

PR #311, opened 2017-12-10
Fixes #196, calculations only return numbers, not Result instances

### LINKED ISSUE #196: Remove need for `Results.js`
## Purpose
`Result` is a class we use to format results from benefit calculations. @knod has opined we probably don't need it anymore and we can probably just have benefit calculations return numbers.

Note: this is a pure refactor. Probably not the highest priority for now.

## Implementation
To make this happen, we just need to track down all the imports of `Result` and rewrite the code that uses it to work without the `Result` file.

knod:


src/forms/BenefitsTable.js
```diff
@@ -32,11 +32,11 @@ const BenefitsTable = function ( props ) {
...
 
-  var SNAPBenefitCurrent  = client.current.hasSnap ? Math.round( getSNAPBenefits( client, 'current' ).benefitValue * 12 ) : 0,
-      SNAPBenefitFuture   = client.current.hasSnap ? Math.round( getSNAPBenefits( client, 'future' ).benefitValue * 12 ) : 0,
+  var SNAPBenefitCurrent  = client.current.hasSnap ? Math.round( getSNAPBenefits( client, 'current' ) * 12 ) : 0,
+      SNAPBenefitFuture   = client.current.hasSnap ? Math.round( getSNAPBenefits( client, 'future' ) * 12 ) : 0,
       SNAPDiff            = SNAPBenefitFuture - SNAPBenefitCurrent,
-      sec8BenefitCurrent  = client.current.hasHousing ? Math.round( getHousingBenefit( client ).benefitValue * 12 ) : 0,
-      sec8BenefitFuture   = client.current.hasHousing ? Math.round( getHousingBenefit( client ).benefitValue * 12 ) : 0,
+      sec8BenefitCurrent  = client.current.hasHousing ? Math.round( getHousingBenefit( client ) * 12 ) : 0,
+      sec8BenefitFuture   = client.current.hasHousing ? Math.round( getHousingBenefit( client ) * 12 ) : 0,
```
wacii: Now that `getSNAPBenefits()` and `getHousingBenefit()` return numbers, these conditionals could be moved into those functions.

knod: Hmm, interesting thought. I can see the perspective that if a person isn't receiving a benefit, they're getting 0 for that benefit, but I could also see that if someone isn't receiving a benefit, we shouldn't be trying to calculate it in the first place. We should keep it mind as we see how things pan out.

PR #313, opened 2017-12-11
#167, reorganize components to 'src/components'

knod:


knod:
Need this one to move forward, so moving forward.

knod:
Oh, didn't see the review. Thanks @wacii! Yeah, the exports are inconsistent. I'll mark this with 'discussion needed' as well and at some point I'll figure out how projects work.

Issue #315, opened 2017-12-11
Test snap.js with sample clients


knod:
Add sample client tests
- [x] CLIENT_DEFAULTS
- [x] CLIENT_ROW_3
- [x] CLIENT_ROW_4
- [x] CLIENT_ROW_5
- [x] CLIENT_ROW_6
- [x] CLIENT_ROW_7
- [x] CLIENT_ROW_8 (row 8 may need to be removed because it doesn't have SNAP benefits)
- [x] CLIENT_ROW_9
- [x] CLIENT_ROW_10

johnschoeman:
Added tests for basic calculation of snap benefits with default client.  Would be good to have calculations verified against MA calculators and to add more test cases.

knod:
This PR didn't use sample clients yet and it'd be good to have more extensive testing before closing. Definitely a good first step, though. Perhaps we can assign more folks to this one.

knod:
This was covered in more detail in #329 as well. We'll take care of sample row 8 some other time.

PR #326, opened 2017-12-14
Allows more granular testing of SNAP funcs

knod:
#315. First step in adjusting SNAP clacs for testsing.

### DIFF
--- a/src/programs/federal/snap.js
+++ b/src/programs/federal/snap.js
@@ -26,16 +26,16 @@ const getSNAPBenefits = function ( client, timeframe ) {
   client = client[ timeframe ];
 
   var finalResult = 0;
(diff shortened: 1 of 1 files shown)


maxsean:
I agree with wacii about exporting `getSNAPBenefits()` as default and helpers as individually named exports, but again it's just a style. I do like the clear separation between the two. This is a lot easier to interpret for new members.

knod:
@wacii, @maxsean: Glad to hear it. We can definitely do the default thing sometime. Named functions are a good idea and can be the next step. For now, the current console logs are more verbose, but they do provide the method names. Maybe that's just a Chrome thing.

wacii:
@knod If it shows up named in the console, specifically in stack traces, that's fine then. Although I wish I understood why. Hmm...

PR #327, opened 2017-12-14
Adds ability to check section 8 calculation helpers

knod:
Exports them as properties of a "helper" object.


knod:
Glad to hear it! And I am hearing that the notes are the same for these two PRs.

PR #328, opened 2017-12-14
Basic consolidation of function calls. Nothing too complex.

knod:
I think a couple names were changed too.

### DIFF
--- a/src/programs/federal/snap.js
+++ b/src/programs/federal/snap.js
@@ -28,8 +28,8 @@ const getSNAPBenefits = function ( client, timeframe ) {
   var finalResult = 0;
   var grossIncomeTestResult   = hlp.getGrossIncomeTestResult( client );
   var netIncomeTestResult     = hlp.getNetIncomeTestResult( client );
-  var maxSnapAllotment        = hlp.getMaxSnapAllotment( client );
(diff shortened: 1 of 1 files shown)


src/programs/federal/snap.js
```diff
@@ -288,39 +267,32 @@ hlp.monthlyNetIncome = function(client) {
...
-  if ( hlp.monthlyNetIncome(client) * SNAPData.PERCENT_OF_NET > 0 ) {
-    return hlp.monthlyNetIncome(client) * SNAPData.PERCENT_OF_NET;
+  var maxNetIncome = hlp.getMaxNetIncome(client)
+
+  if ( maxNetIncome === "no limit" ) {
+    return true;
+  } else if ( hlp.monthlyNetIncome(client) < maxNetIncome ) {
+    return true;
+  } else {
+    return false;
```
wacii: Not a fan of `if(true) return true;`. I always assume the extra code is there for some reason I'm not seeing, and end up rereading it a few times.

But that could just be me.

knod: No, I hear you there. I'm not fixing everything at once otherwise it's going to be hard to examine changes. I think I've already fixed that in one of my more recent edits.

PR #337, opened 2017-12-14
Start of rake tests, #315

knod:
This is the logic testing I was talking about. I know folks aren't fans, but gotta do something within the coming week, and this allows for fully examining edge case behavior. We'll see how long the tests take to run as we build up iterations.

### DIFF
--- a/src/test/programs/federal/expectedSNAPValues.js
+++ b/src/test/programs/federal/expectedSNAPValues.js
@@ -4,12 +4,24 @@
 // testing stuff, and be able to maintain that presence
 // through beta testing, please give me some leeway.
 const expectedSNAPValues = {
-  'with household of __, current earnings of 0': 192,
-  'with household of __, current earnings of 100': 192,
-  'with household of __, current earnings of 1000': 42.900000000000006,
-  'with household of __, current earnings of 5000': 0,
-  'with household of __, current earnings of 10000': 0,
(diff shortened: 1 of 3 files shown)


wacii:
@knod Just to be sure. Is this basically what you were trying to write?

```js
const expected = {
  0: 192,
  100: 192,
  1000: 42.9
};

Object.keys(expected).forEach(earned => {
  const client = cloneDeep(CLIENT_DEFAULTS);
  client.current.earned = earned;
  it(`with household of __, current earnings of ${earned}`, () => {
    expect(getSNAPBenefits(client, 'current')).toBeCloseTo(expected[earned], 4);
  });
});
```

knod:
@wacii: No, actually, not quite. That doesn't account for further permutations. There will be a different result for a household with that earned income that has one member vs. one that has 8 members. I'm just showing the first of many iterators. If it would help, I'll add another one.

wacii:
Ah, I see. So you will be looking at combinations of number of members and earned income. The underscores in the test case names make sense now.

knod:
Another layer has been added, so maybe it's a bit easier to see how this will grow. It adds a bit of utility testing.

knod:
Branches still exist on my repo, but wacii is taking the functionality shown in those to try to pre-build test those test cases instead of generating them at run-time.

PR #339, opened 2017-12-15
Add regression tests for SNAP benefits

wacii:
Use `npm run generateSNAPTests` to create the test cases on disk for the current implementation of `getSNAPBenefits`. These tests are read and run in 'snap.test.js'. All 2304 of them.

I didn't generate combinations of members, that would be too many tests. Instead I just created random members as needed, limiting up to one head and spouse member.

Related to #315


wacii:
There are now more tests. So many tests.

knod:
New version being created.

PR #340, opened 2017-12-16
Implements hack using 'voucher' shelter value, fixes bug in snap 

knod:
while converting snap to the new system.

The state setting is pretty much a hack. Don't have a lock on a system to put together on the fly.

### DIFF
--- a/src/containers/VisitPage.js
+++ b/src/containers/VisitPage.js
@@ -42,6 +42,8 @@ class VisitPage extends Component {
         isBlocking: true,
         redirect: false,
         client : { ...CLIENT_DEFAULTS },
+        // Hack for MVP
+        oldShelter: CLIENT_DEFAULTS.current.shelter,
         userChanged: {}
     };  // end this.state {}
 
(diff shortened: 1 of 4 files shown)


wacii:
@knod Don't know that I can review this. It took me five minutes to figure out how `route` could be 'shelter'. I understood very little after that short journey other than that `route` is a very mysterious property.

But from what I can tell `shelter === 'voucher'` is the same as `hasHousing`, so there's no reason for the `oldShelter` gymnastics? But that is from someone who has no idea what these calculations are ultimately supposed to look like.

knod:
@wacii: Yeah, route is, in my opinion, the most confusing part of that. It basically just defines a path to the correct object and property, but most times the path is only one name long. It seemed the smoothest way at the time.

In my mind, 'voucher' makes calculations, and will make testing, clearer and easier. Neither of them will have to worry about overriding with `.hasHousing`, giving it precedence over `shelter`, when what they're really asking is about client shelter type. Other than that, 'voucher' *is* a shelter type while `.hasHousing` is not.

knod:
It actually occurs to me that we may have to add a fourth 'Other voucher' option to shelter anyway, though maybe they'll feel fine taking care of that with 'Rent'. It'll be an interesting problem if it comes up. We may just be able to call the 'Rent' cash flow input 'Paid rent'. Hmm, thoughts.

PR #344, opened 2017-12-16
Fixes child support showing without SNAP

knod:
Thought there were going to be more changes, but not sure they're worth it without some major changes. Among other things, we can prevent form elements from being shown if benefits are missing, but then form steps would be completely blank at times. In which case the empty steps really shouldn't be available, etc. In other cases, the complexity may just not be worth it at this point.

### DIFF
--- a/src/forms/CurrentExpenses.js
+++ b/src/forms/CurrentExpenses.js
@@ -288,13 +288,14 @@ const ExpensesFormContent = function ({ client, current, time, setClientProperty
         : null
       }
 
-      {/** Wrapper here or else margins get added here, but not other
-      * places, making spacing hard to manage */}
-      <wrapper>
-        <FormHeading>Child Support</FormHeading>
-        <IntervalColumnHeadings type={type}/>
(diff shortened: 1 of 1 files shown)


src/forms/CurrentExpenses.js
```diff
@@ -288,13 +288,14 @@ const ExpensesFormContent = function ({ client, current, time, setClientProperty
...
-        <CashFlowRow {...sharedProps} generic={'childSupportPaidOut'}> <strong>Legally obligated</strong> child support </CashFlowRow>
-      </wrapper>
+      { current.hasSnap
+        ? <wrapper>
+          <FormHeading>Child Support</FormHeading>
+          <IntervalColumnHeadings type={type}/>
+          <CashFlowRow {...sharedProps} generic={'childSupportPaidOut'}> <strong>Legally obligated</strong> child support </CashFlowRow>
+        </wrapper>
+        : null
+      }
```
wacii: You can do `{hasSnap && <div />}`. `false` acts just like `null` here.

knod: Yeah, I've seen that around a bit. Wasn't sure exactly what it was doing or how. I have to say, even knowing that I find that syntax a bit harder to read, so I'm going to stay away from it for now, at least for this project.

PR #345, opened 2017-12-17
Fixes prop name bug (unintended) and cleans up 'Expenses'

knod:
This was meant to just be clean up. Unfortunately, the 6th one down is actually a bug fix. Bit of a bummer.

### DIFF
--- a/src/forms/CurrentExpenses.js
+++ b/src/forms/CurrentExpenses.js
@@ -183,7 +183,7 @@ const Housing = function ({ current, type, time, setClientProperty }) {
   // We're using a bunch of radio buttons. Since `checked` is defined
   // in Radio components, `setClientProperty()` would store it, but we
   // want the value, so get rid of checked.
-  /** Makes sure values are propagated to 'current' properties if needed */
+  /** Makes sure values are propagated to 'current' properties if needed. */
   let ensureRouteAndValue = function ( evnt, inputProps ) {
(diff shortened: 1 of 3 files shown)


src/forms/CurrentExpenses.js
```diff
@@ -168,20 +161,22 @@ const ShelterRadio = function ({ currentValue, label, time, setClientProperty })
...
-* 
-* @returns Component
-*/
+/** 
+ * @function
+ * @param {object} props
+ * @property {object} props.current - Client data of current user circumstances
+ * @property {string} props.type - 'expense' or 'income', etc., for classes
+ * @property {string} props.time - 'current' or 'future'
+ * @property {function} props.setClientProperty - Sets state values
```
wacii: `@property` here refers not to properties on the parameter, `prop`, but on the function `Housing()`. So you are suggesting that `typeof Housing.props.time === 'string'`.

Granted this usage makes sense to people, so if that's all your intending to use it for, it's fine. But if you ever wanted to generate documentation from this, a computer is going to misunderstand you.

src/forms/CurrentExpenses.js
```diff
@@ -168,20 +161,22 @@ const ShelterRadio = function ({ currentValue, label, time, setClientProperty })
...
+ * 
+ * @returns Component
```
wacii: It returns a React element. It is a React component.

PR #348, opened 2017-12-17
Removes unnecessary `client` passing, removes wrong description

knod:
A pretty simple one.


wacii:
@knod This looks familiar. Pretty sure this was done in #345.

knod:
I believe you.

PR #353, opened 2017-12-18
Add unit tests for getMembers utils

wacii:


src/test/utils/getMembers.test.js
```diff
@@ -0,0 +1,131 @@
...
+
+describe('dependents', () => {
+  const dependents = [
+    { m_age: 18, m_role: 'member' },
+    { m_age: 17, m_role: 'member' },
+    { m_age: 19, m_disabled: true, m_role: 'member' }
+  ];
+  const notDependents = [
+    { m_age: 18, m_role: 'head' },
+    { m_age: 18, m_role: 'spouse' },
```
knod: Next iteration, might check disabled head and spouse too.

src/test/utils/getMembers.test.js
```diff
@@ -0,0 +1,131 @@
...
+  const notDisabled = { m_disabled: false };
+  const allMembers = [disabled, notDisabled, notDisabled];
+
+  test('isDisabled()', () => {
+    expect(isDisabled(disabled)).toBe(true);
+    expect(isDisabled(notDisabled)).toBe(false);
+  });
+
+  test('getDisabledMembers()', () => {
+    expect(getDisabledMembers(allMembers)).toHaveLength(1);
```
knod: What happens if there are no disabled members? I don't even remember what's supposed to happen.

src/test/utils/getMembers.test.js
wacii: Added them.

src/test/utils/getMembers.test.js
wacii: I added a test to `getEveryMember()` for this. But it's an empty array.

Issue #354, opened 2017-12-18
Unit tests of snap helper functions


knod:
More people can join in if they wish.
- [x] `SNAPhelpers.householdSize()`
- [x] `SNAPhelpers.isElderlyOrDisabled()`
- [x] `SNAPhelpers.hasDisabledOrElderlyMember()`
- [x] `SNAPhelpers.hasDependentsOver12()`
- [x] ~`SNAPhelpers.getChildSupportPaid()`~
- [x] `SNAPhelpers.getAdjustedGross()`
- [x] `SNAPhelpers.getPovertyGrossIncomeLevel()`
- [x] `SNAPhelpers.getGrossIncomeTestResult()`
- [x] `SNAPhelpers.isHomeless()`
- [x] `SNAPhelpers.getNonUtilityCosts()`
- [x] `SNAPhelpers.getUtilityCostByBracket()`
- [x] `SNAPhelpers.getTotalshelterCost()`
- [x] `SNAPhelpers.getStandardDeduction()`
- [x] `SNAPhelpers.getEarnedIncomeDeduction()`
- [x] `SNAPhelpers.getMedicalDeduction()`
- [x] `SNAPhelpers.getDependentCareDeduction()`
- [x] `SNAPhelpers.getHalfAdjustedIncome()`
- [x] `SNAPhelpers.getRawShelterDeduction()`
- [x] `SNAPhelpers.getShelterDeduction()`
- [x] `SNAPhelpers.getHomelessDeduction()`
- [x] `SNAPhelpers.getAdjustedNotGrossIncome()`
- [x] `SNAPhelpers.monthlyNetIncome()`
- [x] `SNAPhelpers.getMaxNetIncome()`
- [x] `SNAPhelpers.getNetIncomeTestResult()`

knod:
I think 4810b2d closed this!

PR #355, opened 2017-12-18
Snap unit tests/bug fixes, part 1, #354

knod:


src/test/programs/federal/snap/unit.test.js
```diff
@@ -0,0 +1,349 @@
...
+    });
+  });
+
+
+  // `SNAPhelpers.getChildSupportPaid()`
+  describe('`.getChildSupportPaid( timeClient )` given a time-constrained client paying', () => {
+    let current;
+    beforeEach(() => { current = cloneDeep( defaultCurrent ); });
+
+    describe('no ', () => {
```
wacii: 'no one' ?

src/test/programs/federal/snap/unit.test.js
```diff
@@ -0,0 +1,349 @@
...
+      it('that is not disabled and under 60 should return false', () => {
+        expect(SNAPhelpers.hasDisabledOrElderlyMember( current )).toBe(false);
```
wacii: We're testing `.getChildSupportPaid()` here.

knod: Don't know why this is even there - the function is gone. Thanks for the catch!

PR #364, opened 2017-12-20
added more tests for all housing.js functions

maxsean:


maxsean:
Thanks Don,
I am following the conventions in src/test/programs/federal/snap/unit.test.js. I think it is primarily a styling to separate a function name from other text in a string.

donblair30:
Ok thanks for the explanation.  I'm not sure it's doing anything, but I
also don't think it hurts anything.
Change reviewed and merged.  All set!

Don

PR #365, opened 2017-12-20
Don't modify snap data in getMedicalDeduction

wacii:


knod:
Shoot, thought I had changed that in an old version. I guess it just never made it through. Thanks!

PR #371, opened 2018-01-02
[WIP] Download session data for support requests

wacii:
Add me somewhere:
```
<DownloadSessionData client={client} />
```
This creates a download link to a JSON file populated with the client object, the last fifty actions taken by the user, and the last uncaught error with stack trace.

Not sure what exactly the feedback/support request feature is going to look like, but I figure much of this code will be relevant regardless.


wacii:
Edited a couple of the commits. There was no reason for the HOCs to set `displayName`.

wacii:
Removed `recordActions()` and updated `recordErrors()` to open a prompt when an uncaught error is detected.

Can be added anywhere like so:
```js
<DownloadErrorPrompt />
```

The two file downloads are currently separate, but that can be changed if need be.

wacii:
This batch of commits was almost entirely concerned with adding a download link to the 'are you sure you want to leave' prompt.

The code is pretty miserable. While I'm sure I could improve it with subsequent passes, I'm working against the browser here, so I don't think it'll ever be all that great.

While I did add client data to the error download, I didn't change the prompt to a popup. This could turn even harmless errors into serious ones. Imagine a popup being launched on every keystroke. The app would be unusable.

There are various solutions to this. For instance, we could interpret closing one file download popup to mean the user does not want to see any further popups that session.

wacii:
Not sure why I removed the `<ConfirmLeave>` component, but I added it again. Making a few changes now that I don't need to access the event handler from the reset button.

wacii:
Also fixes bug where client state wasn't being passed to `<OnLeavePrompt>`.

wacii:
Errors now trigger a the same prompt that attempting to reset the app does.

Small changes to the prompt itself. The download link is no longer one of the modal actions, but part of the modal body. This provided an opportunity to introduce the link.

Various refactorings.

Issue #297, opened 2017-12-08
Possible pretty solid way to get data for testing as we do the beta


knod:
So I'm not sure how to do this, but this would be my dream scenario:

1. Beta testers fill out the data for a situation that already exists and for which they already know the past (our 'current') and present (our 'future') values.
2. Beta testers fill out the actual values.
3. We get sent both the data and the values that go with it.

Instant data to check our calculations against. Is that even a possibility we can consider?

knod:
For the SNAP tests, we wouldn't even need two sets of data, we'd just need their current data and current benefits.

knod:
This is mostly taken care of in those and other commits, but we don't yet have a form that pre-fills information and queues the user to give us useful responses (like the amounts they expected and any additional data that may have caused the discrepancy).

knod:
See #423

PR #375, opened 2018-01-15
Tweaks text of 'prompts', different text based on situation. #297.

knod:


knod:
@wacii: Nice catch, thanks

src/containers/VisitPage.js
```diff
@@ -70,8 +70,8 @@ class VisitPage extends Component {
 
   componentDidMount() {
     const data = { client: this.state.client };
-    const confirm = (message, callback) =>
-      this.prompt(callback, data, message);
+    const confirm = (message, callback, leaveText, header) =>
```
wacii: This won't cause a problem, but it isn't technically right. It is React Router, or more specifically [history](https://github.com/ReactTraining/history), that will be calling this function. So we can't change its method signature.

The default confirm function is [here](https://github.com/ReactTraining/history/blob/master/modules/DOMUtils.js#L7).

PR #387, opened 2018-01-24
Wording

knod:
The stuff we discussed with Kristin plus the change to 'Graph' form section name.


knod:
@wacii: Yeah, it's a good point. I think the fact that it's there is a start and we'll work on improving it later in whatever ways are needed.

PR #391, opened 2018-01-26
Fixes #381, adds fixed 'Download data' button to all form sections

### LINKED ISSUE #381: Add sticky 'Feedback' button (discuss?)
Right now the user will be prompted to download data and send feedback when they go to a different page or reset the form. Should there be a way for them to do this whenever they want? That would make sense to me.

knod:


wacii:
`<a>` elements are not supposed to have interactive children, like a `<button>`. Browsers might not care, I know it works in Chrome, but I think it would be better not to risk it.

There is an example of how to render the Download like as a button in `<OnLeavePrompt>`.

knod:
@wacii: I'll try to catch you online. The styling got very strange when I attempted that. Good point, though.

knod:
For now we're keeping it this way. When we have a chance, we can look into how we might be able to alter `<DownloadFile>` in order to get a result that's styled correctly and doesn't use this hack.

Issue #399, opened 2018-01-29
Download data appears after the user agreed to "Reset" data.


kalpana-k:
### Images:
![issue1](https://user-images.githubusercontent.com/4665716/35515340-e238bc4a-04d6-11e8-8aae-07022cf08220.JPG)
![issue 2](https://user-images.githubusercontent.com/4665716/35515347-e7989462-04d6-11e8-8c68-8007db7398dc.JPG)

### Replicate:
1. Click on "Reset" button on "Graphs" screen.
2. Application prompts user to Reset/Stay/Download Data.
3. Click on Reset on the prompt.
4. Application directs user to Home page.
5. Click on About
6. Application prompts user to Leave/Stay/Download Data

### Result I'd Expect:
- Once, a user hits Reset at Step 3. above the user should not be prompted to again at Step 6.

### Actual Result:
- User is prompted to download data when they agreed to Reset the form.

### Environment:
Operating system and version: 
Browser type and version: 

App version or commit id at the time: 

### Other Notes
Not sure if this is a major issue. I could see when this prompt is helpful. Eg: User fills in data and is on the "Graphs" screen and by accident hits the "Home/About" on the Menu.

knod:
Fixed by #401

PR #401, opened 2018-01-29
Fixe #399, user gets download prompts without form interaction

knod:


wacii:
@knod When the client is reset, `isBlocking` should be set to false.

kalpana-k:
pulled the branch and tested. i was able to replicate the issue

PR #406, opened 2018-01-31
Validate rent share

### LINKED ISSUE #388: Rent share higher than contract rent value creates negative result
### Replicate:
1. In 'Benefits' form section, activate 'Section 8'.
2. In 'Expenses' form section, put in $1 for 'Rent Share'.
3. Go to 'Predictions' form section.

### Result I'd Expect:
- Not sure. Validation error on the 'Expenses' page at the very least. Maybe I'd be prevented from leaving that form section.

### Actual Result:
- In the 'Predictions' section, the current value for Section 8 is negative.

### Environment:
Operating system and version: OSX v10.11
Browser type and version: v62...
App version or commit id at the time: e6d5e503adb4cfb1e588d2b97a8eb2b06315384a

### Other Notes:
For Section 8, 'current' values are calculated completely differently from 'future' values, and they're not screened to make sure they don't come out to negative numbers. This should stay the same. What needs to change is the validation on the form's side of things where the user is entering the information.

wacii:
Rent share should be less than contract rent. This is now enforced, with error messages being displayed so the user knows why we are potentially throwing away their input.

Closes #388

### DIFF
--- a/src/forms/CurrentExpenses.js
+++ b/src/forms/CurrentExpenses.js
@@ -15,6 +15,7 @@ import {
   IntervalColumnHeadings,
   CashFlowRow
 } from './formHelpers';
+import { ContractRentField, RentShareField } from './rentFields';
 
 // COMPONENT HELPER FUNCTIONS
 import { getTimeSetter } from '../utils/getTimeSetter';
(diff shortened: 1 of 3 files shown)


wacii:
There is a slight visual bug. When the error message is added, the label moves slightly. This is because we are currently misusing the field styling, fixing it would mean writing new CSS specifically for the control.

knod:
Looks good!

Issue #41, opened 2017-09-20
MOD: Making data visualization more useful


zzyyfff:
We should consider changing the type of graph to a "stacked area graph" that combines the client's income as well as all benefit programs.

see: http://www.datavizcatalogue.com/methods/stacked_area_graph.html

zzyyfff:
Example-

![image](https://user-images.githubusercontent.com/7986137/30622741-8feed454-9d81-11e7-983c-332f6b2027d1.png)

knod:
I think this question is now for @anhase and the Project Hope folks to tell us the kind of feedback they want to be able to show clients and/or investors (of whatever flavor).

anhase:
Here's what Kristin had to say yesterday: "As for the final results page, I
think the graph you guys have was probably an attempt to mimic the graph on
page 17 of this CSP document:
https://www.umb.edu/editor_uploads/images/centers_institutes/center_social_policy/Rock_and_a_Hard_Place_Sept_2016.pdf.
I agree that it's difficult to read. I think there are 2 different data
visualizations that would be helpful to have if they're not too difficult
to develop. One would be a "before" and "after" bar graph very similar to
the table you guys have, where the y axis represents total net resources
(income + benefits) and there is 1 bar for "before" and 1 bar for "after."
Each bar would consist of 2 subsections: earned income (in 1 color) and
benefits (in another color). The second visualization that might be helpful
is just some indication of where the family's new/proposed income lies in
relation to the eligibility drop-offs for each benefit. I'm picturing
something almost like a timeline (you're HERE; you'll no longer be eligible
for Section 8 HERE; you'll no longer be eligible for SNAP HERE)."

knod:
Ah, ok, so the first is like the stacked bar chart I had talked about, but only for the two given values. The second one, though, sounds like the graph we have except for with a 'you are here' marker. It shows where the benefit fall to 0. Does that sound right? Do you think she wants them separated out into a graph for each program? Or does she mean something else?

I'm not sure how difficult they are to develop. @sashamaryl has had a look at chart.js, etc., and might have a better idea.

knod:
@anhase: Or are you just talking about having a number value. 'You would lose SNAP at income of $500/month'.

sashamaryl:
[...]

PR #407, opened 2018-01-31
#41, graph clarity, add vertical line for future income

wacii:
Added to both charts on the graphs page. There is no tooltip for the line. I could add one, but the code would be pretty hideous.

### DIFF
--- a/src/forms/ResultsGraph.js
+++ b/src/forms/ResultsGraph.js
@@ -63,6 +63,35 @@ const ResultsGraph = (props) => {
     return yearlySubsidy;
   });
 
+  const income = props.client.future.earned * 12;
+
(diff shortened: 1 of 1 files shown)


knod:
This is great! Thank you! It might be good if we could at least add a label to it, but I'm not familiar with that functionality. Note to self: we may end up using the 'current' income value.

wacii:
@knod I could add a label if you want.  It is easy to do.  There was one before I just didn't like the look of it.

And if you want to play around with the implementation, `ctx` is a [CanvasRenderingContext2D](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D), which is part of the browser API, not _Chart.js_. It took me awhile to realize that...

PR #409, opened 2018-01-31
Fixes #89, allow graphs to be shown by weekly and monthly income amounts

### LINKED ISSUE #89: Allow chart to show income by other timeframes
(Not for MVP) Right now the result chart/graph's ticks are annual income amounts. Considering clients will probably be putting in weekly amounts, presumably because they're more familiar with them, it might be more useful to the client if we initially show the tick mark values as weekly amounts. We can probably find a way to allow the user to switch to monthly or yearly amounts too.

knod:


knod:
I've gotten verbal consent from enough people that I'm letting this through.

PR #410, opened 2018-02-01
Add label to vertical line annotation

wacii:
Spent a little time trying to get this to look good. Was not very successful. PRing this mostly as a code example for anyone who wants to take a crack at it.

note: Splitting text on different lines is definitely tedious using 2d context.


wacii:
Maybe just using the income value as a label. The user would understand the relevance of the number, and it wouldn't add much by way of clutter.

knod:
@wacii: Hmm, that's a decent thought, though honestly I think it also looks good now. Our liaison at Project Hope said that if possible, she'd like lines for both current and future values. We don't have to do that right now, but something to think about if/when implementing that is how to keep labels from overlapping. I have some thoughts.

PR #412, opened 2018-02-03
Only shows active programs in graphs

knod:
Via setting up a flexible scheme for adding data to the graphs.


wacii:
The only thing that jumps out at me is that there is no reason to show a graph at all if no programs have been selected. Not sure if that's worth addressing.

Otherwise everything appears to work. I don't understand all the code, it would take me a few passes to get everything. Not sure that's really necessary for a chart though, as it is so self-contained. As long as it works.

PR #413, opened 2018-02-04
With special address, can load client data from file

knod:
I'd like it to be use-able by non-coding collaborators too.


wacii:
I didn't know you could do `"foo".indexOf("oo")`. Probably faster than doing regex.

knod:
@wacii: */load* and */visit/load* is a bit of an assumption on my part. Having been in a similar state of mind myself often enough, I suspect that folks, especially non-coders, would have trouble remembering the exact correct address to use. I'm seeing if being a bit flexible about it will help smooth things out.

Issue #404, opened 2018-01-30
Consolidate getting current and future income and program values


knod:
In the 'Predictions' form section, our table and our chart are both getting user current and future values, but they're doing it separately. It'd be good to figure out a way to pull these up into the parent, abstract them somewhere, or something else, so that we're not duplicating work to improve maintainability.

migueldcepeda:
Fixed by #422

PR #416, opened 2018-02-19
Consolidate program and income values on predictions page (Issue #404)

migueldcepeda:
Wasn't sure if the best place to refactor this was in the 'getTimeSetter.js' helper file in the utils directory but it seemed related.


knod:
We looked into the p-lock thing more and apparently we need an update. Am taking care of that in #418.

knod:
Changed in #422 instead without a package-lock.json change.

PR #418, opened 2018-02-21
Update package-lock for some reason

knod:


knod:
@wacii: I thought that problem was fixed in v5.4 and up or something.

knod:
Do we all need to be on the same version of node? We can try to make that happen.

wacii:
@knod Going by the discussion [here](https://github.com/npm/npm/issues/17749), there seemed to be a number of attempts to fix the problem, but people are still reporting issues, so I'm not sure it has actually been resolved.

I don't think this has to do with node. Although now that I think about it, I'm not sure why people don't specify a version for node in the _package.json_ or something, even if it would just be used in development.

Issue #424, opened 2018-02-28
Change default age for non-head, non-spouse family members


ethanbb:
Instead of defaulting all ages to 30, would make sense to default to something like 12, for the common case where other members are children.

PR #432, opened 2018-03-03
Change default age of non-spouse/non-head family members to 12

ethanbb:
Addresses issue #424, to address a slight point of annoyance by defaulting non-head, non-spouse family members to a younger age. I think someone said 12 for children was good? But this can be changed.

(As for the other changes - I think Git for Windows's auto-line ending handler automatically deleted some extra spaces in the comments (line ending characters were NOT affected). Someone with a different computer can check but I'm pretty sure it did the right thing.)

### DIFF
--- a/src/forms/Household.js
+++ b/src/forms/Household.js
@@ -66,11 +66,11 @@ Columns.Four = function ({ noMargin, children }) {
 
 
 /** @todo description
-* 
+*
 * @function
 * @param {object} props
 * @property {object} props.__ - explanation
-* 
+*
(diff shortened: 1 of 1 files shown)


ethanbb:
Maybe. I guess that would mean changing it back to 30 if they changed it back to spouse (and the age is still untouched). I could see it being either helpful or slightly more confusing; can someone with more UX experience i.e. @kalpana-k  weigh in?

knod:
You know what, we'll go with this for now and see how it goes.

PR #433, opened 2018-03-04
Fixes #431, Rent fields to only monthly values

### LINKED ISSUE #431: Make Rent questions only have a monthly money input field
Requested by Project Hope. Also, change questions to include the word 'monthly' in them?

knod:
Also changes abstractions and refactors. Would appreciate another eye on it for a sanity check.

If you're looking at this fresh, I'd suggest you not dig too much into the diff. Get a feel for the diff and then look at the new relevant code to see what you think. If you feel ambivalent about it, look at the old relevant code and see how it compares.

Or possibly just test out behavior a bunch. Good behavior may be good enough.

### DIFF
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -402,7 +402,7 @@ class ManagedNumberField extends Component {
   }
   
   handleBlur = ( evnt ) => {
-    this.props.updateFieldError(false)
+    this.props.updateFieldValidity(false)
     this.setState({ focused: false, valid: true });
   }
 
@@ -415,7 +415,7 @@ class ManagedNumberField extends Component {
       this.props.store( evnt, inputProps, this.props.otherData );
     }
 
-    this.props.updateFieldError(!valid);
+    this.props.updateFieldValidity(!valid);
     this.setState({ valid: valid, focusedVal: value });
   }  // End handleChange()
 
@@ -479,7 +479,7 @@ const CashFlowRow = function ({ generic, timeState, setClientProperty, children,
         store:    updateClient,
         validate: isPositiveNumber,
         format:   toMoneyStr,
-        updateFieldError: function () {}
+        updateFieldValidity: function () {}
       };
 
   return (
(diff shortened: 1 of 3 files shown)


wacii:
There is a change in behavior. Now negative rent values can trigger the `"Rent share must be less than contract rent"` message.

wacii:
Also `updateFieldValidity()` is a little misleading, as a name, because you are setting the `invalid` property. So `updateFieldValidity(false)` actually means that there's no error.

knod:
Speaking of the field validity, we've got something going on here that I'm not sure about. `<ManagedNumberField>` already has a validity state and now the container of a specific row, like `<ContractRentField>`, also has a validity state. I've already run into a couple of issues with this redundancy.

knod:
@wacii, @shp54: Refactored, renamed some, and, I hope, fixed behavior, though altered the behavior just a touch. Need a second eye on it, though.

src/forms/formHelpers.js
```diff
@@ -471,52 +490,61 @@ const CashFlowRow = function ({ generic, timeState, setClientProperty, children,
...
+        className:  'cashflow-column',
+        store:      updateClient,
+        validation: isPositiveNumber,
+        format:     toMoneyStr,
+        onBlur:     function () { return true; }
       };
 
   return (
-    <Form.Field inline className={'cashflow'}>
+    <CashFlowContainer label={children} validRow={validRow || true} message={message || null}>
```
wacii: `validRow || true` will always be truthy.

wacii: The only reason this doesn't cause a problem is because `<RowMessage>` checks both `validRow` and `message` before showing the label.

knod: It will right now. I was making this in hopeful anticipation of getting rid of `<MonthlyCashFlowRow>` and adapting plain old `<CashFlowRow>` to be more flexible. Nothing I've worked out so far works very well for that, though. As for `<RowMessage>` catching it, if `validRow === true`, no message would show anyway.

Still, may be time to get rid of it.

knod: I think `validRow` will still always need to be true for the sake of clarity.

wacii: Reviewing the code I was a little off base as this isn't actually a problem, even if `val || true` is a little worrying. I guess I thought this was being used by `<RentShareField>`, but that's the other one.

Issue #427, opened 2018-02-28
Move reset and download button and change color of reset


nicklee100:
add reset button below download and make them vertical.   Make it look like a tab.

> "vertical"

knod:
Not quite 'make it look like a tab' situation. Imagine the buttons are rotated 90 degrees to the left.

knod:
I'll call this closed. We'll look into color choices separately. Thanks!

PR #436, opened 2018-03-13
Rotate download button (Issue #427)

migueldcepeda:
Roted 'Download Data' Button vertically. Also hollowed it out to mesh into the background.

### DIFF
--- a/src/components/DownloadAnytime.js
+++ b/src/components/DownloadAnytime.js
@@ -10,10 +10,10 @@ const DownloadAnytime = function ( props ) {
 
   return (
     <DownloadFile data={{client: props.client}}>
-      <Button type='button' color='teal' size='medium' className={'fixed'} id={'download_data_fixed'}>
+      <Button type='button' color='teal' size='medium' id={'download_data_fixed'}>
         Download Data
       </Button>
-    </DownloadFile>   
+    </DownloadFile>
   );
 
 };  // End <DownloadAnytime>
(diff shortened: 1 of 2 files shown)


src/components/DownloadAnytime.js
```diff
@@ -10,10 +10,10 @@ const DownloadAnytime = function ( props ) {
 
   return (
     <DownloadFile data={{client: props.client}}>
-      <Button type='button' color='teal' size='medium' className={'fixed'} id={'download_data_fixed'}>
+      <Button type='button' color='teal' size='medium' id={'download_data_fixed'}>
```
knod: I think the 'Reset' button will also be rotated and fixed. Could we add a rotated class that would do for both of those?

src/index.css
```diff
@@ -171,3 +172,26 @@ input[type=number]::-webkit-outer-spin-button {
...
+#download_data_fixed {
+    transform: rotate(90deg);
+    /* background-color: red; */
+    background-color: #fff;
+    color: #1CB5AC;
+    border-left: solid 1px #1CB5AC;
+    border-top: solid 1px #1CB5AC;
+    border-right: solid 1px #1CB5AC;
+    position: fixed;
+    top: 20%;
```
knod: We'd discussed changing the horizontal alignment to vertical. I think it would be good to make it's top-most edge align with its previous topmost edge so that it doesn't go into the header.

src/index.css
```diff
@@ -171,3 +172,26 @@ input[type=number]::-webkit-outer-spin-button {
   top: 2em;
   right: 0;
 }
+
+/* DOWNLOAD DATA BUTTON */
+#download_data_fixed {
+    transform: rotate(90deg);
+    /* background-color: red; */
+    background-color: #fff;
```
knod: Let's talk about the color scheme change in real-time.

PR #439, opened 2018-03-16
Fixes #428, toggles and text change for current benefits question

### LINKED ISSUE #428: Change text and format of 'Current Benefit' questions
We can try y/n radio buttons and text that says 'Are you currently getting ___' where '___' is the name of the benefit.

nicklee100:
- Added a class based component for a 'Yes No' toggle radio button 

- being used now for 'under currnet benefits' 

- Added css in index.css for styling

### DIFF
--- a/src/forms/CurrentBenefits.js
+++ b/src/forms/CurrentBenefits.js
@@ -22,7 +22,6 @@ import { getTimeSetter } from '../utils/getTimeSetter';
 const CurrentBenefitsContent = ({ current, setClientProperty }) => {
 
   let setChecked = function ( evnt, inputProps ) {
-    inputProps.name = inputProps.questionname;
     inputProps.checked = inputProps.label === 'yes' ? true : false;
     var obj = { ...inputProps, value: inputProps.value };
     setClientProperty( evnt, obj );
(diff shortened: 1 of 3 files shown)


src/forms/CurrentBenefits.js
```diff
@@ -3,50 +3,49 @@ import React from 'react';
 import { Form } from 'semantic-ui-react';
 
 // PROJECT COMPONENTS
-import {
-  FormPartsContainer, MassiveToggle
-} from './formHelpers';
+import { FormPartsContainer, MassiveToggle, ControlledRadioYesNo } from './formHelpers';
```
knod: Getting warnings in the console. Travis is set up to not allow those.

src/forms/formHelpers.js
```diff
@@ -538,6 +539,51 @@ const MonthlyCashFlowRow = function ({ inputProps, baseValue, setClientProperty,
...
+    super(props)
+    this.state = {
+    }
+  }
+
+  onChange(e,inputProps){
+    inputProps.checked = inputProps.label === 'Yes' ? true : false;
+    var obj = { ...inputProps, value: inputProps.value };
+    this.props.setClientProperty( e, obj );
+  }
```
shp54: should be renamed to `this.props.onChange`, because you could reuse it in cases where you're not setting a client property.

knod: @shp54: Hmm, this would require a deeper change where this prop is passed in. May be worth it in the long run, but may be a better long term goal, not an immediate issue.

shp54:
Looks good to me!

PR #442, opened 2018-03-18
Fixes #441, explanation of SNAP being calculated

### LINKED ISSUE #441: Include a line about SNAP calculations in the 'Income' section
Just to say that we'll be calculating SNAP.

knod:


knod:
I've gotten the go-head on this one, so I'm merging it.

Issue #423, opened 2018-02-27
Feedback as embedded google form or something else more useful?


knod:
I think we could find a way to attach a file, or include the data in some other way, but I'll have to look into it.

ethanbb:
Update: we are going to try creating our own form instead of using a Google form. Then we will either use a tiny server to send the form data over email, or add it to a Google spreadsheet using Apps script.

ethanbb:
Branch with development of form and request to send data to a google spreadsheet: https://github.com/ethanbb/cliff-effects/tree/feedback-form

knod:
Fixed with #465

PR #444, opened 2018-03-20
Addresses #423 Implement basic feedback form as modal

ethanbb:
Addresses #423 - adds feedback form, accessible from the Results page, which submits the data to our google spreadsheet (https://docs.google.com/spreadsheets/d/15LyR9yELAfcngj-c7vMdI630b6DwuuXV-dQEJDvU4gE/edit?usp=sharing).

The styling etc. could use some work from someone who knows what they're doing, but it seems to be functional!


src/components/FeedbackPrompt.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+* Form for submitting user feedback.
+*/
+class FeedbackForm extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      submissionFailed: false,
+    };
```
wacii: This state isn't used.

ethanbb: Actually it is used on line 144.

wacii:
This should probably live in a separate branch. Beyond styling changes, we should probably convert the other methods of feedback to use google forms. As well as update the import client data component. And all that probably shouldn't be contained within a single PR.

src/components/FeedbackPrompt.js
wacii: That is `<FeedbackPrompt>` state, not `<FeedbackForm>` state, isn't it?

ethanbb: Oh you're right, sorry, that's leftover from before `<FeedbackPrompt>` existed.

src/containers/VisitPage.js
```diff
@@ -244,6 +251,11 @@ class VisitPage extends Component {
           prompt={this.prompt}
         />
         <ConfirmLeave isBlocking={this.state.isBlocking}/>
+        <FeedbackPrompt
+          open={this.state.feedbackOpen}
+          callback={() => { this.setState({ feedbackOpen: false }); }}
```
shp54: Could rename to something like `close`

ethanbb:
@wacii Can we put it in a separate branch, as is, and then make separate pull requests into there for the other issues you mentioned? It sounds like these are things we should do at some point to clean up our feedback-related data flow, but not directly needed for the feedback submission functionality.

src/containers/VisitPage.js
ethanbb: Yeah good point, I'll do that

src/forms/ResultsGraph.js
```diff
@@ -361,6 +364,8 @@ const ResultsGraph = ({ client, previousStep, resetClient }) => {
       >
         <GraphHolder client={client} Graph={GrossGraph} />
         <GraphHolder client={client} Graph={BenefitGraph} />
+
+        <FeedbackForm />
       </FormPartsContainer>
```
shp54: Are you sure this still exists?

ethanbb: Thanks, forgot to change that back.

knod:
@wacii, @ethanbb: I'll set up a new branch and we can make the PR into there and continue working on this features. Now that this is live, everything is different!

knod:
@wacii, @ethanbb: New feedback branch created. If it's possible to transfer a PR, do so. If not, I guess just make a new PR there. I haven't checked it out yet, and I will do, but I also trust you folks if someone wants to merge it there.

Also, I think, in like with what @wacii was suggesting, we can replace our 'Download Data' button with one that says 'Feedback' and tie this whole operation to it. My bad for being confusing about that, @ethanbb. All I can say is that I wasn't thinking clearly at the time.

src/components/FeedbackPrompt.js
```diff
@@ -0,0 +1,143 @@
...
+      formData: {},
+      submissionFailed: false,
+      submitting: false
+    });
+    this.props.close();
+  }
+
+  submit = (event) => {
+    this.setState({ submitting: true });
+    const data = Object.assign(this.state.formData, { clientData: this.props.data });
```
wacii: `Object.assign({}, this.state.formData, { clientData: this.props.data });`

Don't mutate state outside of `setState()`. I can't think of anything in particular this might mess up, but still better not to do I think.

src/components/FeedbackPrompt.js
```diff
@@ -0,0 +1,143 @@
...
+  },
+  bugReport: {
+    inputType: Form.TextArea,
+    text: 'If there was a bug or error, describe the bug and what you were trying to do when the bug happened.'
+  },
+  comments: {
+    inputType: Form.TextArea,
+    text: 'Do you have any other comments?'
+  }
+};
```
wacii: It would take less space and be clearer and more flexible if you just wrote out the components. If you want to keep the `FeedbackPrompt.render()` from getting too big, you could always just extract out the question fields into a component.

src/components/FeedbackPrompt.js
```diff
@@ -0,0 +1,143 @@
...
+  // returns promise that succeeds if submission is successful, else rejects.
+  sendDataToSpreadsheet(data) {
+    const fetchOptions = {
+      method: 'POST',
+      body: JSON.stringify(data)
+    };
+    return fetch(postUrl, fetchOptions)
+    .then((response) => {
+      if (response.ok) {
+        return true;
```
wacii: This should resolve to something meaningful, like `response` or the processed response body, or nothing at all, `undefined` or `null.

src/components/FeedbackPrompt.js
```diff
@@ -0,0 +1,143 @@
...
+      submitting: false
+    });
+    this.props.close();
+  }
+
+  submit = (event) => {
+    this.setState({ submitting: true });
+    const data = Object.assign(this.state.formData, { clientData: this.props.data });
+    this.sendDataToSpreadsheet(data)
+    .then((result) => {
```
wacii: This promise isn't resolved with a result. Just a random boolean. `sendDataToSpreadsheet()` should be changed and/or this parameter removed or renamed or something.

src/components/FeedbackPrompt.js
```diff
@@ -0,0 +1,143 @@
...
+          <Form>
+            {this.getInputs()}
+          </Form>
+          <Message hidden={!this.state.submissionFailed} error>
+            Error submitting data, please try again or <a href="mailto:andrew@codeforboston.org">email us</a>.
+          </Message>
+        </Modal.Content>
+        <Modal.Actions>
+          <Button onClick={this.cancel}>Cancel</Button>
+          <Button onClick={this.submit} loading={this.state.submitting} primary>Submit</Button>
```
wacii: When submitting this button should probably be disabled. To avoid double submissions.

src/components/FeedbackPrompt.js
ethanbb: My bad about this, I didn't realize Object.assign modifies the first argument.

src/components/FeedbackPrompt.js
ethanbb: OK, I agree actually. I did it this way to avoid repeating the same common properties in each input component, but I just realized you can use {...props} spread notation to factor them out.

src/components/FeedbackPrompt.js
ethanbb: I don't think we need the response for anything, but I guess may as well return it in case we want to use it for anything in the future.

src/components/FeedbackPrompt.js
ethanbb: Yes true

src/components/FeedbackPrompt.js
```diff
@@ -0,0 +1,143 @@
...
+        <Modal.Content>
+          <Form>
+            {this.getInputs()}
+          </Form>
+          <Message hidden={!this.state.submissionFailed} error>
+            Error submitting data, please try again or <a href="mailto:andrew@codeforboston.org">email us</a>.
+          </Message>
+        </Modal.Content>
+        <Modal.Actions>
+          <Button onClick={this.cancel}>Cancel</Button>
```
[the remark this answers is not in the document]
ethanbb: I was thinking actually this is good, since in case the submission is taking a really long time for some reason you wouldn't want to trap the user with the modal open. Closing the modal shouldn't interrupt the submission; the only unexpected behavior I can see happening is submissionFailed gets set after the modal is closed and shows up if it's opened again.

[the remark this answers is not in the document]
knod: If the purpose of that is to allow someone to close the modal when the submission is being sent, then maybe the text for the button should change to 'Done' or 'Close'. Otherwise people will think they're canceling the submission. I suspect they'd end up making the same submission multiple times trying to get it to work.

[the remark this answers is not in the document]
ethanbb: Yeah, I guess. I don't know if we want to be encouraging people to close the modal before they've confirmed that the submission was successful, since if it fails, they can try to resubmit if the modal is still open. If it's taking that long to submit that they decide to click cancel, it was probably failing anyway, so they probably should try to resubmit. 

Maybe we could disable the "cancel" button, but have an x in the corner that doesn't get disabled. And maybe ideally there could be a component visible in the corner or something after closing that reports whether the submission was successful (maybe in the same place as the button to submit feedback, if we decide on a place for that that's visible from all pages?).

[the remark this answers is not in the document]
knod: I like the idea of an 'x' in the corner. Is there a chance the 'cancel' button *could* be made to, in actual fact, cancel a submission while it's being submitted, in case we want them to actually be able to halt the process? We should discuss more about the flow in person, including the successful submission bit. For now in this PR, red 'x' in the corner and 'Cancel' gets disabled when 'Submit' is clicked sounds good to me.

Also, see my very important Slack question.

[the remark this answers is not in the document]
ethanbb: >Is there a chance the 'cancel' button could be made to, in actual fact, cancel a submission while it's being submitted, in case we want them to actually be able to halt the process?

Probably not, since once the request is sent and we're waiting for a response from the Google server, we have no knowledge or control over whether the request will be successful (and if it is, there's no way to undo it (at least without things getting really messy)).

Sounds good re what to do for this PR!

wacii:
Looks good. If no one has further requests by tomorrow I'll merge it. This is a lot better than downloading the JSON data and attaching to email. Thanks for doing this.

knod:
Yeah, very into this 👍

PR #445, opened 2018-03-20
Add specialized EarnedBecauseOfChildcare component

### LINKED ISSUE #437: Increase visibility of "Income made possible by child care expenses"
Field in "expenses" form that is visible when the household contains a child under the age of 13.
We said we would put this in a new section to try to increase its visibility, since it has a major impact on the future income calculation.

wacii:
This field was being routinely ignored during user testing, so now instead of immediately asking the user for the value of the field, we first ask if childcare has allowed the user to make additional income. The hope being that the break in rhythm will draw more attention.

Does the radio button look off to anyone? The dot in a checked radio button looks too square to me. Not particular to this component, but still.

Closes #437

### DIFF
--- a/src/forms/CurrentExpenses.js
+++ b/src/forms/CurrentExpenses.js
@@ -20,6 +20,7 @@ import {
   RentShareField,
   PlainRentRow
 } from './rentFields';
+import EarnedBecauseOfChildCare from './EarnedBecauseOfChildCare';
 
 // COMPONENT HELPER FUNCTIONS
 import { getTimeSetter } from '../utils/getTimeSetter';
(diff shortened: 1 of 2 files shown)


knod:
@wacii: We also talked about breaking it out into its own section, and I think that would be a really good idea. We can make that a separate PR if that would work better.

I haven't noticed anything about square radio buttons... pic?

knod:
We also need to do this for the other 'earned because of' items.

wacii:
@knod I forgot the childcare expenses field was only one of several. Oops. I'll make the component more generic.

PR #455, opened 2018-03-27
Prompt before accepting other "earned because of" incomes

### LINKED ISSUE #451: Add same toggle interaction for other 'earned because of' form items.
See issue #437 and merge #445.

wacii:
This generalizes the previous `<EarnedBecauseOfChildCare>` component to `<CashFlowRowWithPrompt>`, then applies it to `"earnedBecauseOfAdultCare"`. I think this was the only other field.

Closes #451


wacii:
Agree about the name. I went with after confirm, as the user is confirming that they have income to be reported before inputting it. I think that works.

Fixed the bug. That was there previously. I just forgot to address that case. Which is to initialize the control with the value from the client, rather than just assume it hasn't been inputted yet.

PR #456, opened 2018-03-27
Fixes #454 - add 'X' close button in feedback modal and disable cancel button on submit

### LINKED ISSUE #454: Add "X" close button on feedback modal and disable "Cancel" button during submission
As discussed in #444

ethanbb:
Fixes #454, plus makes the modal a little prettier.

(Also includes recent commits to dev. I meant to do this in a separate feature branch so I could update "feedback" to match "dev" separately, but I made a mistake, sorry. All my changes are in FeedbackPrompt.js.)

### DIFF
--- a/src/components/FeedbackPrompt.js
+++ b/src/components/FeedbackPrompt.js
@@ -48,7 +48,7 @@ class FeedbackPrompt extends React.Component {
     });
   }
 
-  cancel = (event) => {
+  close = (event) => {
     // Reset state for next time it's opened
     this.setState({
       formData: {},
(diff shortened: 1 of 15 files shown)


src/components/FeedbackPrompt.js
```diff
@@ -82,9 +79,16 @@ class FeedbackPrompt extends React.Component {
     });
 
     return (
-      <Modal open={this.props.isOpen}>
+      <Modal
+        size='large'
+        open={this.props.isOpen}
+        onClose={this.close}
+        closeOnDimmerClick={false}
+        closeOnEscape={false}
```
knod: I love closing on escape, but maybe that's just me ;)
Also, not something we have to mess with right now.

ethanbb: Fair! I disabled it though because I wanted to make it harder to accidentally close the modal and delete your answers.

I also think showing a 2nd warning modal on close with some nonempty answers could be a good idea.

PR #458, opened 2018-03-28
Change all download buttons to feedback buttons

ethanbb:
Change all the download stuff to feedback stuff. Includes the buttons in modals, which pop up the feedback modal on top of them so that the user can return after submitting feedback.

Edit: also includes changes to prompt (modal) text in order to direct users to submit feedback rather than to download the data and email it to us.


knod:
@migueldcepeda: Just pinging you on this one because the setup for the style of the download button got changed when it turned into the feedback button and it'd be nice to have your eyes on the new setup to see if it lines up right.

src/components/OnLeavePrompt.js
```diff
@@ -55,28 +42,18 @@ class OnLeavePrompt extends React.Component {
     var realLeave   = leaveText || 'Leave',
         realStay    = stayText || 'Stay',
         realMessage = message;
-    if ( message === 'default' ) { realMessage =  'Selecting "' + realLeave + '" will erase the information you have put into the form. You will still be able to click it after downloading.'; }
```
wacii: I think it was @knod that added in that last sentence. Might want to check with her before taking it out.

ethanbb: @knod does the thumbs up mean keep it out or put it back in?

src/index.css
```diff
@@ -200,12 +189,20 @@ input[type=number]::-webkit-outer-spin-button {
...
+    border-bottom-right-radius: 0;
+}
+
 /* RESET BUTTON */
 #reset_fixed {
     position: fixed;
-    top: 253px;
+    top: 270px;
     left: -21px;
-    z-index: 3000;
```
ethanbb: @migueldcepeda Just to point out, I got rid of the z-index property on both of the side buttons because it seemed confusing to be able to select one of them while a modal was already open (and all the modals now allow submitting feedback as an option). Does that sound reasonable and was there any other reason for setting the z-index?

src/components/OnLeavePrompt.js
knod: @ethanbb: I was thinking for putting it back in and add your proposed changes. The wording will also need work at some point, but that'll be a task for another time.

src/index.css
migueldcepeda: @ethanbb That makes total sense and something I noticed but didn't want to mess with. I believe the z-index was something that was coded in prior to the edits I made.

migueldcepeda:
@knod Has change from download button to feedback button been merged. I don't see it on the latest version.

ethanbb:
@migueldcepeda Not yet, I'm gonna make that one change to the modal copy that @knod suggested tonight. After that I think we can merge?

migueldcepeda:
@ethanbb Oh ok, just wanted to make sure I wasn't doing anything wrong on my end.

src/index.css
knod: Yes, sorry, the z-index was there when the button was horizontal to make sure that the buttons were never covered by the questions.

PR #460, opened 2018-03-28
Change CustomClient to import from JSON

wacii:
The client data is now sent to Google Forms as JSON rather than being manually downloaded as a file and emailed to an admin. The load client function needed to reflect that.

### DIFF
--- a/src/components/CustomClient.js
+++ b/src/components/CustomClient.js
@@ -1,6 +1,21 @@
 import React from 'react';
 import { Divider, Form, Message } from 'semantic-ui-react';
 
+/**
+ * Load previous session from object.
+ * @callback loadClient
+ * @param {object} clientContainer - Object containing previous session as `client`.
+ */
+
+/**
+ * Form which loads previous session from JSON.
+ * 
+ * @param {object} props
+ * @param {boolean} props.mayLoadCustomClient - Whether form should be visible
+ * @param {loadClient} props.loadClient
+ * 
+ * @extends React.Component
+ */
 class CustomClient extends React.Component {
   state = {
     client: null,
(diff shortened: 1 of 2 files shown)


src/components/CustomClient.js
```diff
@@ -1,45 +1,66 @@
...
 
-// Reference
-// https://codepen.io/hartzis/pen/VvNGZP?editors=0010
-// https://developer.mozilla.org/en-US/docs/Web/API/FileReader
+class CustomClient extends React.Component {
+  state = {
+    client: null,
+    error: null,
+    json: ''
+  };
```
ethanbb: Setting the initial state should happen in a constructor, right?

wacii: This is taking place in the constructor. I think the feature is called class fields, and I believe it is still just a proposal.

wacii:
That sounds like a good change. Might as well add it while I'm here. I'll sneak that in tomorrow.

src/components/CustomClient.js
ethanbb: Ah OK cool, I wasn't aware of that feature!

src/containers/VisitPage.js
```diff
@@ -95,8 +95,9 @@ class VisitPage extends Component {
     getUserConfirmation.unset();
   }
 
-  loadClient = ( clientContainer ) => {
-    this.setState({ client: clientContainer.client });
+  loadClient = (clientContainer) => {
+    const client = Object.assign(cloneDeep(CLIENT_DEFAULTS), clientContainer.client);
```
ethanbb: OK thanks, one more change, I believe we want to Object.assign separately on client.current and client.future, in order to merge their fields rather than overwriting them entirely.

knod:
Hmm, not sure what you mean by merging `current` and `future`. I'll try to take a look, but if it makes sense to you, I'll take your word for it.

PR #462, opened 2018-04-04
Cleanup OnLeavePrompt

wacii:
It took me a bit to figure how the component worked and I wrote most of it. Figured I should clean things up for others.

The `<OnLeavePrompt>` itself should probably be renamed since it is being used for other purposes now, but I didn't think that quite as important as some of the other changes.


knod:
It looks like it makes sense to people! I approve!

PR #465, opened 2018-04-04
Pull feedback system into dev

ethanbb:


knod:
Weird, wonder why package-lock is still different. Any ideas?

ethanbb:
Looks like es6-promise was upgraded. Probably no big deal - it might've been triggered by installing `isomorphic-fetch`. But we're pretty much using the latest versions of everything, right?

knod:
I'm not sure what you mean by 'everything' because we haven't upgraded to new versions of react and such. I haven't noticed anything that's broken, though, so it may be fine. Any reason to think otherwise?

PR #478, opened 2018-04-17
Make ManagedNumberField.handleFocus set 'focused' and 'focusedVal' correctly

### LINKED ISSUE #475: Number field over-restricts values
tl;dr: Now the field will only show valid numbers, even while you're typing them instead of waiting to return to a valid number after you're done.

### Replicate:
1. Click in a number field
2. Enter a number
3. Click out of the number field
4. Click back into the number field by any decimal
5. Try to edit the decimals
  a. Delete old numbers
  b. Insert a new number at a decimal place
  c. Select the whole number and replace it with another number

### Result I'd Expect:
a. If deleting decimals, to reduce the number of decimals until the decimal point itself is deleted.
b. If inserting a new number, for the number I've inserted to appear, lengthening the count of the numbers past the decimal point.
c. If selecting and replacing, to see one number without any decimal places.

### Actual Result:
a. The last number just turns into a 0, while the tenths place turns into the number that was at the hundredths place.
b. If at the last decimal place, usually a different number appears than what I've typed in. If at the first decimal place, the previous number there is pushed into the hundredths place.
c. A whole number with two decimal places that are both 0 is shown.

### Environment:
Browser type and version: Chrome v65.0.3325.181

App version or commit id at the time: 9145d7ed8d881a5b63d71261378ec8736df0bcbf

### Other Notes
This may have happened because the changes made to create a blank field on a 0 value.

ethanbb:
Fixes #475. The main problem was that handleFocus wasn't setting "focused" to true if the value was nonzero, so it would only show the rounded value even if the user invisibly enters a number with more than 2 decimal places.  In addition, the function wasn't setting "focusedVal" to the value on focus, which is especially important if switching from one box to another on the same row.

### DIFF
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -400,14 +400,18 @@ const IntervalColumnHeadings = function ({ type }) {
 class ManagedNumberField extends Component {
   constructor ( props ) {
     super( props );
-    this.state = { valid: true, focused: false, focusedVal: this.props.value, };
+    var { format, value } = props;
+    this.state = { valid: true, focused: false, focusedVal: format( value ) };
   }  // End constructor()
 
   //change form to blank string after click, before input
   handleFocus = ( evnt, inputProps ) => {
     // This makes sure that only zeroes and blanks get reset
+    var { format, value } = this.props;
     if (!Number.parseFloat(evnt.target.value)) {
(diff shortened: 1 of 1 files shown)


knod:
Awesome! We'll test it out tonight!

PR #480, opened 2018-04-22
[WIP] Enforce style

wacii:
Adds a linter and code formatter (eslint and prettier).

Mostly default settings are used. Exceptions being I disabled the required prop types and allowed unused parameters if they begin with an underscore.

The linter caught a few real problems and a bunch of unused variables. `npm run lint` to see what it produces.

I should probably be using the create-react-app linter. And I would like to ultimately use pre-commit hooks. Opening this PR mostly to have something concrete to discuss.


knod:
If everyone's up for it, maybe we can all look through the various changes and then take some time on Tuesday for a Slack discussion about preferences and pros and cons.

src/components/MainMenu.js
```diff
@@ -1,33 +1,41 @@
...
+            href="http://www.codeforboston.org"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            <Image src={logo} size="tiny" />
+          </a>
+        </Menu.Item>
         <Menu.Item>
-          <a href="http://www.codeforboston.org" target="_blank" rel="noopener noreferrer"><Image src={logo} size='tiny' /></a>
+          <Link to="/">Home</Link>
```
knod: Side-note: These questions may sound stupid, but I'm going for them anyway. Did the linter do this? Is this somehow still linking to the correct page?

knod: lol, forget it. The lines of the diff are just off.

wacii:
I separated add eslint from apply prettier into separate commits. Might make things a bit easier to review.

ethanbb:
Cool, looks like most of what's left  to refactor is  getting rid of unused parameters without underscores. Mostly unused `params` objects. I'm afraid it might be hard to maintain a rule like "call it `_params` if it's not used, `params` otherwise" over several generations of newcomers to the project. Maybe there's a better way?

knod:
See #485 for voting on which formatting discussions to have.

PR #484, opened 2018-04-25
fixes #477 switchedRadioLabelYesNo

### LINKED ISSUE #477: 'Yes/No' radio buttons to go on the left, label to go on the right
Find the component that is being used to create the yes/no choices. Move things around so that the 'yes/no' choice is on the left while the label asking the question is on the right.

Leftere:
Fixed check box for "Do you have Section 8 Housing?"


knod:
Other than that, though, great job finding a simple solution!

PR #489, opened 2018-05-03
Fixes #488 - Sets up initial localization structure

### LINKED ISSUE #488: Set up initial localization architecture

knod:
This is one way to do it. I'm open to other ideas.


knod:
@ethanbb: Heh, wacii asked the same thing. It's complicated and may not be absolutely necessary, but we can have a deeper discussion on Slack.

PR #502, opened 2018-05-09
Add eslint with initial rules

### LINKED ISSUE #490: What code style items can be automated?
These are the rules which were agreed upon. Check off the ones that can be done by a formatter (prettier or something else). If two items conflict, check off one and it'll open up a discussion.

- Ternary operators need more discussion. Initial conclusion is 'none except in React', but that conversation isn't over.
- Code width (default at 80) still to be determined
- [x] Put an extra line at the end of each file
- [ ] Use jsdoc formatting conventions for jsdoc-type stuff (is that possible?)
- [x] Put spaces in between brackets (curly and square), e.g. `{ propName }` and `[ item ]`
- [x] Auto-close components without contents
- [x] Inline comments allowed
- [ ] Put two spaces before an inline comment
- [x] Put trailing comma after last item of object keys or array items
- [x] EXCEPT don't put trailing comma after last item on SINGLE-LINE object keys or array items
- [x] Put spaces around `=`
- [ ] EXCEPT no spaces around `=` for one-line attribute declarations
- [x] Put semicolons everywhere!
- [x] Use single quotes, not double quotes
- [x] No hoisting functions
- [x] Multiple lines required for things like if, for, while, functions

```js
// This:
if ( x === y ) {
  return true;
}

// Not this:
if ( x === y ) { return true; }
```

- [x] EXCEPT one-line arrow functions allowed
- [x] In if statement or function declarations, include a space before parens
- [x] Do not put spaces just inside parens

```js
// Declarations end up looking something like
var foo = function (x, y) {
  // Do things
}

// And
var foo = function ({ x, y }) {
  // Do things
}
```

- [x] Fix indenting to 2-spaces per indent level
- [x] Put items of lists or objects, and multiple component attributes, on separate lines
- [ ] Line up the names of variables/attributes and their `=` in variable/attribute declarations, e.g.

```js
var fooBar = 5,
    x      = 10;
```

- [x] Alternatively, don't mess with the white space of variable/attribute declarations
- [x] Always include brackets in if statements, function declarations, etc. (even arrow function declarations)
- [x] Put parents around arguments in arrow-function declarations, e.g. `(foo) => { return true; }`

DON'T REMOVE CHECK MARKS FROM THIS SECTION. These are new items. Like the poll, if you want to discuss one, mark it.
- [ ] `/>` always on the same line as the last attribute
- [ ] Inline `/>` always has a space before it, e.g. `<Walrus />` and `<Dome night={ true } />`

I think @wacii is the one that knows Prettier, and/or other formatting systems, the best, so I'm assigning it to them, but other folks who know stuff can also get involved. If you find yourself in a checkbox war, though, just add a comment below.

wacii:
I only addressed the first batch of style rules. The new set haven't been marked for discussion yet, so I wasn't sure if anyone wanted to include them or not.

Didn't fix the fixable violations this time. Just to keep the diff more readable.

Closes #490


wacii:
The lint command now modifies code, so I thought it important to run the command as part of the PR, even if it confuses the diff.

Found conflicting rules. One required each item in a list or object on a new line while another required spaces before and after inline lists. Not sure what you want to go with.

ethanbb:
Ooh I may have missed the multi-line lists/objects item on the survey. I don't think the items should be on multiple lines in all cases.

wacii:
> There were no spaces between brackets (src/components/ConfirmLeave.js 15)

The linter will add spaces between brackets for objects and arrays, but expressions in code bodies are supposed to be on separate lines. The linter caught the error, it just could not fix automatically.

> Space before inline /> or > (src/components/CustomClient.js 73 and other places)

I added a rule enforcing spaces before the `/>` in self-closing tags, but I don't think there is an existing rule for the other one. Specifically I cannot find a way to enforce a space before `>` in the opening tag only when props are present.

> With multiple attributes, all attributes on separate lines (src/components/FeedbackAnytime.js 9)

Moved the first prop to a new line when there are multiple props.

> Set state stuff looks too long but is one-line (src/components/FeedbackPrompt.js 27)

Do you want to enforce a max line length?

> Weird same-line child when parent and child attributes are on second line (src/components/MainMenu.js 22)

Enforced consistent formatting of closing tag.

> Correcting indenting messes up well-aligned variable declarations - variable names in same group should line up with each other. Not sure there's a way to fix this. (src/components/OnLeavePrompt.js 45 and more)

I left-aligned the variable declarations. The extra spaces around the equals isn't great, but I imagine eliminating all extra whitespace would mess up formatting elsewhere in the project.

> Btw, lined-up spacing would be for object props too, not just regular variable declarations

Lined-up spacing? You mean aligning the values in multi-line objects? Unlike variable initializations, that is actually possible with [key-spacing](https://eslint.org/docs/rules/key-spacing).

knod:
Wow, that should solve a lot of those issues. Not sure I have the fortitude to double check everything that's going on, but I'm sure we can deal with what doesn't work later.

> Lined-up spacing? You mean aligning the values in multi-line objects?

Yeah, values in multi-line objects. Why the heck is that possible while multi-line variable declaration aligning isn't? This world is a nutty place.

wacii:
Object values are now aligned. You can change the rule to `{ "align": "colon" }` if you want the colons to be aligned as well.

PR #504, opened 2018-05-12
Adds graphs back into the app, but in a way that's more opt-in

knod:
Styling for tabs, different names, and a default to be displayed for the 'Benefit Changes' graph when no benefits have been selected.


knod:
@wacii: Thanks, and you're probably right. That snapshot... may be premature...

Issue #507, opened 2018-05-14
Use eslint to fix code style issues (ongoing)


ethanbb:
We just added the linting tool eslint to make our code style more clear and consistent. Using the `--fix` option, it fixed many issues automatically, but several others (many relating to jsdoc comments) are still present. We should gradually fix these (while also avoiding introducing more issues in new features).

To use eslint to _view_ current style issues, do one of the following:
- Command line: with a branch derived from the current 'dev' branch checked out and all packages installed with `npm install`, run `npm run lint`. Warnings and errors for each file with issues should appear. Fix some issues, prioritizing errors.

- If you use Atom: install the package `linter-eslint` (including dependencies). Then issues and warnings should appear inline in each file when you have the project open.

- If you use VSCode: Install the package `vscode-eslint`. Click the install link on [this page](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint), which will open the extension in VSCode and allow you to install it there. Issues and warnings should appear inline in each file when you have the project open.

knod:
I think for Sublime Text 3, you can use this package: https://github.com/SublimeLinter/SublimeLinter-eslint

knod:
ESLint installation descriptions in our wiki now: https://github.com/codeforboston/cliff-effects/wiki/ESLint-Setup

knod:
I think it'd be great if someone could edit in a more detailed description of using Atom's package manager.

ethanbb:
Awesome @knod, just updated the Atom instructions.

ethanbb:
Closing because @ethanstrominger has gotten rid of all the lint errors and warnings! Yay!

PR #509, opened 2018-05-14
Linter warning fixes for a couple of files, #507

knod:


knod:
@wacii: I'll take both of those, thank you very much! Maybe we'll wait till after this PR is merged, though.

knod:
I'll look into implementation 👍

src/forms/BenefitsTable.js
```diff
@@ -73,36 +78,41 @@ const BenefitsTable = function (props) {
...
 
 
-
-
-
-
   const SNAPBenefitRow = function(props){
-    if (!client.current.hasSnap) {return (null);}
+    if (!client.current.hasSnap) {
+      return (null);
```
ethanbb: Why have parentheses around return values?

knod: Not sure if I did that or if it was someone else. If it was me, I was hallucinating at the time. Maybe it was to mimic the way that Components are usually returned in parens. Not sure, though.

PR #510, opened 2018-05-15
Updates package-lock.json... but wasn't that already done?

knod:
Did anyone else have this issue come up with they `npm install`ed
the new stuff?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2078,7 +2078,7 @@
     "circular-json": {
       "version": "0.3.3",
       "resolved": "https://registry.npmjs.org/circular-json/-/circular-json-0.3.3.tgz",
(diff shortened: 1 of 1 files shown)


ethanbb:
Yeah I did as well. Could be an npm version mismatch perhaps? I know npm 6 was recently released (and I have it installed).

knod:
I was at npm v5.5.0. Now I'm at v6.0.1. I started a new branch from 'dev' with no 'package-lock.json' changes, I destroyed my node_modules folder, I `npm install`ed, and I've still got a new 'packate-lock.json'. I'll look into some kind of SHA config.

knod:
Followed these instructions as well and still ended up with a different 'package-lock.json': https://stackoverflow.com/questions/45553401/npm-using-stale-package-data

knod:
Also, I do get a SHA512 now, it's just a different one than the one in 'dev'...

wacii:
Uh, different checksum with the same version of the module? That's not supposed to happen. Can you push the new package-lock?

knod:
Looks like that'll have to do, then. Life is a mystery.

PR #522, opened 2018-05-20
Experiments in plainer language.

knod:
Thoughts?


ValerieKenyon:
I think this new language works well - it's more user-friendly.

knod:
@ethanbb: I think it does need some kind of label. Any specific suggestion for different wording?

knod:
Our discussion seemed to settle pretty firmly on this version. I'm calling it.

PR #523, opened 2018-05-20
Lint config to add line at end of file

knod:


knod:
@wacii: Ah, got it. Assumed it was a typo. I'll look into their string versions.

wacii:
0 = "off", 1 = "warn", 2 = "error"

knod:
What's the deal with the strings in arrays vs. just strings, btw? Can they all be in arrays?

knod:
I'll make a different PR with this config stuff. I was planning on doing some cleaning in here anyway.

wacii:
I think the default way to provide the arguments is with an array. The just-string way is a shortcut for when you only have the one argument.

PR #524, opened 2018-05-20
eslint config format

knod:
Alphabetized some stuff, among other things.


knod:
@ethanbb: Yeah, seemed safer.

PR #525, opened 2018-05-21
Changes `housing` to `section8` and `shelter` to `housing`

knod:
The first is because many programs can be housing programs.
The second is to be more consistent with the language used by
case managers, etc.

Should maybe have been two PRs. Sorry, folks.


knod:
@wacii : Wow, nice finds, thanks!

PR #528, opened 2018-05-23
Moves `snippets` to top level of app.

knod:


src/App.js
```diff
@@ -13,35 +13,84 @@ import VisitPage from './containers/VisitPage';
...
-          component={ VisitPage } />
+class App extends Component {
+  constructor (props) {
+    super(props);
+    this.state = { language: 'en', snippets: getTextForLanguage('en') };
+  }
+
+  setLanguage = (inputProps) => {
+    var snippets = getTextForLanguage(inputProps.value);
+    this.setState({ snippets: snippets });
```
ethanbb: Should this function also set `this.state.language`?

src/App.js
```diff
@@ -13,35 +13,84 @@ import VisitPage from './containers/VisitPage';
...
+        <HashRouter getUserConfirmation={ getUserConfirmation }>
+          <div id='HashRouter'>
+            <Route
+              path="/:rest+"
+              component={ (props) => {
+                return (
+                  <Header
+                    { ...props }
+                    snippets={ snippets } />); 
+              } } />
```
ethanbb: I think each of the `component` function return statements could easily be one line, as in:

```
            <Route
              path="/:rest+"
              component={ (props) => {
                return (<Header { ...props } snippets={ snippets } />); 
              } } />
```

ethanbb: Oh I guess that's against our linting rules. I don't love it but never mind then.

src/App.js
knod: Good point!

shp54:
I think (React's `context` feature)[https://reactjs.org/docs/context.html] is worth investigating if we want to do this, so that we don't have to manually pass snippets down through chains of components.

shp54:
The API went through a significant change in v16, so I think it's worth
looking up and will be plenty explicit.

knod:
@shp54 : Hmm, I think you're right, it's pretty explicit. Maybe I was thinking of something else or maybe it did change. On the other hand, we're going to have a bunch of stuff wrapped inside a `<LanguageContext.Consumer>`. Unless we abstract that somehow...? We could bring the translate function back to return that component, but then we're basically in the same boat we're in right now, passing stuff around. Maybe we can make `translate()` importable, yet able to use context. Then it wouldn't have to be passed through `props`. Hmmm.

@ethanbb, @wacii : Thoughts?

knod:
I'm going to merge this PR so that people can get started working on other pages and I think we'll adjust once we have a chance to continue discussions.

wacii:
@knod If you use context, I would create a translate/message/snippet component, so you could hide all the context consumer details in that and not leak them all over the application.

knod:
@wacii: Yeah, exactly. The question is, are we going back to the old `translate()` with the non-js-like-object-syntax? I'd rather not, but the only other option would be to do something like this: https://codepen.io/knod/pen/GdaXpr. It hides a lot, so I don't know if I'm a huge fan, but this way, each component script could import a `snippets` 'object' and still use it like a js object.

PR #530, opened 2018-05-23
Adjusts snippets for incomes to match current income labels

knod:


knod:
Ok, folks, this has been hanging out here too long. It's pretty simple stuff, so I'm going to call it done.

knod:
@KevinSJ : If you're up for adjusting the Chinese snippets to match these, we'd really appreciate it. We also hope to have some input from case managers soon to provide clearer and more specific explanations for the various incomes, so we'll ping you when we get some of those in place. Thanks for your help so far!

PR #533, opened 2018-05-27
Adds German for homepage, footer, labels, and hints

ValerieKenyon:
Creates de.js, updates all.js, and adds German text for the homepage and footer.


ethanbb:
I'll just fix the offending newline if it's OK, actually, since it's such a small thing.

ValerieKenyon:
Thanks Ethan! (Sorry that it took me a couple of days to get back to you)

Issue #513, opened 2018-05-16
Convert hard-coded strings to keys for localization


ethanbb:
We want to factor out any and all strings in the interface into localization structures (like the one in `src/localization/en.js`) in order to let the whole app be translated into whatever languages we want/can. We also need to pass the `translate` function down to each display component so that it can retrieve the correct string(s) using a unique key.

See #489 for an example.

knod:
Closed by #521

knod:
@ethanbb: Isn't that covered by #500? I thought this was to change to a nested object instead of a flat one.

ethanbb:
My impression was that #500 is to actually translate existing English strings into other languages, whereas this issue is to factor remaining string literals in the rest of the codebase out into the English localization file.

knod:
Ah, ok. I'd been thinking of #500 for both, but maybe it's worth separating them.

knod:
Maybe we could turn this into specific assignments - have an issue for each page that still need the localization infrastructure.

knod:
@ethanstrominger , this was probably the issue you should have been referencing all along :/

knod:
Just so we're clear, this is for purposes of localization.

PR #541, opened 2018-05-30
Reorganize snippets into subobjects for each page

ethanbb:
As a first step to #513, this reorganizes `en.js` and the other language files somewhat so that each step is in its own nested object. The key matches the step title as specified by `VisitPage.js`, and this is used to pass only the snippets that are needed to each step (and likewise for HomePage and Footer). I also alphabetized the languages and made a couple aesthetic changes.

### DIFF
--- a/src/containers/VisitPage.js
+++ b/src/containers/VisitPage.js
@@ -71,7 +71,26 @@ class VisitPage extends Component {
     };  // end this.state {}
 
     this.steps = [
-      { title: 'Current Benefits', form: CurrentBenefitsStep }, { title: 'Household', form: HouseholdStep }, { title: 'Income', form: CurrentIncomeStep }, { title: 'Expenses', form: CurrentExpensesStep }, { title: 'Predictions', form: PredictionsStep },//,
+      {
+        title: 'Current Benefits',
+        form:  CurrentBenefitsStep,
+      },
+      {
+        title: 'Household',
+        form:  HouseholdStep,
+      },
+      {
+        title: 'Income',
+        form:  CurrentIncomeStep,
+      },
+      {
+        title: 'Expenses',
+        form:  CurrentExpensesStep,
+      },
+      {
+        title: 'Predictions',
+        form:  PredictionsStep,
+      },//,
     //  { title: 'Graphs', form: ResultsGraph }
     ];  // end this.steps {}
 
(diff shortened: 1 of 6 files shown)


knod:
It would add consistency in general - each page being represented by a key. I'm not completely sure that `VisitPage` isn't significant, but it was kinda nice to have just one level of nesting. I know we haven't done anything with that yet, but it's often useful to prefer flat over deep as far as architecture goes.

Honestly, I'm torn on which way to jump. I'm leaning more towards consistency because I've seen inconsistency be a big barrier to bringing new folks on, but that may just be a story I've built in my head and I'm not sure the structure you have here lacks clarity. Maybe we can call out some of our members who are new to coding to pipe in on Slack about what they'd find least blocking.

knod:
It looks like a nested object makes sense to folks. Let's go with that for now.

ethanbb:
OK, I added `visitPage`. Also, separated the names of steps from their keys in the translation files in case we want to change the names in the future.  Finally I removed the quotes from the keys in the translation files for consistency, and since they should be valid identifier names anyway I would think.

src/containers/VisitPage.js
```diff
@@ -72,24 +72,29 @@ class VisitPage extends Component {
 
     this.steps = [
       {
-        title: 'Current Benefits',
-        form:  CurrentBenefitsStep,
+        displayName: 'Current Benefits',
```
knod: Not blocking, but this may be better as `tabName` if we're going to change the prop name. Is this related to the snippets changes?

src/containers/VisitPage.js
```diff
@@ -71,7 +71,31 @@ class VisitPage extends Component {
     };  // end this.state {}
 
     this.steps = [
-      { title: 'Current Benefits', form: CurrentBenefitsStep }, { title: 'Household', form: HouseholdStep }, { title: 'Income', form: CurrentIncomeStep }, { title: 'Expenses', form: CurrentExpensesStep }, { title: 'Predictions', form: PredictionsStep },//,
+      {
+        displayName: 'Current Benefits',
```
knod: Not a blocking note, but this might be better called `tabName` or `tabTitle`.

src/containers/VisitPage.js
knod: Forget this one, look at the next one instead. I just saw your note about what you were trying to do with this. I'm still not sure it's part of the snippet changes, but meh.

ethanbb: We can keep it as `title` actually...I changed it to try to make it more clear that this was a display string as opposed to an internal identifier, but that's probably not necessary.

Although we could actually probably get rid of this field entirely, and instead add the title to the snippets for each step and use `snippets[ step.key ].title` for the tab text.

ethanbb: Or `tabTitle` to keep it unambiguous

ethanbb: Actually I guess it would be better to have all the tab names together in one section of the snippets

PR #559, opened 2018-06-04
Use snippets in AboutContent.js and Footer.js, and allow interpolation of arbitrary inline components

### LINKED ISSUE #558: Create mechanism to indicate where certain items like links should be inserted within a translated paragraph
Basically I want to allow flexibility for different languages to order the links we have on a page in a way that makes sense, while still keeping the translation files pretty simple. I anticipate doing this with arrays that can contain both translated snippets and objects to identify specific JSX items to interpolate, along with a function to parse these arrays. See #557

ethanbb:
I went through and added the English text for the About page, and modified `AboutContent.js` and `Footer.js` to use the snippets rather than hardcoded strings. Along the way (as mentioned in #557, to fix #558), I made a way to refer to specific inline components, such as links or icons, within translation files and specify where they should be placed. This is done by creating an `InterpolatedText` component, which takes an array from the snippets object as `template` and an object containing the menu of components as `components`. Strings from `template` are passed through unchanged, and each object is interpreted as follows:

* If the object has a `name` key, it is replaced by a copy of the component with that name in `components`.
* Additionally, if the object has a `text` key, the children of the object are replaced with the "`text`". I called it `text` because in most cases, for things like links and `strong` spans, the children are just the displayed text. We could change it to something more generic if we want to though.

The issue of the order of links changing depending on the language might be pretty minor and infrequent, but I think doing inline components this way is also just clearer when reading the translation files, since you can see where each component goes. For example, see `cfbCredit` in the `footer`.


ethanbb:
Actually, I think there's a cleaner way to do this without having to insert a new element into the page in seemingly arbitrary places. Will update tonight.

ethanbb:
OK, I think this version is a lot cleaner. Let me know what y'all think!

knod:
Just to get it on record, I think the current plan is to test out a codepen mockup of this format with some translators and see how it goes!

shp54:
I'm sure you thought of this, but why not just use JSX snippets directly? React can do this interpolation automatically, I think. Is there anything that would make that a no go?

ethanbb:
@shp54 you mean just put the JSX elements instead of the names of them in `en.js`, etc? Could work, but my thought was that we should keep the duplicated information to a minimum, in case we need to change one of the links or something. It would also further increase the complexity of these files, potentially making it harder for translators.

Not to say I think this way is perfect though - definitely don't hold back with other feedback!

knod:
To catch people on on Slack conversations - one idea for `name` values is to make them visually distinct:
```js
{ name: '__someKeyName__' }
```
We'd still have to let people know not to edit the stuff in there, but it would be easier to describe.

src/localization/en.js
```diff
@@ -1,13 +1,13 @@
 export default {
 
-  langName: 'English',
+  langname: 'English',
```
knod: I feel like I must have asked this before, but why was this changed away from camel case?

knod:
I keep forgetting to mention this - on one of our next passes, I suggest we change this particular file to use double quotes so we don't have to escape single quotes. I think the character escaping may be confusing to some folks. We could possibly use template ticks, but I'm not sure. Maybe we should make that an issue for discussion...

src/localization/en.js
ethanbb: Oops, that was definitely a mistake. It got caught in my find and replace for `name: "xxx"` fields. Will fix now.

PR #561, opened 2018-06-07
Unit tests for ManagedNumberField for handleFocus and handleBlur

snyderc:
I wrote a few unit tests for ManagedNumberField -- is this along the lines of what you were looking for? If yes, then I'll continue writing other tests based on the Google Doc.

Specifically, this PR implements the tests requested in the Google Doc:
- handleFocus: should blank the component before focus, should set focused in state to true
- handleBlur: should set state to focused: false

I also added a snapshot test to make sure ManagedNumberField was rendering as expected.

### DIFF
--- a/src/test/forms/formHelpers.test.js
+++ b/src/test/forms/formHelpers.test.js
@@ -1,0 +1,48 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import { ManagedNumberField } from '../../forms/formHelpers';
+
+test('ManagedNumberField should match snapshot', () => {
+  const wrapper = shallow(
+    <ManagedNumberField
+      format={ jest.fn() }
+      value={ 0 } />
+  );
+  expect(wrapper).toMatchSnapshot();
+});
+
+test('should set focused state to true on focus', () => {
+  const value = 10;
+  const wrapper = shallow(
+    <ManagedNumberField
+      format={ jest.fn() }
+      value={ value } />
+  );
+  wrapper.find('FormInput').simulate('focus', { target: { value }}, undefined);
+  expect(wrapper.state('focused')).toBe(true);
+});
+
+test('should blank focusedVal state before focus if current value of FormInput is 0', () => {
+  const inputValue = 0;
+  const wrapper = shallow(
+    <ManagedNumberField
+      format={ jest.fn() }
+      value={ 0 } />
+  );
+  wrapper.find('FormInput').simulate('focus', { target: { value: inputValue }}, undefined);
+  expect(wrapper.state('focusedVal')).toBe('');
+});
+
+test('should set focused state to false on blur', () => {
+  const value = 10;
+  const wrapper = shallow(
+    <ManagedNumberField
+      format={ jest.fn() }
+      onBlur={ jest.fn() }
+      value={ value } />
+  );
+  wrapper.setState({ focused: true });
+  wrapper.find('FormInput').simulate('blur', {});
+  expect(wrapper.state('focused')).toBe(false);
+});
--- a/src/test/forms/formHelpers.test.js
+++ b/src/test/forms/formHelpers.test.js
@@ -6,20 +6,31 @@ import { ManagedNumberField } from '../../forms/formHelpers';
 test('ManagedNumberField should match snapshot', () => {
   const wrapper = shallow(
     <ManagedNumberField
-      format={ jest.fn() }
+      format={ () => {} }
       value={ 0 } />
   );
   expect(wrapper).toMatchSnapshot();
 });
 
-test('should set focused state to true on focus', () => {
+test('should set focused state to true on focus if current value of FormInput is positive number', () => {
   const value = 10;
   const wrapper = shallow(
     <ManagedNumberField
-      format={ jest.fn() }
+      format={ () => {} }
       value={ value } />
   );
-  wrapper.find('FormInput').simulate('focus', { target: { value }}, undefined);
+  wrapper.find('FormInput').simulate('focus', { target: { value }});
+  expect(wrapper.state('focused')).toBe(true);
+});
+
+test('should set focused state to true on focus if current value of FormInput is 0', () => {
+  const value = 0;
+  const wrapper = shallow(
+    <ManagedNumberField
+      format={ () => {} }
+      value={ value } />
+  );
+  wrapper.find('FormInput').simulate('focus', { target: { value }});
   expect(wrapper.state('focused')).toBe(true);
 });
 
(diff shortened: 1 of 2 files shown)


knod:
Life's a little crazy right now, but if no one gets to it before I can, I'll check it out as soon as I can!

ethanbb:
(After reading a little documentation I've concluded that it's autogenerated)

snyderc:
The snapshot file is autogenerated, but it's good to look at it to make sure it looks correct. When a snapshot test is first run, a snapshot is generated, and future iterations of the test assume that the original snapshot is the desired state. A snapshot test only fails if a future iteration returns a snapshot different than the original saved snapshot.

That being said, this may be too simple a component to be worth doing a snapshot test for.

src/test/forms/formHelpers.test.js
```diff
@@ -0,0 +1,48 @@
...
+});
+
+test('should set focused state to true on focus', () => {
+  const value = 10;
+  const wrapper = shallow(
+    <ManagedNumberField
+      format={ jest.fn() }
+      value={ value } />
+  );
+  wrapper.find('FormInput').simulate('focus', { target: { value }}, undefined);
```
ethanbb: Any reason to have `undefined` as the third argument? It looks like Enzyme's `simulate` only normally takes an event name and mock event object.

src/test/forms/formHelpers.test.js
```diff
@@ -0,0 +1,48 @@
...
+
+test('should blank focusedVal state before focus if current value of FormInput is 0', () => {
+  const inputValue = 0;
+  const wrapper = shallow(
+    <ManagedNumberField
+      format={ jest.fn() }
+      value={ 0 } />
+  );
+  wrapper.find('FormInput').simulate('focus', { target: { value: inputValue }}, undefined);
+  expect(wrapper.state('focusedVal')).toBe('');
```
ethanbb: We should also make sure `focused` is true here - `value` being 0 is a special case and will not necessarily be handled the same way.

src/test/forms/formHelpers.test.js
```diff
@@ -0,0 +1,48 @@
...
+
+test('should set focused state to false on blur', () => {
+  const value = 10;
+  const wrapper = shallow(
+    <ManagedNumberField
+      format={ jest.fn() }
+      onBlur={ jest.fn() }
+      value={ value } />
+  );
+  wrapper.setState({ focused: true });
```
ethanbb: Might want to simulate focus here rather than explicitly setting `focused` to true, to mimic exactly how it would work in a real interaction.

snyderc: I chose to explicitly set focused=true in the onBlur test to keep this test specific to testing the onBlur call, and let the onFocus test take care of simulating focus. (I don't have an explicit focused=false statement in the onFocus test because the default value of focused is false anyway.) I would suggest keeping it this way to keep the tests specific.

snyderc:
Thank you for the feedback, Ethan! I have made some updates and also explained my reasoning on other questions you had.

My new commit also changed the spies (jest.fn()) to empty arrow functions, because I realized the spies were unnecessary -- the tests just need to know that certain props are of type function, but these particular tests aren't testing whether the function hasBeenCalled.

src/test/forms/formHelpers.test.js
ethanbb: I would argue that, without assuming anything about how `handleFocus` works, setting `focused` to true explicitly could put the component in an invalid state. Although maybe that doesn't matter for such a simple test, and doing it this way does mean the test will still work even if `handleFocus` is broken. 

Although now that I think of it, shouldn't we be testing the props of the underlying `input` component, not the state of our component which could change arbitrarily?

snyderc: I thought about whether the FormInput component itself needed to be tested. Looking at the component, it appears to be an out-of-the-box semantic-ui-react component. If you take a look at the snapshot file, there's very little that could be tested inside of FormInput -- we could replace the function props with spies to make sure that onBlur, onFocus, etc. are called, but that would suggest we're not trusting the semantic-ui-react module to have correctly wired onBlur/onFocus/etc. Only by rendering the test at the ManagedNumberField level can we see that FormInput's onBlur calls the ManagedNumberField's handleBlur function.

Running the state-setting unit tests at this level is also consistent with how it was done in the React/Jest tutorial I took, if that helps.

I think your concern is addressed by having the snapshot test. If the ManagedNumberField component is modified to the extent that it renders differently, the snapshot test will catch that. If the ManagedNumberField component is modified so that a function passed down as a prop changes (e.g. it passes down a different function to FormInput as the onBlur prop), the render doesn't change so the snapshot test would pass, but the corresponding state-setting test in ManagedNumberField would fail.

For your reference: I went through a lengthy React tutorial that included testing with Jest/Enzyme, but this is the first time I'm writing tests in a non-tutorial environment -- which is why I started with a few small unit tests :-) So I'm basing my comments on the principles I learned in the tutorial.

ethanbb: OK, I think you're right about not testing `Form.Input` in this case actually. I was thinking that the focused state of the underlying input was managed by `ManagedNumberField` similarly to its `value` (i.e. passed into `Form.Input` as a prop), but I now realize that the `focused` field just keeps track of the focused state of the component, which is managed by `Form.Input`. When we get to making tests about the value of the input, we should check that the `Form.Input` value prop matches what we want it to be.

The other issue I had is pretty minor the more I think about it, so I think this is ready to be merged.

src/test/forms/formHelpers.test.js
```diff
@@ -0,0 +1,59 @@
...
+test('ManagedNumberField should match snapshot', () => {
+  const wrapper = shallow(
+    <ManagedNumberField
+      format={ () => {} }
+      value={ 0 } />
+  );
+  expect(wrapper).toMatchSnapshot();
+});
+
+test('should set focused state to true on focus if current value of FormInput is positive number', () => {
```
knod: Shouldn't `focused` state be set to true any time someone clicks in the field? It's just the resulting value of the field after that focus that's in question, right?

knod:
Great looking tests! And awesome to see this stuff in action. Among the other thoughts in my head - we've talked about refactoring 'formHelpers.js' before. This might be a good opportunity to do that. Putting this here as a reminder for a discussion item. If I don't forget, I'll make an issue for it. Anyone can feel free to bring it up if I space out!

src/test/forms/formHelpers.test.js
ethanbb: I think the idea was just to test that `focused` is true when the field is actually focused, and we do this check separately for zero and nonzero values of the field (since we handle them differently).

knod: It's not 0 vs. non-zero values, right? It's for any falsy float-parsed value, right? I don't exactly recall the situation that made this necessary, but it had something to do with `NaN`. @shp54 : do you recall?

knod: y'know what, we can always fiddle with it later. Thanks, @snyderc!

PR #565, opened 2018-06-10
Makes changes for lint msgs issue #507 - CustomClient.js

ethanstrominger:


ethanstrominger:
Made a mistake so redoing

PR #571, opened 2018-06-10
Makes changes for lint msgs issue #507 - formHelpers.js

ethanstrominger:


src/forms/formHelpers.js
```diff
@@ -158,11 +159,12 @@ const FormPartsContainer = function(props) {
         </Header>
         { !props.clarifier
           ? null
-          : <Header
+          :
```
ethanbb: I'd prefer having an opening parenthesis here, so the `Header` under it is indented, and then a closing parenthesis on its own line after line 167. I think the indenting looks better that way. Anyone else have a preference?

src/forms/formHelpers.js
```diff
@@ -336,11 +342,12 @@ const RowMessage = function ({ validRow, message }) {
 
   var result = null;
   if (!validRow && message) {
-    result = <Label
+    result = 
+    <Label
```
ethanbb: Same as above, I'd probably enclose the `Label` in parentheses.

src/forms/formHelpers.js
```diff
@@ -497,8 +506,12 @@ class ManagedNumberField extends Component {
     var { value, name, className, format }  = this.props;
 
     // Format correctly when neighbors are updated, if needed
-    if (!focused) { value = format(value); }
-    else            { value = focusedVal; }
+    if (!focused) { 
+      value = format(value); 
+    }
+    else {
```
ethanbb: I think this can go on the same line as the closing bracket above - that's how I usually see it.

src/forms/formHelpers.js
knod: I'll give my thoughts on other stuff later, but I think the closing paren would be on a new newline after line 165.

src/forms/formHelpers.js
knod: Parens here makes sense to me, too. I'm not conceding on the ternaries yet, though.

PR #576, opened 2018-06-10
Makes changes for lint msgs issue #507 - CurrentExpenses.js

ethanstrominger:
Commented out in case useful when implement


ethanbb:
I feel the same way about these ternary operators as in #571. Anyone else have an opinion about that?

Issue #550, opened 2018-05-31
consider text change "reset"


ohmegasquared:
Consider changing “Reset” button text to “New Client” or similar sentiment

knod:
'New Client' to match the left-hand-side fixed button would be consistent. (The button on the left used to say 'Reset', too. I think we just forgot the last-form-section button during that change.)

PR #577, opened 2018-06-10
#550 Changed Last Button from Reset to New Client and Edited CSS to have Text on One Line

cohalls:


knod:
Thanks for the PR! When we get a chance, we'll figure out the smoothest way to get this matched up wit the tests and our other requirements, and then we can pull it in!

PR #579, opened 2018-06-10
Issue #507 - clears lint warnings for 'src/forms/Predictions.js'

knod:


src/forms/Predictions.js
```diff
@@ -171,17 +171,23 @@ const TabbedVisualizations = ({ client }) => {
...
-            client={ client }
-            Graph={ GrossGraph } />
-          </Tab.Pane>;}, 
+          render:   () => { return (
+            <Tab.Pane>
+              <GraphHolder
+                client={ client }
+                Graph={ GrossGraph } />
+            </Tab.Pane>
+          );},
```
shp54: I'd put the `return` on the line after the braces, personally. Not a dealbreaker.

PR #581, opened 2018-06-10
Switches to 'consistent' bracket style. #507. No fixes needed.

knod:
As per discussion. Can't find the issue for it. Anyone remember one?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -20,9 +20,8 @@
     "babel/semi": 1,
     "react/prop-types": 0,
 
-    "array-bracket-newline": ["warn", { "multiline": true }],
+    "array-bracket-newline": ["warn", "consistent"],
     "array-bracket-spacing": ["warn", "always"],
-    "array-element-newline": ["warn", { "multiline": true }],
     "arrow-body-style": ["warn", "always"],
     "arrow-parens": ["warn", "always"],
     "comma-dangle": ["warn", "always-multiline"],


ethanbb:
I think we just talked about it on Slack without making an issue.

knod:
I thought it allowed
```js
var a = [ 1, 2 ];
// and _also_
var a = [
  1,
  2
];
```

Which is a thing I thought someone wanted and made sense to me.

ethanbb:
Oh whoops, you got rid of `array-element-newline` - didn't see that. That's good. My first comment was about changing `array-bracket-newline`, but that's not as big a deal as getting rid of the other one.

knod:
Hmm, thought I'd read over those thoroughly. Maybe I should re-read instead of trusting my past self.

knod:
So my understanding was that we wanted both to be able to put array/bracket items on one line and, at other times, to choose to put the array/bracket items on multiple lines. That is, we wouldn't require non-empty brackets to have new lines in them. I think that's what `"consistent"` allows: https://eslint.org/docs/rules/array-bracket-newline

ethanbb:
As I understand it, `array-bracket-newline` is _just_ about newlines directly after opening brackets and directly before closing brackets. Whereas `array-element-newline` deals with newlines _between_ array elements.

knod:
Right, but we don't want to require new lines in brackets, right? We just want them there when elements are on new lines, otherwise we want everything to be on the same line.

knod:
Hmm, I'm rethinking this whole thing. This may not be what we were looking for.

knod:
With `"array-bracket-newline": ..."always"`, though, I don't think we can have the two things I showed.

knod:
Ok, new plan. We'll require new lines for elements and brackets stays the same. New PR coming up.

PR #583, opened 2018-06-12
Non ternary option for JSX

knod:
This is a possible alternative. Don't actually merge this before we figure out @ethanstrominger's current changes. I'm just trying to illustrate my proposal for what I think can be a clear and well-contained alternative to JSX's ternary operators.

This would need to go in written a style guide.

### DIFF
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -125,9 +125,9 @@ const BottomButtons = function({ left, right }) {
 };  // End BottomButtons() Component
 
 
-const Clarifier = function ({ children }) {
+const Clarifier = function ({ shouldShow, children }) {
 
-  if (children) {
+  if (shouldShow) {
     return (
       <Header
         as='h3'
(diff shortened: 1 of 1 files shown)


knod:
We actually could abstract this further. Hmmm. Maybe that makes sense. One moment please!

knod:
This is pretty blatant, but maybe that's better:
https://github.com/knod/cliff-effects-1/tree/conditional

ethanbb:
Interesting, I think the `Conditional` element would definitely be useful and it seems pretty clean to me. You could also make a version with contents for both true and false cases. Is there any reason to enclose the `children` in a div rather than returning them directly?

knod:
@ethanbb : No reason to not just return the children. Just didn't think of it! As for doing something for a true and false case, I was hoping to keep this pretty simple, without having to define components to put in as properties. I was thinking that if it got more complex than 'something or null', we'd have what I have here - a unique component that knows more about what it wants to show. Thoughts?

ethanbb:
Yeah, we definitely don't have to add unnecessary complexity! I was just thinking about the types of things that we could do with this type of design if we needed it.

knod:
That worked just fine! I'll remove the div. We can try to brainstorm about whether it'd be useful to be able to give two possible elements.

Also, since this only takes one possible outcome or `null`, should this have a different name? `<OrNull>`?

ethanbb:
You could just make two of these components, with opposite conditions, if you wanted to have different non-null contents depending on whether a condition is true or false.

Still thinking about the name...

PR #588, opened 2018-06-13
Makes changes for lint msgs issue #507 - generateTestCases.js

ethanstrominger:


src/test/programs/federal/snap/generateTestCases.js
```diff
@@ -99,6 +99,8 @@ const utilityBrackets = (changes) => {
   case 'Telephone':
```
knod: Not a comment on this PR in particular, but shouldn't all these `case` statements require an indentation? @ethanbb or @wacii - thoughts? Is eslint failing us here or is this a convention I'm unaware of?

ethanbb: It is a convention I've seen, specifically for switches. Probably because the inside of the switch is technically just a regular code block with some jump labels (which are actually _outdented_ relative to the code).

wacii: @knod (indent)[https://eslint.org/docs/rules/indent] refer to the `"SwitchCase"` option. It defaults to 0.

knod: Alright, we'll leave this for a later discussion. Someone remember this!

PR #589, opened 2018-06-13
Makes changes for lint msgs issue #507 - Household.js

ethanstrominger:


src/forms/Household.js
```diff
@@ -214,7 +222,8 @@ const MemberField = function ({ household, time, setHousehold, setClientProperty
             className={ 'remove' }
             onClick={ removeMember }
             iconName={ 'remove' } />
-          : <span>{ household.length > 1
+          : 
+          <span>{ household.length > 1
             ? <Icon
```
ethanbb: Let's wait on the pending decision about avoiding ternary operators before merging this - we might want to do it differently.

knod: Hmm, how about we accept all of these fixes and then just start going through the code to abstract the ternary operators. I think that would simplify things a lot.

PR #591, opened 2018-06-13
Makes changes for lint msgs issue #507 - FMRS_MA_2018.js

ethanstrominger:


ethanbb:
@knod oh you're right actually, I agree.

ethanstrominger:
@ethanbb  I made the change.

knod:
@ethanstrominger : btw, you can have a different commit message for subsequent commits. Something that's descriptive about that specific commit.

PR #593, opened 2018-06-14
Fixes #542 - move logo to bottom right

### LINKED ISSUE #542: make top left nav home to app; move CfB logo
Common design patterns, noticed when an fellow member clicked it to try to go to the app home

misspran:
Moved Code for Boston logo from header to bottom footer right

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,9 +75,9 @@
       }
     },
     "address": {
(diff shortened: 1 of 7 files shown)


knod:
From discussions: The plan is this will be split into two commits and `package.json` and `package-lock.json` won't be changed.

PR #598, opened 2018-06-15
Fixes #595 - graphs don't match each other

### LINKED ISSUE #595: Graphs have different data than each other
### Replicate:
1. Go to https://codeforboston.github.io/cliff-effects/#/visit/load in one window
2. Go to https://docs.google.com/spreadsheets/d/15LyR9yELAfcngj-c7vMdI630b6DwuuXV-dQEJDvU4gE/edit#gid=1140929936 in another tab and copy the text in the 'Client Data' column for the second row.
3. Go back to that first page and paste the data in the top input field.
4. Click 'Import Data'
5. Go to the 'Predictions' section.
6. Compare the table and bar chart data to the other two graphs.

### Result I'd Expect:
- The table and bar chart show Section 8 values. I'd expect the stacked area graph and the benefit line graph to show Section 8 still existing where the 'Future Income' line is.

### Actual Result:
- In the stacked area graph and the benefit line graph, Section 8 dies out way before the 'Future Income' line.

### Environment:
Browser type and version: Chrome v66.0.3359.139

App version or commit id at the time: b90ec11123c461fde7ed46e8be5ef94a64186018

knod:


knod:
@wacii : Thanks! Also, `client` there is a clone and  Section 8 needs to use both `current` and `future` values. Does that answer any of the concerns?

PR #601, opened 2018-06-16
Refactor graphs (PR 1)

knod:


ethanbb:
Great, this looks like a good thing to factor out! Is the plan to move the other graph files to the `output` folder as well?

PR #603, opened 2018-06-17
Fixes #550 - Updates for buttons and last reset button

### LINKED ISSUE #550: consider text change "reset"

knod:


knod:
It's a temporary fix, unfortunately, but I don't see a reason to pursue it for now unless someone especially wants to, or unless we start using these buttons for something new.

ethanbb:
Oh I wasn't saying it's bad if it's a temporary fix, I was just saying you should add a comment if it's temporary...

PR #604, opened 2018-06-17
Refactors graph functionality (PR 3)

knod:
I _think_ it's not broken on this commit.

### DIFF
--- a/src/utils/convert-by-timescale.js
+++ b/src/utils/convert-by-timescale.js
@@ -87,7 +87,7 @@ timescaleMultipliers.fromMonthly = {
 timescaleMultipliers.fromWeekly = {
   'Weekly':  1,
   'Monthly': (4 + 1 / 3),
-  'Yearly':  12,
+  'Yearly':  52,
 };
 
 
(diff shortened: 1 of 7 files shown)


src/utils/convert-by-timescale.js
```diff
@@ -70,9 +70,30 @@ var toWeeklyFrom = function (amount, startTimescale) {
...
+timescaleMultipliers.fromMonthly = {
+  'Weekly':  1 / (4 + 1 / 3),
+  'Monthly': 1,
+  'Yearly':  12,
+};
+
+timescaleMultipliers.fromWeekly = {
+  'Weekly':  1,
+  'Monthly': (4 + 1 / 3),
+  'Yearly':  12,
```
ethanbb: This should be `12 * (4 + 1 / 3)` (or just 52).

knod: Well shucks. Yeah, you're right. It's not exactly 52 - it's 52.14 and change, but maybe that's the way they do it anyway.

ethanbb: 12 * (4 + 1/3) does equal 52 though...

knod: Well, then I'm probably hallucinating. I got that number somehow, but not sure how ¯\_(ツ)_/¯

PR #605, opened 2018-06-18
Splits stacked area graph from graphs file

knod:
Not sure about this one - some graphs use the same imports, etc., and `<GraphHolder>` has a direct relationship with them. Splitting them up seems weird for that reason. On the other hand, the file gets super long the way it is, and each graph should be controlling some of its own stuff, and maybe separate files gets that across.

Thoughts?


knod:
Yeah, that's the plan. I already have a PR set up for `BenefitGraph` ;)

PR #608, opened 2018-06-19
Removes outdated, renames to 'GraphHolder' (PR 6)

knod:


knod:
For now, I think so. I have thoughts about bringing the bar chart and table into the `GraphHolder` fold, but that may happen some other time.

PR #610, opened 2018-06-19
Small prompt shifts. Changing some names, etc. (PR 2)

knod:
Some of these are longer and a bit more awkward, but I think it's easier to tell what they're doing.

Going to change `OnLeavePrompt` to `FeedbackPrompt` next. Figured it might be too confusing to do all in one step considering the previous `FeedbackPrompt` component.

Also, may need to move `FeedbackForm` out of prompts. Not exactly sure what the definition of a 'prompt' is.


src/components/prompts/ReactRouterConfirmLeave.js
```diff
@@ -2,26 +2,26 @@ import React from 'react';
 import { Prompt } from 'react-router-dom';
 
 /**
- * Customize React Router on leave prompt to use <OnLeavePrompt>
+ * Customize React Router on leave/unmount prompt to use <OnLeavePrompt>
```
wacii: Not sure what `<ReactRouter.Prompt>` has to do with unmounting.

src/components/prompts/ReactRouterConfirmLeave.js
```diff
@@ -2,26 +2,26 @@ import React from 'react';
...
  * Hijacks history to show a modal first before
  * letting history take its course.
  * 
  * @param props {object}
- * @param props.prompt {function}
+ * @param props.askForFeedback {function}
  * @param props.message {string} - Passed indirectly to <OnLeavePrompt>
  * @param props.confirmer {object} - For hijacking history
- * @param props.isBlocking {boolean} - Whether the prompt should be shown
+ * @param props.isBlocking {boolean} - Whether this prompt should be shown
```
wacii: This component doesn't return any visible component. It just sets up a callback. But maybe it is calling the component a prompt in the first place which is the problem. Or maybe I'm just being too picky.

src/containers/VisitPage.js
```diff
@@ -107,50 +108,63 @@ class VisitPage extends Component {
     this.setState({ client: nextClient });
   };
 
-  resetClient = () => {
+  resetClient = (shouldReset) => {
+
+    if (!shouldReset) {
+      return;
+    }
```
wacii: Calling `resetClient()` when you don't actually want to reset the client is a little weird. Not a big deal though.

src/components/prompts/ResetAnytime.js
```diff
@@ -7,7 +7,7 @@ const ResetAnytime = function (props) {
   return (
     <div>
       <Button
-        onClick={ props.resetClient }
+        onClick={ props.askToResetClient }
```
wacii: This is certainly a better name for it.

src/components/prompts/ErrorPrompt.js
```diff
@@ -5,14 +5,15 @@ import React from 'react';
  * @extends React.Component
  *
  * @param props {object}
- * @param props.callback {onLeaveCallback}
+ * @param props.callback {function}
```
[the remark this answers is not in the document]
knod: The callback is still used, though, right? Do you mean that documentation should be added to describe it?

src/components/prompts/ErrorPrompt.js
```diff
@@ -5,14 +5,15 @@ import React from 'react';
...
  * @param props.message {string}
- * @param props.prompt {prompt}
+ * @param props.askForFeedback {function}
  */
 class ErrorPrompt extends React.Component {
   handleError = ({ error }) => {
-    const { callback, prompt, ...promptProps } = this.props;
-    prompt(callback, promptProps);
+    // Does `promptProps` need to be changed to `rest`?
```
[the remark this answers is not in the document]
knod: Nice, I'll remove this line then.

src/components/prompts/ReactRouterConfirmLeave.js
knod: I thought this whole thing was triggered when `<VisitPage>` is about to unmount. Isn't that what 'leaving' is in React?

src/components/prompts/ErrorPrompt.js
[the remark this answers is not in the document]
wacii: Yeah, I'm just talking about the documentation. `onLeaveCallback` is documented elsewhere and specifying the name here would link to that definition...if we were using jsdoc to process these comments.

src/components/prompts/ReactRouterConfirmLeave.js
knod: Yeah, I'm still not sure about the definition of prompt. Maybe it's a word we should move away from. Though the folder is called 'prompts'...

Would this work better? 'Whether to interrupt the user's navigation.'

src/containers/VisitPage.js
knod: Yeah, I know. I was hoping to save on duplication or on having to make a `resetIfOk` method or some such. Never came to a solid conclusion on that.

src/components/prompts/ReactRouterConfirmLeave.js
wacii: Ah. I used "leave" as a shortcut for "navigate away from". Just saying `<ReactRouter.Prompt>` would have probably been better in the first place.

src/components/prompts/ResetAnytime.js
knod: 🎉

src/components/prompts/ErrorPrompt.js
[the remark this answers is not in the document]
knod: Oh! I wondered how we could do that with jsdoc. Had no idea that's what it did! That would be awesome! I'll change that back in that case.

src/containers/VisitPage.js
knod: Maybe I'll just change the name of that function itself, lol

PR #611, opened 2018-06-19
Localization current benefits

ValerieKenyon:
Adds text strings for the "Current Benefits" page.


src/localization/en.js
```diff
@@ -117,14 +117,19 @@ export default {
     },
 
     currentBenefits: {
-      hasSection8: {
+      currentBenefits: 'Current Benefits',
+      selectBenefits:  'Select the benefits you currently receive',
+      yes:             'Yes',
+      no:              'No',
```
ethanbb: The "yes" and "no" strings are actually provided by `ControlledRadioYesNo` - meaning they're in `formHelpers.js` rather than `CurrentBenefits.js`. So you can remove them from here for now, and we'll make a separate section for formHelpers strings. Otherwise looks good.

ethanbb:
Superseded by #612.

knod:
Do we not need the other strings for Current Benefits?

knod:
My bad - #612 includes changes for Current Benefits as well.

PR #612, opened 2018-06-20
Localization formHelpers and CurrentBenefits

ValerieKenyon:
Adds text strings for formHelpers to en.js; 
Removes yes/no from my previous PR (#611 Localization current benefits)


src/localization/en.js
```diff
@@ -116,15 +116,27 @@ export default {
       predictions:     'Predictions',
     },
 
+    formHelpers: {
+      type:    ' Type',
+      weekly:  'Weekly',
+      monthly: 'Monthly',
+      yearly:  'Yearly',
+      yes:     { label: 'Yes' },
+      no:      { label: 'No' },
```
ethanbb: We could just make these `yesLabel: 'Yes'` and `noLabel: 'No'`, for simplicity. I do think having 'label' in there is a good idea though, to emphasize that we would only be putting it in the `Radio`'s `label` prop and not `value`.

src/localization/en.js
```diff
@@ -116,15 +116,27 @@ export default {
       predictions:     'Predictions',
     },
 
+    formHelpers: {
+      type:    ' Type',
```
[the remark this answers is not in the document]
knod: That's a good point, and definitely not what I was considering. That change would take a bit of a deeper dive into the code, though. Maybe we pull this and you create an issue for that @ethanbb, that one of us gets to right quick?

Also, do we create a component for 'Income Type' and 'Expense Type'? Or does the form section component set that?

src/localization/en.js
knod: First, why is there an object instead of just a word? Is this in the model of `label` and `hint`? I'm trying to think if that makes sense considering we're not going to have a hint for these. Is there an argument for consistency here?

Second, I agree that the key may be missing something, but I'm not sure about `Label`. Do we make that a standard for all labels? If not, why here and not for others?

It's occurred to me that, in general, having the key of the snippet be the same as the text of the snippet seems not quite right. Does it only describes the contents of the snippet, not the purpose? That would seems fragile. What if the contents changes? Will we need to change the key? Or is it that way because that really is the most succinct and clear way to describe the purpose of the snippet?

I know it can sound purely philosophical to some people, but I've found that when variable names go wrong, they make the code almost unreadable and untraceable.

src/localization/en.js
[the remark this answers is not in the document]
ethanbb: I was just suggesting passing the whole string "Income Type", "Expense Type".... in as a property (through the snippets on each page) rather than just "Income" or "Expense." We wouldn't pass in the whole header component.

I can make an issue for it after this is merged. It's not an issue for this PR, just something related that I noticed.

src/localization/en.js
ethanbb: I think "consistency" could actually be confusing here. The "labels" here are not doing the same job as the labels on an instance of a form component; they're really just part of the radio button control, and there's a separate label which describes what the radio buttons are for. We don't have to put `label` in the key. I don't have a strong opinion on what the keys should be; could be something involving `true` and `false` rather than `yes` and `no`, maybe.

knod: Hmm, what's the rationale behind `true` and `false`?

src/localization/en.js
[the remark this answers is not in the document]
knod: I'm not saying we'd make a new component on each page. I was wondering if we would make two components, both in 'formHelpers.js' for example, that could be used repeatedly, in case we change something important about the 'Income Type' and 'Expense Type' so that it won't have to be changed in multiple snippets.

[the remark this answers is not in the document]
ethanbb: OK I see. What if instead of doing that, we just put those strings in a different section of the snippets, maybe just under `visitPage`, and pull them from there on all the pages that need them? That way we don't have to make any special functionality within the component for those strings, but we're still not repeating ourselves.

Something like
```
visitPage: {
  incomeHeader: "Income Type",
  expenseHeader: "Expense Type",

  stepBar: {
...
```

Or we could also put them under a key like 'common' or 'shared' if we want to keep everything at the same level of nesting.

src/localization/en.js
ethanbb: It corresponds the value of the boolean that the radio button controls. That should never change. 

(Maybe something like `trueLabel` and `falseLabel`? Why shouldn't we use the word "label" again?)

src/localization/en.js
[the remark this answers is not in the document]
knod: How would the form sections get access to those properties? They only have the snippets for themselves, they don't have the `VisitPage` snippets. I'm not sure I want to hand down another property to each form section. It might be most simple, in the end, to make a component for those two things.

src/localization/en.js
knod: I hear what you're saying. That does correspond with the radio button value, but these snippets aren't for the value, they're for the label, so they don't actually represent the values `true` and `false`.

I wasn't necessarily saying to not use the word 'label'. It was more of a question about how we're going to be deciding keys in general. Other labels don't have 'label' in their name, though, honestly, other labels' keys are `label`. If it helps, it's like naming your variable `cooridnateArray` or something - labeling types instead of purpose - but I've always looked at it as a general guideline, not a rule. I think if you can't find anything better and that kind of language clarifies what's going on, than that's what you use.

Just to check in here, maybe just `yes` and `no` are fine. I never stopped to ask why those weren't sufficient. What was the downside?

src/localization/en.js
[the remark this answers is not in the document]
ethanbb: Oh good point. So to be clear, when you say "make a component for" Income type and expense type, do you mean specialized versions of `IntervalColumnHeader` with those headings plugged in, or do you mean components that *just* render the heading text? (The first one makes somewhat more sense to me.)

src/localization/en.js
ethanbb: I'm fine with `yes` and `no`, but you said

>It's occurred to me that, in general, having the key of the snippet be the same as the text of the snippet seems not quite right.

so I was trying to think of other names. If the snippet is so short that it would also work well as a key, I think it makes sense to use it as a key, as long as the meaning is clear.

src/localization/en.js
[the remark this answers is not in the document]
knod: I'm not really attached to an implementation right now. I can see how specialized versions could be clearer to read elsewhere, though it'd be good to do it in a way that avoids duplicate code.

src/localization/en.js
knod: Yeah, this might be a case where the purpose and the text are the same. As for it being confused for the value instead of the label, maybe that's a good point. I'd assumed the coder using the snippets would be aware of how the snippets should only be used for visible text, but maybe that's not a great assumption. Don't know why I didn't consider that before. Hmm, you have some good points.

knod: You know what, forget everything I said. Well, don't forget it. The discussion about purpose was one I found useful. Where I'm at right now, though, it looks like your proposal of `yesLabel` and `noLabel` sounds like the way to go.

src/localization/en.js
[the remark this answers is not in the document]
knod: So, that said, should we make an issue for the implementation of that and just leave this snippet as is for now?

src/localization/en.js
ethanbb: OK, you had good points too! But I do still like `yesLabel` and `noLabel`. So @ValerieKenyon, could you please change lines 124 and 125 to
```
yesLabel: 'Yes',
noLabel: 'No',
```
Thanks!

src/localization/en.js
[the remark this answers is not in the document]
ethanbb: Well, let's get rid of this snippet for now, since we're not planning to use it after we make that change (we'll either use snippets on individual pages or 'Income Type'/'Expense Type' snippets that will be added). So @ValerieKenyon can you get rid of this line with `... ' Type'` please?

src/localization/en.js
ethanbb: Thanks but actually lines 124 and 125 are still not quite right. We want them to just be strings - `'Yes'` and `'No'` rather than `{ label: 'Yes' }` and `{ label: 'No' }`. Could you change that?

ValerieKenyon: So sorry @ethanbb , I just fixed it.

PR #613, opened 2018-06-21
Changes `OnLeavePrompt` to `FeedbackPrompt`

knod:


knod:
Hmm, I think this wasn't up to date with the latest dev. Hang on a sec.

knod:
I am very confused.

knod:
I apologize. I obviously thought I'd added changes that I hadn't.

src/containers/VisitPage.js
```diff
@@ -297,7 +297,7 @@ class VisitPage extends Component {
 
         <ConfirmLeave isBlocking={ this.state.isBlocking } />
         <FeedbackForm
-          isOpen={ this.state.feedbackOpen }
+          isOpen={ this.state.shouldShowFeedback }
```
ethanbb: What's the reason for this change? "should show feedback" seems a little more vague to me than "feedback open" (the word "open" indicating that this is describing whether something is currently open or not, not just whether it "should" be).

knod: So it's a fuzzy line, and here's my reasoning for what I did. I called it `shouldShowFeedback` because it isn't an indication that the feedback form is open. It happens before the feedback form opens and then the feedback form looks at it and decides for itself whether it wants to be open or not. Maybe it should be called `userWantsToGiveFeedback` or something that's less long that I can't figure out right now because I'm too tired. Does that make sense?

ethanbb: Hm, maybe... It seems like in this case, the feedback form really is open if and only if `feedbackOpen` is true. But I guess you're saying that doesn't have to be the case? It sounds like you're trying to think of changing a state value as passing a message to components that care about it rather than mutating the whole app instantaneously, which is probably more accurate with what actually happens but also breaks the React abstraction in a way that I'm not sure is necessary here.

knod: Just for the sake of posterity, the conclusion elsewhere was to call it `feedbackFormRequested` or something more like that.

knod:
Apology not accepted! That was a great conversation! And I think it was one worth having.

PR #614, opened 2018-06-23
Localization current expenses

ValerieKenyon:
Adds some text strings to en.js for formHelpers and currentExpenses.
Also adds a few lines for currentBenefits.
Please check lines 208 and 230 - I included 'Expense' to match up with ' Type' in line 120.
Please also check line 226 - this is for 'Income made possible by childcare expenses'.


ethanbb:
We should wait to merge #612 before this one, since this also includes the `formHelpers` and `currentBenefits` snippets.

src/localization/en.js
```diff
@@ -188,7 +200,65 @@ export default {
...
+        childTransportation: {
+          label: 'Transportation costs',
+          hint:  'How much do you pay for transportation?',
+        },
+        childOtherCare: {
+          label: 'Other care',
+          hint:  'How much do you pay for other child care?',
+        },
+        earnedBecauseOfChildCare: 'Does child care allow you to make additional income?',
+        ifYes:                    'Income made possible by childcare expenses',
```
ethanbb: How about putting these two in a sub-object called `earnedBecauseOfChildCare` (you could call the question itself something else) and also including a hint, like the other rows?

src/localization/en.js
```diff
@@ -188,7 +200,65 @@ export default {
...
+      },
+      childSupport: {
+        sectionHeading:      'Child Support',
+        columnExpenseType:   'Expense',
+        childSupportPaidOut: 'Legally obligated child support',
```
ethanbb: We should indicate that "Legally obligated" is bold somehow, either using interpolation of a `<strong>` tag (like in the about page), or maybe using Markdown as I suggested in slack.

src/localization/en.js
```diff
@@ -188,7 +200,65 @@ export default {
...
+        sectionHeading:      'Housing',
+        monthlyContractRent: {
+          label: 'Monthly Contract Rent (the total rent for your apartment)',
+          hint:  'The total rent for your apartment',
+        },
+        monthlyRentShare: {
+          label: 'Your Monthly Rent Share (how much of the total rent you have to pay)',
+          hint:  'How much of the total rent you have to pay',
+        },
+        subheading:     'Which of these utilities do you pay for?',
```
ethanbb: Could we call it something more specific like `utilitiesSubheading`?

src/localization/en.js
```diff
@@ -188,7 +200,65 @@ export default {
...
+        phone: {
+          label: 'Telephone service',
+          hint:  'How much do you pay for basic telephone service',
+        },
+        fuelAssistance: {
+          labelText: 'Do you get Fuel Assistance?',
+          hint:      'Fuel Assistance helps you pay for heating fuel',
+        },
+      },
+    },
```
ethanbb: We should add snippets for the "Previous" and "Next" buttons as well.

src/localization/en.js
knod: Indeed - that already is in its own node. I'd suggest interpolation for now if we want to get this done more quickly.

src/localization/en.js
knod: Yeah. In future we might think of abstracting that text out, though, into their own components, since that's repeated in almost every section.

src/localization/en.js
knod: What's the reasoning behind that design decision? Right now it's just two questions - one of them just reveals the other, like many questions - and I don't see why we'd put two questions into a new object of their own.

I can see how this would be a sub-section of a form in some ways (and one of the unfinished discussions is about if this should actually be a sub-section). So far we haven't created objects for sub-sections. We could possibly talk about breaking out all sub-sections into objects. I'm not sure how much more nesting we want. What are the pros and cons?

knod:
As a general note, try keep each PR to one thing. If you're editing the snippets for one section, don't include edits for snippets in another section too.

src/localization/en.js
```diff
@@ -116,15 +116,27 @@ export default {
       predictions:     'Predictions',
     },
 
+    formHelpers: {
+      type:    ' Type',
```
knod: @ValerieKenyon : Could you undo the changes you made to parts that weren't in `currentBenefits`? You can just re-type them as you currently see them in the dev branch.

src/localization/en.js
knod: I agree, that would be more clarifying.

src/localization/en.js
ethanbb: OK, I was thinking of the radio button and income row as one unit, since the radio is just there to draw attention to / introduce the income row. If we are going to treat it as its own thing though, then I think the income row about earned income because of child care should be called something more descriptive.

We could use `doEarnBecauseOfChildCare` for the radio button and `earnedBecauseOfChildCare` for the amount, maybe?

src/localization/en.js
ValerieKenyon: Just did that.  Hope I did it okay.

src/localization/en.js
knod: Yeah, that makes sense. Something like that sounds good.

src/localization/en.js
knod: @ValerieKenyon : That means add snippets for 'Previous' and 'Next' for now.

knod:
Looks good to me. @ethanbb ?

PR #615, opened 2018-06-24
More attempts at clarifying prompt related functionality

knod:


knod:
@wacii , @shp54 , @ethanbb : Any thoughts?

knod:
Yeah, we should go over this sometime so that more of us have a mental model of what's going on. Thanks for checking it out!

PR #618, opened 2018-06-26
Allows template literals. Maybe use these instead of quotes?

knod:
Pros/cons?

I'd like to switch all the localization files to use back ticks, btw. Makes a lot of things more simple.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -37,7 +37,7 @@
     "object-curly-newline": ["warn", { "multiline": true }],
     "object-curly-spacing": ["warn", "always", { "arraysInObjects": false, "objectsInObjects": false }],
     "object-property-newline": ["warn", { "allowAllPropertiesOnSameLine": true }],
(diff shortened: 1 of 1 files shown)


knod:
Wow, I didn't know @shp54 was working on that! Great!

Ideally, I'd like others interested in formatting to weigh in too before merging, if possible. Like @nicklee100 and maybe @wacii.

wacii:
Template strings do make interpolating much cleaner. But I didn’t realize the quotes rule disallows them by default. I would not have expected that.

As to whether we use template literal exclusively, it doesn’t matter to me.

knod:
Maybe we make it the norm, but don't require only template literals. No idea why we'd want to use quotes instead, but I'm new to template literals, so I don't want to shut that door.

Reminder: We should add a beginner task to replace all quotes with template literals.

knod:
Ok, it's been up here long enough. I'm calling it!

PR #619, opened 2018-06-26
Fixes #450 Makes sub-section for income earned because of care

### LINKED ISSUE #450: Break 'income from x' into their own sections.

knod:
Also changes to language that's more simple. Sorry @ValerieKenyon.
It won't interfere with your changes directly, but the text you
wrote will have to be adjusted somepoint soon, though wait till
after we discuss whether that lanugage adjustment makes sense.

### DIFF
--- a/src/forms/CurrentExpenses.js
+++ b/src/forms/CurrentExpenses.js
@@ -317,7 +317,7 @@ const ExpensesFormContent = function ({ current, time, setClientProperty }) {
             cashflowProps ={{
               ...sharedProps,
               generic:      'earnedBecauseOfChildCare',
-              confirmLabel: 'If you didn\'t have that child care, would it affect your current job?',
+              confirmLabel: `If you didn't have that child care, would it change how much money you can bring home?`,
             }}>
             How much less would you make?
           </EarnedFrom>
@@ -383,7 +383,7 @@ const ExpensesFormContent = function ({ current, time, setClientProperty }) {
             cashflowProps ={{
               ...sharedProps,
               generic:      'earnedBecauseOfAdultCare',
-              confirmLabel: 'If you didn\'t have that assistance, would it affect your current job?',
+              confirmLabel: `If you didn't have that assistance, would it change how much money you can bring home?`,
             }}>
             How much less would you make?
           </EarnedFrom>
--- a/src/forms/CurrentExpenses.js
+++ b/src/forms/CurrentExpenses.js
@@ -38,6 +38,18 @@ import {
 // ========================================
 // COMPONENTS
 // ========================================
+const EarnedFrom = function ({ heading, cashflowProps, children }) {
+  console.log('heading:', heading)
+  return (
+    <div>
+      <FormHeading>{ heading }</FormHeading>
+      <CashFlowRowAfterConfirm { ...cashflowProps }>
+        { children }
+      </CashFlowRowAfterConfirm>
+    </div>
+  );
+};  // End EarnedFrom
+
 
 const Utilities = function ({ current, type, time, setClientProperty }) {
 
@@ -267,7 +279,7 @@ const ExpensesFormContent = function ({ current, time, setClientProperty }) {
         ? 
         <div>
           <FormHeading subheading = { 'A "child" is a person 12 or younger. Don\'t include amounts that are paid for by other benefit programs.\n' }>
-            Reasonable Unreimbursed Non-Medical Child(ren) Care
+            Reasonable Unreimbursed Non-Medical Child Care
           </FormHeading>
           <IntervalColumnHeadings type={ type } />
           <CashFlowRow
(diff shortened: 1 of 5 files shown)


ValerieKenyon:
@knod don't be sorry.  I fully expect most of my work to have to be changed, fixed, etc.  About to start reviewing.

src/forms/CurrentExpenses.js
```diff
@@ -286,12 +297,16 @@ const ExpensesFormContent = function ({ current, time, setClientProperty }) {
...
           </CashFlowRow>
-          <CashFlowRowAfterConfirm
-            { ...sharedProps }
-            generic={ 'earnedBecauseOfChildCare' }
-            confirmLabel={ 'Does childcare allow you to make additional income?' }>
-            <span style={{ textDecoration: 'underline' }}>Income</span> made possible by child care expenses
-          </CashFlowRowAfterConfirm>
+
+          <EarnedFrom
+            heading={ 'Effect of Childcare' }
```
ethanbb: Are we treating "childcare" as one or two words? Should be consistent. A quick google search seems to indicate "child care" is more common.

src/forms/CurrentExpenses.js
```diff
@@ -286,12 +297,16 @@ const ExpensesFormContent = function ({ current, time, setClientProperty }) {
...
+            cashflowProps={{
+              ...sharedProps,
+              generic:      'earnedBecauseOfChildCare',
+              confirmLabel: 'If you didn\'t have that childcare, would it affect your current job?',
```
ethanbb: I'd get rid of the word "that," since it sounds like you're assuming they do get child care currently. Also maybe replace the word "job" with "income" to make it a little less vague.

src/forms/CurrentExpenses.js
```diff
@@ -348,12 +363,16 @@ const ExpensesFormContent = function ({ current, time, setClientProperty }) {
...
-            confirmLabel={ 'Do assistance expenses allow you to make additional income?' }>
-            <span style={{ textDecoration: 'underline' }}>Income</span> made possible by assistance expenses
-          </CashFlowRowAfterConfirm>
+
+          <EarnedFrom
+            heading={ 'Effect of Assistance' }
+            cashflowProps={{
+              ...sharedProps,
+              generic:      'earnedBecauseOfAdultCare',
+              confirmLabel: 'If you didn\'t have that assistance, would it affect your current job?',
```
ethanbb: Again I'd change it to "income"

src/forms/CurrentExpenses.js
```diff
@@ -348,12 +363,16 @@ const ExpensesFormContent = function ({ current, time, setClientProperty }) {
...
+            }}>
+            How much less?
```
ethanbb: I think we should change this to "How much less would you make?" for consistency.

knod:
I suppose my rationale for making it a whole section of its own is that I'm not convinced people will take notice of it the way it was before. I'm not sure I want to add a third styled element into the mix, though. What we probably need is some user testing, but that hasn't been in abundance. Any ideas?

src/forms/CurrentExpenses.js
knod: Hmm, what we need here is to only show that option if the person has child care. First, then the 'that' would make sense. Second, it really doesn't make sense to ask that if there's no child care expense. It also might be a good way to get them to notice the question about income earned from child care. I was hoping to avoid that annoyance, but maybe we'll need to add it in.

The reason it's 'that' child care is because people also get their child care subsidized, but we're not asking about the income they make from subsidized child care, just the child care they listed above.

We used 'job' instead of 'income' because it's a word with which more people are familiar. I have to admit it doesn't feel vague to me. The way I think about it, having to work fewer hours or not being able to keep your job seem to be well covered by it, but maybe I've been playing with the wording too long to see it objectively. We played around with the wording for a while, but there are probably better options. Can you think of any that stick to plainer language?

src/forms/CurrentExpenses.js
knod: Oh, shoot. Definitely. I just forgot to update that.

ethanbb:
I think this way is OK if you don't want to use another level of headings. That would be my preference though. I am also not a designer.

src/forms/CurrentExpenses.js
ethanbb: The reason it feels vague to me is someone might think "well my job would be the same, I'd just work fewer hours" and answer no. I honestly think "income" is a pretty simple word for someone who has a job. Would "wages" be better? Or "pay"?

knod: I don't like the word 'affect' either, btw. Still trying to figure out that wording.

knod: Great words! I just had a thought, though. How about, 'would it change how much money you bringing home'?

knod: Dang, this is an 'outdated' hidden thread now. Damn you, github!

knod:
I have no idea what happened to the text I wrote, but this was my proposal:

Really, this second thing shouldn't appear unless the client says they have expenses. I was trying to avoid that, but it's probably what needs to happen. What if we don't give it a new section and set it up the right way, and then color it in some different color, like red or something. I haven't been able to think of a color other than red, to be honest.

ethanbb:
@knod Yeah that makes sense to me. Red seems harsh though, why not yellow?

I feel like seeing something pop up after you change something on the form is a pretty good indication that you should pay attention to it (especially if it's colored).

knod:
@ethanbb : Yellow's a bit hard to read on white. I don't think it has to be red, I just don't know what other color would be good. Introduce a new green? A blue?

ohmegasquared:
I would recommend checking the colors on something like
https://contrastchecker.com/

knod:
@ohmegasquared : Any ideas for the design of this feature? Colors, for example?

ethanbb:
@knod Oh I didn't realize you meant changing the text color, I thought you meant put the part that appears in a colored container.

knod:
Hmm, a colored container. Will have to think about that. For me, colored containers bring to mind warnings, info blurbs, and asides - informative stuff as opposed to items that need action.

Another idea is an orange arrow/triangle to the left, pointing at the question. Here's the result of 20 min. of work: 
arrow

knod:
Different words plus an attention 'arrow'. This is what it should look like now:

attention-arrow

knod:
Dangit, hold on. It's not opening up well.

knod:
There, that works now.

PR #620, opened 2018-06-30
Moves responsibility for rendering to parent and puts bools...

knod:
for 'is this being rendered?' logic in the same location -  as
sibling properties.

### DIFF
--- a/src/components/prompts/FeedbackForm.js
+++ b/src/components/prompts/FeedbackForm.js
@@ -80,7 +80,7 @@ class FeedbackPrompt extends React.Component {
     return (
       <Modal
         size='large'
-        open={ this.props.isOpen }
+        open={ true }
         onClose={ this.close }
         closeOnDimmerClick={ false }
         closeOnEscape={ false }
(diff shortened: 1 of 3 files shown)


ethanbb:
Wait why are we putting the prompts inside of `RenderIfTrue`s? I thought the point of that was to get rid of ternary operators?

knod:
Where's the ternary? I don't think I used one. I conditionally render them in the parent so the children don't have to decide whether to render themselves. Like you said, that shouldn't be the job of the children. That conversation didn't get in here, I guess.

ethanbb:
No there's no ternary here, I was just saying, I thought the point of `RenderIfTrue` was just to get rid of ternaries.

I think that since the `Modals` already have an `open` prop, it makes sense to let that control whether they're visible. Seems like that's how it was designed to work. I was just talking about whether the variable name controlling whether a prompt is open should imply that it's controlling that property directly or not.

knod:
After some conversation, it sounds like this model doesn't fit into what we're doing. Might do something similar, but different.

PR #621, opened 2018-07-02
make top left nav home to app; move CfB logo  closes #542

### LINKED ISSUE #542: make top left nav home to app; move CfB logo

misspran:


src/components/MainMenu.js
```diff
@@ -17,15 +17,7 @@ const MainMenu = function (props) {
         inverted
         secondary
         size='large'>
-        <Menu.Item>
-          <a
-            href="http://www.codeforboston.org"
-            target="_blank"
-            rel="noopener noreferrer"><Image
-              src={ logo }
```
ethanbb: You can get rid of the `logo` and `Image` import statements in this file (lines 5 and 10) now that they're not being used here. I believe that's what's currently causing the Travis build to fail.

misspran: Thanks! Forgot about those.

src/components/Footer.js
```diff
@@ -20,7 +22,7 @@ const Footer = () => {
         inverted
         stackable>
         <Grid.Row>
-          <Grid.Column width={ 7 }>
+          <Grid.Column width={ 4 } floated='left' left>
```
knod: The attribute `left` is causing errors. I believe it can just be removed.

src/components/Footer.js
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
+import logo from '../images/logo.svg';
 
 import {
   Grid,
   Header,
   Icon,
   Segment,
+  Image
```
knod: Note for us for followup (no worries to @misspran, we can take care of this one if you want): This, along with a couple other small things, are causing lint warnings. Someone should run `npm run lint` when we get a chance.

knod:
Good PR! And thanks for fixing the import statements. A small change, and then I think we're good to go!

src/components/Footer.js
knod: (They show up in the browser console, not in the terminal)

misspran: The left thing is placed in there to keep the styling from Semantic UI React. I dunno at least that's why I did it that way without using custom CSS. Unless we prefer to use custom CSS I'll try to do that instead.

knod: Really? When I removed it, `floated='left'` seemed to be sufficient. Maybe I'm wrong, though. I've been pretty tired.

misspran: Oh hmm, maybe just the left at the end is needed then. Sounds good! I'll remove it.

knod:
Thanks!

knod:
I guess it's testing the build as if it were already merged, not just the PR build, because this PR build is passing on my local branch. The error looks the same as the one everything else is getting, so I'm just going to go ahead and merge this.

PR #623, opened 2018-07-03
ManagedNumberField cleaning up variable name and function call

snyderc:
When I was writing unit tests, I noticed a couple of seemingly redundant/unnecessary bits of code in the ManagedNumberField.

The handleBlur function appeared to first call this.props.onBlur -- which would be calling itself, right? Removing the call didn't seem to affect app functionality. When I made this change, I also cleaned up a line of code in the test that's no longer needed if this change is made.

handleFocus takes in inputProps but doesn't use it, so I suggest removing the parameter.

### DIFF
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -487,12 +487,6 @@ class ManagedNumberField extends Component {
   };
 
   handleBlur = (evnt) => {
-<<<<<<< HEAD
-=======
-    this.props.onBlur(evnt);
-
-    // Set local state for blur
->>>>>>> dev
     this.setState({ focused: false, valid: true });
   };
 
@@ -598,7 +592,6 @@ const CashFlowRow = function ({ generic, timeState, setClientProperty, children 
         displayValidator: hasOnlyNonNegNumberChars,
         storeValidator:   isNonNegNumber,
         format:           toMoneyStr,
-        onBlur:           function () { return true; },
       };
 
   return (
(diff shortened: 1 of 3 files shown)


shp54:
`inputProps` can be removed, but `this.props.onBlur` allows a developer to add a custom function to be called whenever the input blurs. I'm not super sure of the utility, but I'd recommend making it optional rather than removing it (only running that method if it exists)

knod:
Actually, we're really not using `this.props.onBlur`. If you look in `<CashFlowRow>`, it just returns `true`. I'm not sure why that happened. Maybe we were planning to store values that way, but it became moot. I think we can delete it now and add it later if needed.

As for `inputProps`, they're in a lot of places where they don't need to be. I was the main person putting that in and it was before I understood the React ecosystem at all. I think it'd be great to remove it wherever it isn't needed, though it would have to be a careful extraction because sometimes it is actually used. If no one wants to take on that burden, which I'd completely understand, I can work on it at some point.

Thanks for cleaning this stuff up!

knod:
@snyderc : Are you up for resolving the conflicts here? I think you just have to merge dev.

knod:
See this for desired behavior: https://github.com/codeforboston/cliff-effects/issues/630

knod:
I think I just rememberd our old use for `inputProps` - we were using `inputProps.value` before we found we could use `evnt.target.value`.

knod:
You could probably destroy the sources of the `onBlur` as well, btw, if you're up for it.

snyderc:
Resolved merge conflicts, and removed onBlur from CashFlowRow, with today's commits. Should be all ready unless there are other questions.

PR #625, opened 2018-07-03
ManagedNumberField bug fix - restores previous input after invalid entry

snyderc:
Per Slack channel discussion from today. Also updates a test that would break now that handleBlur sometimes needs the value from evnt.target.

Functionality Q:
Does the 2nd argument of store() require anything from inputProps other than value? I couldn't figure out how to pass inputProps into handleBlur() like it's done in handleChange() (although I'm still trying to wrap my head around what inputProps are and where they come from). So as a workaround, I am passing only the value (which in this case is hardcoded as 0 because it's the case of an empty string).

Style Q:
Is a blank entry considered valid? Right now, if you type 234 and click out, then click back in and empty the field, the field will show in error state until you click out and onBlur changes it back to not being in error state. Is that desired behavior? (If not, this can be handled either here or on a future PR.)

### DIFF
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -487,6 +487,13 @@ class ManagedNumberField extends Component {
 
   handleBlur = (evnt) => {
     this.props.onBlur(evnt);
+    // If field is empty string, set value to be 0 so that a user can clear the field
+    var { store, otherData } = this.props;
+    var value = evnt.target.value;
+    if (value.length === 0) {
+      store(evnt, { value: '0' }, otherData);
+    }
+    // Set local state for blur
     this.setState({ focused: false, valid: true });
   };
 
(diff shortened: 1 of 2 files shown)


snyderc:
For reference:

The goal of this PR was to address the following:
Expected behavior: If I type a 3 in, leave the field, and then type in something invalid, like a letter, and leave the field, it should then change to '3' again.
Actual behavior: When the invalid input is typed in, the program treats it like a 0 was entered.

Reason for the difference in behavior:
handleChange() had a line to treat an empty string like a 0. But if you put that in handleChange() which fires on each keystroke, if someone types a “g” or a “-” (to enter a negative number), the underlying FormInput reports a value of “” which must be triggering the empty string => 0 line.

Commenting that line out, I restore the expected behavior, with the exception that if someone actually wants to backspace to empty the field (instead of typing 0), they won’t get a cleared field. This PR takes care of that issue by moving the field-clearing check over to onBlur.

Issue #500, opened 2018-05-08
Localization


shp54:
Anyone who could help translate our labels into other languages.

Current language assignments:
Ethan - French
Valerie - German

Languages currently seeking (most commonly spoken languages in Mass. besides English)
Spanish
Vietnamese
Haitian Creole
Chinese

knod:
Eugene - Romanian (and another one?)

Leftere:
Romanian and Russian!

KevinSJ:
Hi there, I can help translate the Chinese label. Where should I start？ I saw a localization folder under src, is that all of them?

knod:
@KevinSJ: That would be amazing! All the ones we have so far are in indeed in that folder. The 'en.js' file is where the English snippets are being gathered. A warning - some of the text may go through some changes soon, but we'd be grateful for whatever you can do!

knod:
@KevinSJ: You'd copy 'en.js' and rename it to the language code for Chinese, then change the values from English phrases to Chinese phrases. If you're not familiar with programming, let us know and we'll help you get started.

KevinSJ:
@knod Thanks for the hint. I am actually a programmer, but it's my first time doing localization for a React project. :smiley: I will start working on it now,  let me know when the change happens so I can update the translation.

ethanbb:
@KevinSJ You can use these instructions as well: https://github.com/codeforboston/cliff-effects/wiki/How-to-translate-the-app-into-a-new-language (although they're aimed at non-programmers).

Thanks!

knod:
Thoughts:

1. One of the things I think could really help translators is to have them be looking at the text in an editor that has highlighting so that it's clear what is in quotes and what isn't. What if we make CodePens of 'en.j' and have people fork that and use it to create the translations.
2. Does anyone have good ideas about automatically detecting when the English version has been changed to make sure that the other versions get updated appropriately? Should we give version numbers to each of the keys or something, so that the tests we write will very easily catch when something's not in sync?
3. Should we use double quotes instead of single quotes so translators don't have to deal with escaping an apostrophe? Should we use template tick marks instead?
4. Is there a better place to have general discussion about various aspects of localization?

Responses welcome!

ethanstrominger:
[...]

PR #627, opened 2018-07-04
Localization currentIncome.js issue #500

ethanstrominger:


src/forms/CurrentIncome.js
```diff
@@ -66,53 +66,67 @@ const IncomeForm = function ({ current, time, setClientProperty }) {
       <IntervalColumnHeadings type={ type } />
 
       {/* All kinds of things need to be explained. */}
-      {/* @todo Change 'labelInfo' to visible blurb at top */}
+      {/* @todo Change 'labelInfo' to visible blurb at top - delete below labelInfo? */}
```
ethanbb: I would do this, it's old and we're no longer planning to use that string.

src/forms/CurrentIncome.js
```diff
@@ -130,11 +144,20 @@ const IncomeForm = function ({ current, time, setClientProperty }) {
 * @returns Component
 */
 // `props` is a cloned version of the original props. References broken.
+/** @todo  Make currentIncomeStep compatible with CurrentBenefitsStep? 
+ *         currentBenefits looks like this:
+ *         const CurrentBenefitsStep = ({ changeClient, nextStep, client, snippets })
+ * Which one is correct?
+*/
```
ethanbb: Either way works. The "destructuring" style in `CurrentBenefitsStep` (which assigns each field of the `props` object to a variable immediately) is nicer since it's more explicit, so we should probably switch to that at some point, but it doesn't have to be in this PR.

src/forms/CurrentIncome.js
```diff
@@ -130,11 +144,20 @@ const IncomeForm = function ({ current, time, setClientProperty }) {
...
 const CurrentIncomeStep = function (props) {
 
   const setTimeProp = getTimeSetter('current', props.changeClient);
+  const snippets = props.snippets;
 
-  /** @todo Are these titles accurate now? */
+  /** @todo Are these titles accurate now? 
+   * @todo Change phrase after title to snippets.title after changing en.js?
+   * @todo Change phrase after clarifier to snippets.clarifier after changing en.js?
+  */
```
ethanbb: Why don't you go ahead and add those into `en.js`, so that it all gets done in one pass?

src/forms/CurrentIncome.js
```diff
@@ -44,7 +44,7 @@ import { getTimeSetter } from '../utils/getTimeSetter';
 *
 * @returns Component
 */
-const IncomeForm = function ({ current, time, setClientProperty }) {
+const IncomeForm = function ({ current, time, setClientProperty, snippets }) {
 
   var type = 'income';
```
ethanbb: Just a note, we're going to change this once we've decided how to provide the standard "Income Type" and "Expense Type" headers to all the pages that need them.

src/forms/CurrentIncome.js
```diff
@@ -66,53 +66,67 @@ const IncomeForm = function ({ current, time, setClientProperty }) {
...
       </CashFlowRow>
       <CashFlowRow
         { ...sharedProps }
-        generic='otherIncome'> Other income
+        generic='otherIncome'>
+        { snippets.otherIncome.label }
       </CashFlowRow>
+      {/** @todo Change "This prototype..." to snippets.incomePrototypeCalcuationHelp after changing en.js 
+        * This message is buried.  What does it mean?
+      */ }
```
ethanbb: I believe it's supposed to reassure users who are expecting to input their SNAP benefit amount here, but can't because we're calculating it based on other things. We should probably make it look more like a "note" rather than a random line of text, though.

Can you add this in `en.js`?

src/forms/CurrentIncome.js
```diff
@@ -127,7 +127,7 @@ const IncomeForm = function ({ current, time, setClientProperty, snippets }) {
       {/** @todo Change "This prototype..." to snippets.incomePrototypeCalcuationHelp after changing en.js 
         * This message is buried.  What does it mean?
       */ }
-      <Form.Field>This prototype will attempt to make its own calculations for SNAP amount</Form.Field>
+      <Form.Field>{ snippets.bottomText }</Form.Field>
```
knod: Maybe this can be `snippets.snap` or something, so it's a bit more descriptive.

src/forms/CurrentIncome.js
knod: When you feel you've resolved this, @ethanstrominger , go ahead and remove this question.

src/localization/en.js
```diff
@@ -153,6 +153,9 @@ export default {
     },
 
     currentIncome: {
+      incomeHeader:    'Current Household Income',
```
knod: Sorry I missed this before somehow - `title` and `clarifier` seem more useful to me, as opposed to `incomeHeader` and `incomeClarifier`. If you want, we can discuss that more, though it may work better in real-time.

ethanstrominger:
@ethanbb @knod Changed to title and clarifier.  Please review.

src/forms/CurrentIncome.js
```diff
@@ -66,54 +66,63 @@ const IncomeForm = function ({ current, time, setClientProperty }) {
...
+        { snippets.alimony.label }
       </CashFlowRow>
       <CashFlowRow
         { ...sharedProps }
-        generic='otherIncome'> Other income
+        generic='otherIncome'>
+        { snippets.otherIncome.label }
       </CashFlowRow>
-      <Form.Field>This prototype will attempt to make its own calculations for SNAP amount</Form.Field>
+      <Form.Field>{ snippets.bottomText }</Form.Field>
```
knod: Apparently I forgot to post a comment on this - we can possibly change this later, but `bottomText` isn't very descriptive. What if we add text under it? How about something talking about 'snap'? I figure there's some time to fiddle with it because I'd rather wait till the `ManagedNumberField` error we introduced earlier is all sorted.

ethanstrominger:
@knod @ethanbb Ready for review.  Based on comment from knod that Previous and Next were going to change, I did not touch Previous and Next.  I don't know why it is showing 5 files changed, I only changed en.js and CurrentIncome.js.  If there are additional comments, if not show stoppers, can you approve and merge?  I want to avoid merge conflicts and I can do another round.

knod:
Actually, we need to get rid of the changes to the additional files. It looks like you may have merged misspran's changes into this branch at some point by accident. You can try to reset those changes with git if you want (I'd suggest checking out a branch off of this one to try it out), but it may be easiest to just make a new branch from dev, copy and paste the changes you want to keep into that branch, and make a new PR.

PR #629, opened 2018-07-04
Call empty string valid and store 0 on change

ethanbb:
Fixes the failure of one test from #622, which occurs because we were treating empty strings in a `ManagedNumberField` as invalid while it's focused, and only defaulting to '0' and storing the '0' on blur. This moves that logic to the `handleChange` function, so that empty strings are stored and validated as '0' while focused (meaning they will be seen as valid by `isPositiveNumber`), while keeping the empty string in `focusedVal`. I also modified the test to return `true` from the validation function and expect the field to be valid, since the call to `store` is only supposed to happen if it's valid.

### DIFF
--- a/src/forms/Household.js
+++ b/src/forms/Household.js
@@ -18,7 +18,10 @@ import {
 
 // COMPONENT HELPER FUNCTIONS
 import { getTimeSetter } from '../utils/getTimeSetter';
-import { isPositiveWholeNumber } from '../utils/validators';
+import {
+  isNonNegWholeNumber,
+  hasOnlyNonNegWholeNumberChars,
+} from '../utils/validators';
 
 // OBJECT MANIPULATION
 import { cloneDeep } from 'lodash';
@@ -42,24 +45,24 @@ const Columns = {};
 // `noMargin` is a bit hacky, but it'll do for now
 Columns.One = function ({ noMargin, children }) {
   var marginTop = columnStyle.marginTop;
-  if (noMargin) { 
-    marginTop = 0; 
+  if (noMargin) {
+    marginTop = 0;
   }
   return (<div style={{ ...columnStyle, marginTop: marginTop, width: '5em' }}> {children} </div>);
 };
 
 Columns.Two = function ({ noMargin, children }) {
   var marginTop = columnStyle.marginTop;
-  if (noMargin) { 
-    marginTop = 0; 
+  if (noMargin) {
+    marginTop = 0;
   }
   return (<div style={{ ...columnStyle, marginTop: marginTop, width: '20em', textAlign: 'left', paddingLeft: '1em' }}> {children} </div>);
 };
 
 Columns.Three = function ({ noMargin, children }) {
   var marginTop = columnStyle.marginTop;
-  if (noMargin) { 
-    marginTop = 0; 
+  if (noMargin) {
+    marginTop = 0;
   }
   return (<div style={{ ...columnStyle, marginTop: marginTop, width: '5em' }}> {children} </div>);
 };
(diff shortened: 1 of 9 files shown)


knod:
See this for desired behavior: https://github.com/codeforboston/cliff-effects/issues/630

ethanbb:
OK, after some frustrating testing, I realized that both of those problems come from setting the `type` of the input to `number`, which I didn't know was a thing.

knod:
Do we want them to be able to type regular letters into there? That wasn't originally something we wanted. If you want, you can keep going on this till it matches our desired behavior completely. I'm hoping to have it see if it can fit in with the current validators and fixers, etc., or to fit into a similar kind of structure. Does that make sense?

Note for our future dream selves: Maybe it'd be polite to let people put in commas when they want to. I think we're ok for now, though.

ethanbb:
Alright, I was going off of this:
>User types an INVALID number in the field: 
>Typing shows up as entered 
>Field gives 'invalid value' feedback (turns red)
Which I interpreted to mean we shouldn't interfere with what the user's typing. That seems like the most predictable way to do it.

Currently though I think the only letter that can be entered is "e", due to scientific notation. Maybe there's a way to disallow scientific notation for number inputs?

I don't know what this means:
>I'm hoping to have it see if it can fit in with the current validators and fixers, etc., or to fit into a similar kind of structure.

knod:
Hmm, that could be confusing. Maybe those definitions need to change a bit now that we are considering this new thing. What I'm saying is that I do think we need a 'text' field instead of a 'number' field, but that we should limit the input to a subset - digits or periods. The only invalid value (that I can think of so far) would be one with multiple periods. That's allowed because people need some flexibility sometimes and that flexibility makes sense.

Maybe we'll talk about the 'similar kind of structure' thing in real-time, so we can save some back-and-forth.

snyderc:
The failing test was also flagged in #633. Essentially, I wrote some handleChange tests based on the then-existing code and submitted PR #622, and then intentionally moved the empty string handler from handleChange to handleBlur and submitted PR #625 just for that (since PR #622 hadn't been approved when I submitted #625 later the same day). I should have spaced the PRs out so that the failed test would be avoided.

If the code from #625 delivers the functionality we want, I think updating the unit test is the more straightforward action as opposed to further code modifications. But if the functionality really should be incorporated in handleChange instead of handleBlur, then the intent behind this PR is how we should go.

knod:
@snyderc : got it, thanks for clarifying. In case you find it edifying, the desired behavior is written out here (though I'll try to go over it again later to make sure it still matches what we want): #630.

knod:
Heh, gotta love math language. Non-negative is more accurate and I don't feel strongly one way or another. Works for me!

I like the idea of those separate validators! Thanks!

An unrelated thing I realized (that may already be accounted for and should be included in the spec) - we shouldn't prevent people from entering multiple periods, but multiple periods should be shown as invalid.

ethanbb:
Hold on, I think the `MonthlyCashFlowRow` validators still need to be changed. I can look at it tonight.

snyderc:
Looks good to me. Thank you!

MichaelDimmitt:
Can confirm, test suite is passing. Looking forward to when this commit is ready to be merged @ethanbb. Then I will update the pr's I have open so they will pass TravisCI.

knod:
Haven't checked all behaviors thoroughly, but currently if I have multiple periods, exiting the field just gets rid of the periods, giving me, for example '112.00' from '1.12.00'. Is that what we want? Would it be preferable to have it just restore to the last valid value?

ethanbb:
@knod hmm, that sounds like a bug but I don't observe that behavior (for either example). What field were you editing?

knod:
An income field

ethanbb:
@knod Still not seeing it. Can you try again, and if it's still happening, send me a client data object for just before the change and tell me exactly what to change to replicate it? I don't think it should matter what other client data is entered for any given field (except section 8 rent), but just to be sure...

MichaelDimmitt:
@knob is there a way we can merge the pull request as is and tackle the issue you are describing that still needs to be recreated in a new pull request?

knod:
To sum up conversation elsewhere, it was behavior as expected, it just didn't feel intuitive, but I think we're not going to worry about that for now.

PR #631, opened 2018-07-04
Add CodeCov shield for create react app

MichaelDimmitt:
I will be adding to the documentation for this pull request after it has been merged or closed. To improve readability of the process for setting code coverage up on a create react app. 

link1: [code coverage from create react app issues page](https://github.com/facebook/create-react-app/issues/2825) 
link2: [code coverage from create react app docs](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#coverage-reporting)
link3: [use yarn test instead of npm test](https://github.com/facebook/create-react-app/issues/1363) 

Lastly, codecov needed to be installed to get the shield.
Codecov is a global npm package installed when running TravisCI or CircleCI that hosts your previous and current code coverage information to show the projects progression over time.
Running codecov after the files have been written to the coverage folder, sends the coverage folder info to the codecov website. This service is free for open source projects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ node_js:
 cache:
   directories:
     - node_modules
+install:
+  - npm install -g codecov
 script:
   - npm run build
-  - npm test
+  - npm test -- --coverage && codecov
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# cliff-effects [![TravisCI](https://travis-ci.org/codeforboston/cliff-effects.svg?style=shield)](https://travis-ci.org/codeforboston/cliff-effects)
+# cliff-effects  [![TravisCI](https://travis-ci.org/codeforboston/cliff-effects.svg?style=shield)](https://travis-ci.org/codeforboston/cliff-effects) [![CodeCov](https://img.shields.io/codecov/c/github/codeforboston/cliff-effects.svg)](https://codecov.io/gh/codeforboston/cliff-effects)
 > **cliff effect**: You are a person on government benefits, and you get a raise.  You're making more money!  But now that your income is higher, you don't make the cutoff for the benefits you receive.  Even though you're taking home more money, your situation is worse. Some of your benefits drop to nothing, or almost nothing. You've fallen off "the cliff."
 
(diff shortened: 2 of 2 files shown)


MichaelDimmitt:
to see this commit in practice checkout my dev branch:
https://github.com/MichaelDimmitt/cliff-effects

knod:
Sorry, hectic times for me. If no one else manages to catch this sooner, I'll run this over when I get a chance! Thanks!

wacii:
Looks good. I would probably leave out the `--coverage` flag, to reduce clutter in the Travis build, assuming it isn't required for something else. But that is a minor concern.

This would certainly be nice to have though, now that we have people working on tests.

knod:
@MichaelDimmitt : Thoughts on the `--coverage`? Pros and cons?

MichaelDimmitt:
@wacii, @knod, 
There are a few ways to handle --coverage. 
Apologies in advance for opening so many pr's for this enhancement. 
Only one needs to be chosen.

1) #631, this pull request: `npm run test -- --coverage` inside the .travis.yml
2) #632, same as this pr but adding coverage script to the package.json: `npm run coverage`.
3) #634, Change the test command in `package.json` to: `"test": "react-scripts test -- --coverage",`
then the .travis.yml only needs to perform `npm run test`

wacii:
Alright, I think I misunderstood exactly how this works. It is jest that collects the coverage information, the `--coverage` flag being the easiest way to turn that on, and codecov simply processes those reports and persists them online.

I was assuming the `--coverage` flag is what was causing the coverage report to be printed in the Travis build. Which it is, but it is also why that information is being collected in the first place. It looks like you can collect without reporting, but all those ways introduce complexity, so I am fine with the extra info in the build.

My preference would be for PR #632, but this one is fine too.

MichaelDimmitt:
@wacii, I too am a fan of #632, 

Potential pros:
1) Line coverage shown every-time can help determine what still needs testing.

Potential cons:
1) Line coverage being shown every time can be too much information.
2) When a test fails jest --coverage throws an npm error.

If your test suite fails without `--coverage` option there is no npm error.

PR #632, opened 2018-07-05
Add Code coverage command 

MichaelDimmitt:
Same as #631, but with an additional package json script 
For giving line coverage when the test suite runs.

If adding a new script is not useful. 
the test script itself can be changed to include coverage every time someone runs the test suite.

Options, 
1) #631, ```npm run test -- --coverage``` inside the .travis.yml
2) #632, this pull request: same as #631 but adding coverage script to the package.json: 
```npm run coverage```. Additionally the .travis.yml could ```npm run coverage``` but I feel that is against convention because someone might decide to delete the coverage script one day.
3) #634,  Change the test command in `package.json` to: 
```"test": "react-scripts test -- --coverage",```
then the .travis.yml only needs to perform `npm run test`

I can modify #631 to have these changes if we want all of the discussion info housed in that pr. 
I am adding all of the test info to a wiki page for the project.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ node_js:
 cache:
   directories:
     - node_modules
+install:
+  - npm install -g codecov
 script:
   - npm run build
-  - npm test
+  - yarn test -- --coverage && codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,4 @@ install:
   - npm install -g codecov
 script:
   - npm run build
-  - yarn test -- --coverage && codecov
+  - npm test -- --coverage && codecov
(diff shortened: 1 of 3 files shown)


knod:
Thanks! I'm still swamped, but if you're looking for something to do, we would probably appreciate having that info you provided before in the repo's wiki. No worries if that doesn't work out.

MichaelDimmitt:
Adding all of the test info to a wiki page for the project.

Additionally the .travis.yml could ```npm run coverage``` but I feel that is against convention because someone might decide to delete the coverage script one day.

MichaelDimmitt:
I will update the pr when #629 is merged.

knod:
Thanks. It's a shame things are going so slowly right at the moment. I hope that there's one last check I have to do before we can finally merge that change.

Issue #630, opened 2018-07-04
Correctly implement `ManagedNumberField` behavior (described)


knod:
This is what I've got so far for `<ManagedNumberField>` desired behavior.

**Valid number:** A positive decimal or whole number, 0, or a blank field.

- User clicks on field with '0' in it:
  - Field goes blank, ready for input
- User clicks on field with valid number in it:
  - Nothing changes
- User types a valid number in the field:
  - Typing shows up as entered
- User types a valid number in the field and then leaves the field:
  - Valid value is shown as decimal value
  - (Valid value is stored permanently)
- User types an INVALID number in the field:
  - Typing shows up as entered
  - Field gives 'invalid value' feedback (turns red)
- User types in an INVALID number and then leaves the field
  - The last valid field value previously entered is shown (last valid value typed, or last valid value left there?)
  - (Invalid value is discarded)
- User types 0 in a field:
  - Typing shows up as entered
- User deletes all items in the field
  - Input does not turn red
- User deletes all numbers in the field then leaves the field
  - Field shows '0'
  - (0 is stored permanently as the value)

Is this behavior possible?

knod:
It looks like that behavior isn't possible with the system we currently have. Both an empty field and invalid values like 'e' return an empty string. New suggestion:

Just completely prevent the user from typing in anything other than a number with, at most, one decimal place. This includes allowing 0. It's not what we originally wanted, but maybe it's something that can actually work.

knod:
What I found from experiments:

```
Click 0, click out (no feedback)
focus:      input value: 0.00,      value: 0
blur:       input value: ,          value: 0

Click 0, 2, click out (no feedback)
focus:      input value: 0.00,      value: 0
change:     input value: 2,         value: 0
blur:       input value: 2,         value: 2

Click 2, add .5, click out (no feedback)
focus:      input value: 2.00,      value: 2
change:     input value: 2.50,      value: 2
blur:       input value: 2.50,      value: 2.5

Click 2.5, add e, click out ('invalid' feedback)
focus:      input value: 2.50,      value: 2.5
change:     input value: ,          value: 2.5
blur:       input value: ,          value: 2.5

Click 2, delete all, click out ('invalid' feedback)
focus:      input value: 2.00,      value: 2
change:     input value: ,          value: 2
blur:       input value: ,          value: 2

Click 0, 0, click out (no feedback)
focus:      input value: 0.00,      value: 0
change:     input value: 0,         value: 0
[...]

PR #633, opened 2018-07-05
fix test suite, hoping for discussion.

MichaelDimmitt:
#625, has caused future Travis CI builds to fail because of failing test:
```bash 
 FAIL  src/test/forms/ManagedNumberField.test.js
  ● should change local and app state correctly when user inputs empty string

    TypeError: Cannot read property '1' of undefined
```
@snyderc, I wanted to see if the change in this PR to put back in the value = "0" was acceptable.
I left the blur change in the code, hopefully it still achieves your objective.

Either the blur has a job but this code needs back in,
the blur is not doing its job but could do the job.
or the test needs to be rewritten for the new changes.

If we cant get this resolved maybe #625 should be reverted and that pull request be submitted again with another code review.

Hopefully this will bring up positive discussion. Im having a hard time understanding, because the blur does bring in the values. But does not change the state in the test. Could this be because the blur only fires after a value has changed on the predictions page and a click away from the value column, maybe the test is before that event.

### DIFF
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -506,6 +506,9 @@ class ManagedNumberField extends Component {
     if (valid) {
       store(evnt, inputProps, otherData);
     }
+    else if (value.length === 0) {  // treat empty string as 0
+      store(evnt, { ...inputProps, value: '0' }, otherData);
+    }
     this.setState({ focusedVal: value, valid: valid });
   };  // End handleChange()
 


snyderc:
#629 is also related.

Thank you for flagging this! The reason this test is failing is that I wrote some tests based on handleChange (#622), and then on the same day and in a separate branch submitted #625 to address a bug. (I just added to #625 some explanatory text in a comment.) #625 deliberately moves certain functionality from handleChange to handleBlur, but the unit test from #622 expects the functionality to be in handleChange, so that's why you're seeing this test fail. (In retrospect, I should have waited for one PR to be approved before submitting the other PR.)

I recommend that the failing test be removed, as that will fix the problem being reported in this PR. Then, I'll need to write a new unit test, either based on #625 or #629 (the latter of which has some open discussion).

MichaelDimmitt:
@snyderc, is it possible to get these changes added in a new pr?
As the context of removing the test and writing a new test for these issues is one that I am unfamiliar with. Then I would be happy to close this pull request.

also @knod, I know you've been busy at the moment but do we know others that should be aware of these changes so that when synderc submits a new pr it is ready for acceptance/review.

knod:
I think we're all on the same page, but I'm going to put it out there - my general take on tests that are failing are to not remove the test in order to get rid of the red 'x', but to get solve the failing test by figuring out what's wrong and implementing our desired behavior. @ethanbb has created a PR dealing with the issue, though it needs to be tweaked to conform with the behavior described in #630.

After that, I think it'd be great to write some thorough tests for that behavior. This is a component that's given us plenty of trouble.

Does that answer the concerns or have I missed the mark? Do folks have more thoughts on the issue?

MichaelDimmitt:
@knod, cool. I submitted this pr for a quick fix alternative. In case #630 was rejected. Glad @ethanbb is knocking it out. Ill keep his branch up to date with my own and report in #630 when the tests start passing with the new changes.

PR #647, opened 2018-07-10
Fixes #646 - reduce form section bottom nav buttons to one instantiation.

### LINKED ISSUE #646: Abstract 'Previous' and 'Next' buttons into one location
Probably 'VisitPage.js' would be best.

knod:
Changes the style of the page a little, but that looks ok to me.


knod:
Hmm, I might actually not like that style more, but we'll have to think about how that can change in an effective way.

PR #649, opened 2018-07-11
Spelling out SNAP on the first page of the form

snyderc:
Provides the spelled-out version of SNAP (Supplemental Nutrition Assistance Program) in addition to the acronym. For clarity & for making it easier to translate into other languages.

This came up today while going through translations with a volunteer translator.

### DIFF
--- a/src/localization/en.js
+++ b/src/localization/en.js
@@ -132,7 +132,7 @@ export default {
         hint:  'Section 8 provides rental housing assistance.',
       },
       hasSnap: {
-        label: 'Do you have SNAP?',
+        label: 'Do you have SNAP (Supplemental Nutrition Assistance Program)?',
         hint:  'SNAP provides assistance with buying food',
       },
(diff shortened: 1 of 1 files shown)


snyderc:
Added a spelled-out version on the About page.

I would suggest that it's also important to include on the first form page, because spelling out acronyms upon first reference is a good practice, and the first form page will for many people be the first time they see this acronym while using the app. The "hint" section can still be used for explaining what SNAP is (the current hint snippet in the code is "SNAP provides assistance with buying food") but shouldn't be relied on to spell out the acronym.

knod:
Hmm, I have some thoughts on that. How about we keep the form the way it is for now and discuss it more before making that change?

snyderc:
Ok, I removed the spelled-out version of SNAP from the first form page. This PR now only adds spelled-out SNAP to the about page.

PR #651, opened 2018-07-11
Snippets to backticks instead of quotes

knod:
Can't use the backticks for prop names, so they'll have to be in quotes if we're going to stick with the versioning system ~we currently have~ [we're currently planning].


wacii:
These aren't the same. Newlines in template literals count as newlines.

```js
str1 = `a
b`;
str2 = "a\nb";
// str1 == str2
```

knod:
@wacii : I thought I got rid of all of them. Which lines have I missed?

ethanbb:
So the main purpose of this change is just to get rid of escaped quotes within the snippets, correct?

knod:
@ethanbb : Yes, and to get rid of strings put together with '+'. There are a bunch of changes that need to be made to snippets, and this is the first, and probably most simple, step.

src/localization/de.js
```diff
@@ -1,171 +1,135 @@
...
+    why2: `Die Cliff Effects sind schwer vorherzusehen. Die Interaktionen zwischen dem Einkommen, der Nummer von Personnen im Haushalt, vielen verschiedenen anderen Krieterien, und den Auswirkungen der Programmen selbst haben Auswirkungen aufeinander in unerwarteten Weisen. Wir suchen Methoden, um dieses Problem sich mit der Komplexität zu befassen, und die Familien ihre Situationen besser zu verstehen und vorherzusehen.`,
     
-    linkText1: 'Ein zweiminutenlanges Video über Cliff Effects',
-    linkText2: 'Manche Quantitative Szenarios, die Cliff Effects demonstrieren',
-    linkText3: 'Eine Aufgliederung von verschiedenen Sozialhilfeprogrammen im Bundesstaat ' +
-      'Massachusetts',
+    linkText1: `Ein zweiminutenlanges Video über Cliff Effects`,
+    linkText2: `Manche Quantitative Szenarios, die Cliff Effects demonstrieren`,
+    linkText3: `Eine Aufgliederung von verschiedenen Sozialhilfeprogrammen im Bundesstaat ` +
+      `Massachusetts`,
```
wacii: These lines weren't combined.

knod:
Yeah, that was part of the reasoning. The `+`s are very dependent on the text there and they make editing fragile.

zhujinxuan:
I think this can be automatically fixed by prettier.  I will raise that issue in next week.  (Sorry I fail to attend this Tuesday)

PR #658, opened 2018-07-13
Fixes interpolation not giving keys because of my last fix

knod:


wacii:
The callback for `map()` receives an index as the second argument. So you don't need to maintain the `count` variable yourself. At least not in the first instance.

I don't think you should provide random values for `key` just to make the warning go away.

knod:
@wacii : What do you suggest other than a random key? Is there another purpose for which we need the key? Because I do find the warning pretty annoying.

wacii:
I don't know enough about React internals to say exactly what could go wrong in this case. The components in these lists don't have state (other than maybe `<ExternalLink>`), so we don't have to worry about that. But if there were discrepancies between translations, then two elements you identified as the same may actually use different components. And I'm not sure if that could cause a problem.

A safer alternative would be to use something like a GUID. Just some big random identifier, big enough that the likelihood of collision is negligible. In this case `${lang}-${count}` would probably also work.

wacii:
A simpler approach would be to just move `count` outside the functions. So that they were each iterating the same number. Or you could just leave it as is. Lots of people get away with doing this. And its not like these lists are going to be reordered or anything. So maybe I'm just being picky.

knod:
Eh, like the idea of keeping things unique, though I can't at the moment see under what circumstances adding `lang` would change that. I'll have to look into it more, and/or maybe we can chat if I ever manage to be free the same time you are.

Not sure if this is relevant, but I think, from what I understood in the docs, that keys are unique to their array, so they don't interfere with each other. Maybe I dreamed that up, though.

knod:
My local tests are passing with no warnings, but Travis tests are failing. I can't track down where to see the npm log for Travis. Anyone have a clue?

Also, we have 39 'vulnerabilities' and my memory is failing me - is that new, or have those been there all along?

wacii:
Sorry for the late response, but I suggested adding `lang` because I didn't think the interpolated content would change for any reason other than the language being changed.

And yes, I believe the keys are effectively scoped to the parent element. Still, I thought that in this case, when the language changed, the array of elements here was the only thing that would change. So an item keyed as `1` might refer to an English phrase one frame, and a German one the next.

Regardless, there will certainly be no issue with the new keys.

Issue #648, opened 2018-07-11
Abstract out the form step container.


syeung14:
**example:** 

**Current way of using the form** 
```
   <Form
      size='massive'
      className='household-size-form flex-item flex-column'>
      <FormPartsContainer
        title     = { snippets.currentBenefits }
        clarifier = { snippets.selectBenefits }
        right     = {{ name: 'Next', func: nextStep }}>
        <CurrentBenefitsContent
          setClientProperty = { setTimeProp }
          current           = { client.current }
          snippets          = { snippets } />
      </FormPartsContainer>
    </Form>
```

**Goal**
Abstract out the form step container and make it as react.js common component

knod:
Fixed by #800

PR #660, opened 2018-07-18
#648 - abstract form component in formHelper.js

syeung14:


src/forms/formHelpers.js
```diff
@@ -157,6 +157,9 @@ const FormBottomRow = function({ left, middle, right }) {
 */
 const FormPartsContainer = function({ title, clarifier, children, navData }) {
   return (
+    <Form
+    size='massive'
+    className='household-size-form flex-item flex-column'>
```
knod: Shoot, just noticed - we need to pass a custom class name in for the form - look at that first class name. Is that something you feel up to adjusting?

knod:
Also, the snapshots may need to be updated after some visual confirmations.

src/forms/formHelpers.js
knod: We could possibly just use a changed version of the title to be the class name. Something like ``var formClass = title.toLowerCase().replace(` `, `-`);``

Issue #543, opened 2018-05-31
UI bug: checkmarks not consistent 


ohmegasquared:
Two things are happening that make the check mark metaphor break a bit. 

- The check mark appears in the tabs when you are on that page, even though the user hasn't complete that page yet. If the check mark means complete, it should happen after you complete a page. 

- If a user needs to go back, the check marks are no longer visible. If a page is completed, the check marks don't represent that.

knod:
I had caveats, but thinking on it more, I agree. The check marks are confusing. It may be a good idea to get rid of them completely. All the data in the form is always 'valid'. That is, there is no field that doesn't have a default value. It's done for code logistical reasons.

Either we need to change that or we need to not imply that there is such a thing as an incomplete page. It's a bit of a conundrum. For example, all money values are automatically zero.

Regarding non-completeness - if we required everyone to fill in money amounts for each item, that would be unreasonable. We could handle it a different way - ask a leading question like, 'Do you get money from TAFDC?', but I think it may add unnecessary steps to a process that can already become long.

Hmm, maybe an interaction where someone checked off all the forms of income they receive? I still think that may add unnecessary complication.

Thoughts, anyone?

knod:
Hmm, sure. How about we start by removing the checkmark. We'll make this issue about that. Then we'll think about doing the dot thing in another issue. I'll check on someone to take this one on.

knod:
Closed by #663

PR #663, opened 2018-07-22
Parker/issue #542 remove checkmarks

pporfilio:
Changed StepBar to no longer mark steps as completed, which means that check marks will not appear in the steps.

I couldn't find anywhere else that cared whether or not `completed` was true, but maybe someone more familiar with semantic-ui-react/this project could confirm before merging this.


knod:
Closes #543

Issue #670, opened 2018-07-25
Refactor formHelpers.js


knod:
Long time coming. 'formHelpers.js' is way too long. I'm going to start working on how to break out this stuff into categories that make sense. My plan is still to have multiple components in a file, just to have files that are more specific to function.

knod:
I'm going to take a break from this while I'm shaking the tree elsewhere. Let's see if someone else wants a crack at it for now.

ethanbb:
I think only one more change is needed - separating `SpaceHolder`, `FormBottomRow`, and `FormPartsContainer` into `FormPartsContainer.js`, and probably leaving the last 3 together (which are `ExternalLink`, `AttentionArrow`, and `InvalidMessage`.  I can do this. I'm trying to think of a good name for the remaining components - `textHelpers`? `textElements`? Something else?

ethanbb:
Actually, @knod suggested that this might be a good thing for a beginner to try instead.

knod:
This could probably be finished off with a little bit more work.

knod:
@skudbucket : If all we do is find a more useful name than 'formHelpers', that seems good to me. The rationale is that the name isn't descriptive (in my opinion). I just haven't thought of a good name that covers those functionalities. I'm open to other perspectives, though, so let me know your thoughts.

P.S. `ExternalLink` is being moved out of there already, btw, since non-form stuff uses it. #707 just needs a bit of adjustment.

PR #672, opened 2018-07-25
Interactive UI elements moved to own file - #670

knod:


knod:
You're right about the CSS! I totally forgot!

`MassiveToggle` isn't being used and in another branch (that I reserved for cleaning) I just deleted it. I've tried to think of places we would use it, but whatever situation I thought of, some other way was always more clear.

knod:
There - cleaning PR made.

knod:
Why did the snapshot not catch the padding change?

PR #679, opened 2018-07-28
 Fix language codes not actually specific to snippets

### LINKED ISSUE #678: Assign langCodes correctly when combining snippets objects
If we combine two different language snippets objects (e.g. to fill in gaps in a translation) and then do interpolation, the wrong language code will be propagated to the snippets in one of the objects (based on which one provided the top-level `langCode` value). To fix this, we should do the interpolation process on snippets before combining them.

ethanbb:
Fixes #678. Interpolates snippet objects _before_ merging them, so that the spans created from the snippets in each object get the correct `lang` code during the interpolation process. You can see the change by setting the default language in `App.js` to `de` instead of `en`, and comparing snippets that are already in German with those that are currently defaulted to English.

I also noticed that the `<html>` tag had an invalid `lang` attribute (`[object Object]`), since the `langCode` was also getting wrapped in a `<span>`. I changed the interpolation function to not wrap `langCode`, but just to be doubly safe, I also changed the `<html>` attribute to use the language code from the state directly rather than reading from the snippets.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -21,7 +21,7 @@ import { getTextForLanguage } from './utils/getTextForLanguage';
 class App extends Component {
   constructor (props) {
     super(props);
-    this.state = { language: 'de', snippets: getTextForLanguage('de') };
+    this.state = { language: 'en', snippets: getTextForLanguage('en') };
   }
 
   setLanguage = (inputProps) => {
--- a/src/App.js
+++ b/src/App.js
@@ -21,7 +21,7 @@ import { getTextForLanguage } from './utils/getTextForLanguage';
 class App extends Component {
   constructor (props) {
     super(props);
-    this.state = { language: 'en', snippets: getTextForLanguage('en') };
+    this.state = { language: 'de', snippets: getTextForLanguage('de') };
   }
 
   setLanguage = (inputProps) => {
@@ -30,7 +30,7 @@ class App extends Component {
   };
 
   render () {
-    var { snippets } = this.state;
+    var { language, snippets } = this.state;
 
     // Confirms user navigation
     var confirmer = new Confirmer();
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -30,15 +30,15 @@ class App extends Component {
   };
 
   render () {
-    var { snippets } = this.state;
+    var { language, snippets } = this.state;
```
knod: Note for future: We should probably change this to `langCode` too.

src/utils/getTextForLanguage.js
```diff
@@ -24,23 +26,20 @@ const mergeCustomizer = function (objValue, srcValue) {
  * doesn't exist, it warns the coder and returns English.
  */
 const getTextForLanguage = function (langName) {
+  if (langName === 'en') {
```
knod: Also note for future: This probably should also be `langCode`.

src/utils/getTextForLanguage.js
```diff
@@ -24,23 +26,20 @@ const mergeCustomizer = function (objValue, srcValue) {
...
     // so that we fall back to English if a particular field is missing.
-    snippetsTemplate = mergeWith({}, english, localizations[ langName ], mergeCustomizer);
+    return mergeWith({}, interpolatedEnglish, interpolatedSnippets, mergeCustomizer);
 
   } else {
-
     console.warn('There\'s no localization for ' + langName + '. Defaulting to English.');
-    snippetsTemplate = english;
-
+    return interpolatedEnglish;
```
knod: It's a bit confusing that we're returning something called 'interpolatedEnglish'. That's a wording issue, though, and we just tweak it in a different PR.

knod: Oh, my bad - this only happens if there isn't a matching language. Nvm.

knod:
Adding label to remind myself of possible todo items.

PR #681, opened 2018-07-30
Cache interpolated/merged snippets

ethanbb:
On top of the method for fixing #678 in #679. In that PR, the interpolated English snippets are calculated beforehand and kept as part of the `getTextForLanguage` function (in its closure). This goes further by keeping an object in that function's closure which contains the English snippets as well as all other interpolated and merged snippets that have been calculated so far - basically it caches its output. So `getTextForLanguage` doesn't have to call `interpolateSnippets` or `mergeWith` again if it is called a second time for a given language.

(I played around with inspecting `finishedSnippets` using a breakpoint in Chrome dev tools while changing the language, and it seems to be working correctly (that being said we should write some tests for this file!))

### DIFF
--- a/src/utils/getTextForLanguage.js
+++ b/src/utils/getTextForLanguage.js
@@ -8,8 +8,8 @@ import { interpolateSnippets } from './interpolation.js';
 import { localizations } from '../localization/all';
 import inlineComponents from '../localization/inlineComponents';
 
-// interpolate components into English snippets:
-const interpolatedEnglish = interpolateSnippets(localizations.en, inlineComponents);
+// store interpolated and (if necessary) merged snippets objects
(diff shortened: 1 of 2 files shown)


src/utils/getTextForLanguage.js
```diff
@@ -8,8 +8,8 @@ import { interpolateSnippets } from './interpolation.js';
 import { localizations } from '../localization/all';
 import inlineComponents from '../localization/inlineComponents';
 
-// interpolate components into English snippets:
-const interpolatedEnglish = interpolateSnippets(localizations.en, inlineComponents);
+// store interpolated and (if necessary) merged snippets objects
```
knod: Note for further discussion: Would it be worth making this a jsdoc documented variable?

PR #692, opened 2018-08-02
Adds and implements expandable details - fixes #691

### LINKED ISSUE #691: Expandable explanations/details
Instead of having the explanations of a question or heading written out, have a question mark and details that are initially hidden, but can be expanded.

knod:
It does use a less common color into the app. Should we make that 'x' orange instead? Warnings about invalid stuff is red right now, too, though.

### DIFF
--- a/src/components/details.js
+++ b/src/components/details.js
@@ -1,0 +1,60 @@
+import React, { Component } from 'react';
+import {
+  Icon,
+  Transition,
+  Message,
+} from 'semantic-ui-react';
+
+
+class HeadingWithDetail extends Component {
+  state = { showDetails: false };
+
+  toggleDetails = () => {
+    if (this.state.showDetails) {
+      this.setState({ showDetails: false });
+    } else {
+      this.setState({ showDetails: true });
+    }
+  };
+
+  render () {
+    var { showDetails } = this.state;
+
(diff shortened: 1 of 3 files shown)


knod:
Good idea! Totally spaced on that. @Butterwell , @Irtaj : I'll request reviews from both of you. Just click on 'Show all checks' next to 'All checks have passed' below and then 'Details' next to 'deploy/netlify'.

src/components/details.js
```diff
@@ -0,0 +1,60 @@
...
+
+class HeadingWithDetail extends Component {
+  state = { showDetails: false };
+
+  toggleDetails = () => {
+    if (this.state.showDetails) {
+      this.setState({ showDetails: false });
+    } else {
+      this.setState({ showDetails: true });
+    }
```
wacii: This isn't always safe. I would suggest using the `setState(updater)` pattern to guarantee you are working with the proper version of `state`.

```js
this.setState(prevState => ({ showDetails: !prevState.showDetails }));
```

Or something similar that conforms with our style guide.

wacii:
I would also probably add assertions on the number of children passed to the component. Requiring one or two child child elements is unusual enough to warrant it I think.

Issue #397, opened 2018-01-29
Horizontal line on graph showing total client expenses?


knod:
We once considered having a 'poverty line' horizontally across the chart, but that never seemed right. A friend suggested, instead, a horizontal line to show the total client expenses. Then they can see when they will dip below that line. It sounds like it would be really useful!

There would have to be in input for the client where they show total expenses, though. I think it would have to include both out of pocket and things that other programs pay for, so we may have to think about how to do that. Maybe Kristin could give some input on that.

knod:
Kristin is in favor of the feature and a few interested parties on the team have felt similarly. We've agreed on an option for the user to select to be able to see their expenses relative to their benefits. When the user selects that option (in the 'Expenses' section), new questions appear offering inputs for further expenses that will then be used to add to the visualizations. @Butterwell is looking into whether chart.js has these capabilities or, alternatively, how difficult it would be to switch to raw d3.js.

knod:
I'm working on setting up the extra questions.

knod:
To see expenses questions Project Hope uses, see the last page of their assessment worksheet: https://drive.google.com/drive/folders/0B61uzoaPZg7qUXJuRFhndTFkaTA.

PR #695, opened 2018-08-04
Offers questions about other expenses to the user for graph visuals - #397

knod:
Take a look at the preview in the 'Expenses' section. Also, put in some rent or mortgage and see how that shows up in the questions. It's a bit long, but it's based off of the assessment worksheet Project Hope uses. See the last page at https://drive.google.com/drive/folders/0B61uzoaPZg7qUXJuRFhndTFkaTA.

This section of questions is a bit long, but I'm not sure how we could usefully decrease its length.

This doesn't implement any of the visuals or anything, which is why I'm pulling to a feature branch I made ('expenses') off of 'dev'.

It may be worth implementing `<ShowOnYes>` regardless of how this feature. I have a feeling, though, that there's some kind of better abstraction. A `<ShowWhen>` that's more general? With a question component and a content component? The question could trigger showing the content...


knod:
Fixes that and snapshots.

knod:
Also, `<RenderIfTrue>` is in there too, but it's not finalized and it's not in an appropriate file. How that component's going to be implemented hasn't been worked out yet, but I wanted to at least field test it. You'll also notice that there are ternary operators in there too for simple string choices. It's all a grand experiment!

src/forms/ShowOnYes.js
```diff
@@ -0,0 +1,82 @@
...
+ * @param {object} props.children - Components to be revealed.
+ * 
+ * @returns {object} Component
+ */
+class ShowOnYes extends React.Component {
+  constructor (props) {
+    super(props);
+    
+    const show = props.clientPartial[ props.propName ];
+    this.state = { showChildren: show };
```
wacii: There shouldn't be a need to use state here. Just use `client[propName]` directly.

src/forms/cashflow.js
```diff
@@ -103,7 +103,37 @@ const CashFlowRow = function ({ generic, timeState, updateClientValue, children
...
+      <div className = { colClassName + ` ` + generic + ` output-number` } >
+        { weekly }
+      </div>
+      <div className = { colClassName + ` ` + generic + ` output-number` } >
+        { monthly }
+      </div>
+      <div className = { colClassName + ` ` + generic + ` output-number` } >
+        { yearly }
+      </div>
+      <div className = { colClassName + ` cashflow-column-last-child` }>
```
wacii: It's a little weird to be using template strings here but not interpolation. Not something that needs to be fixed though.

knod:
Oops, never merged this. Will go ahead and do that, since this is on a separate branch from `dev`.

PR #700, opened 2018-08-11
Adds text snippets for Predictions.js to en.js

ValerieKenyon:


ValerieKenyon:
I've added the four tab titles.

knod:
I'm not great at spotting missing/incorrect items, but I trust your judgement!

PR #703, opened 2018-08-14
Switches to current convention for export

knod:
as well as allowing testing of other form container components


knod:
Did we figure out why this was failing?

knod:
Right now it's looking like a timeout error. Re-running tests.

PR #706, opened 2018-08-15
Generate jsdoc documentation and make available at /docs in published build

ethanbb:
This publishes the JSDoc comments at https://codeforboston.github.io/cliff-effects/docs (after running `npm run deploy`). Uses the package documentation.js (so requires an `npm install`), and its HTML output is served via an iframe and React Router. `npm run docs` generates the documentation, `npm run build` now automatically copies the docs to the build folder after building, and `npm run deploy` will perform both of these steps before deploying.

The next step will be to seriously clean up these JSDoc comments, so that the documentation is actually readable and sensical. There's a lot of cleanup that can be done to make the docs more useful.

(Thanks @knod for suggesting the iframe method, resolving our considerable issues with service routers or trying to transform the HTML into a React-friendly format!)


wacii:
I wouldn't have thought of using an iframe. Lots of unexpected uses for those things.

Not sure how to test as deploy is linked with a specific url, right? Still, I took a look at the code as well as the generated documentation and everything looks good.

knod:
@wacii : You should be able to test by running publish for your own repo.

Quick question: If we had a link on the main page to the docs, would we be able to then see the docs in the netlify preview? Is there any way to make that happen? It seems really weird, and somewhat awkward, that there'd be no way to access the docs while developing.

That said, it's a fantastic start and I don't think that should be a barrier to accepting this PR. I just have to figure out how to publish to my repo since I gave it a strange name and am now stuck with it.

knod:
What if we didn't ignore '/docs'?

ethanbb:
I believe you can see the docs in the netlify preview - I think I did it before. Did you go to `/docs` (no hash)?

knod:
Notes for next steps:
1. Is there a way to generate the name and line number of the file for each description?
2. Can we add `npm run docs` to the watchify command, or does it rebuild from scratch each time making it take forever to load the page?
3. Is there a way to organize the descriptions? According to folder and file, or something like that?

PR #710, opened 2018-08-18
Use Switch to enforce one page component

### LINKED ISSUE #682: Add a `<Switch>` to non-repeating `<Route>`s in App.js
Some `<Route>`s need to be on every page (like the one for the header). Those ones shouldn't be in the `<Switch>`. Others cancel each other out - if we're on the About page, none of the form stuff matters.

We need to put a `<Switch>` in the appropriate place and maybe switch up which order the routes appear in so that we can always get to the right place (one 'load' may need to be in a different order).

wacii:
Any Route components that match are rendered, so only accurate paths kept multiple pages from being rendered at the same time. While there are no problems now, using the Switch component guarantees only one page component is rendered.

Forgot to push this. Might have to revisit when the docs change is merged. There should also probably be a fallback page. If not a 404 Not Found sort of page, a redirect to the home page.

Closes #682

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import {
   HashRouter,
   Route,
+  Switch
 } from 'react-router-dom';
 import { Helmet } from 'react-helmet';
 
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -51,50 +52,52 @@ class App extends Component {
...
+                path="/visit/:clientId/:visitId"
+                component={ (props) => {
+                  return (
+                    <VisitPage
+                      { ...props }
+                      confirmer = { confirmer }
+                      snippets  = {{ ...snippets.visitPage, langCode: snippets.langCode }} />);
+                } } />
+              <Route
+                path="/visit/load"
```
knod: Can this ever be reached if it's not put above the previous route?

knod: As long as `"/visit/load"` doesn't match `"/visit/:a/:b"`, then that's fine! You were right about the conflict with Ethan's PR, but once that's worked out, this looks fine!

knod: Come to think of it, I haven't really tested it thoroughly, but I'll do more of that once the conflicts are resolved.

knod:
I've resolved the conflicts, but it will probably end up failing the snapshot tests. We could:
1. Merge it and fix the snapshots later or
2. I could pull it and create a branch on my own repo where the tests work and then make a new PR.

Thoughts?

Issue #642, opened 2018-07-10
Restore disclaimer above 'Predictions' tables


knod:
~Find in old commits and bring it back.~

Use the disclaimer in the 'Predictions' section in commit ~09506a47d6518767c4faa2f1e4efcebd276a3251~ 9718ebd3aadd3b58ac9b66ebc6950ebd3e417757.

shp54:
Add this message and import the right components.
```<Message visible warning style={{ 'textAlign': 'center' }}>
This tool is in testing and these numbers might not be right. If they're not, we'd appreciate your feedback.<br />
      <Button
        fluid
        color='teal'
        style={{ 'display': 'block',
                 'marginLeft': 'auto',
                 'marginRight': 'auto',
                 'marginTop': '10px',
                 'maxWidth': '400px' }}
        onClick={feedbackPrompt}>Submit Feedback</Button>
</Message>```

knod:
Actually, I think we could stick with what it was in commit 09506a47d6518767c4faa2f1e4efcebd276a3251.

knod:
My bad, @shp54 was right. See 9718ebd3aadd3b58ac9b66ebc6950ebd3e417757.

knod:
Closed by #713

PR #713, opened 2018-08-22
restore disclaimer above predictions tables, restore DownloadFile.js …

skudbucket:
…(used in disclaimer) to src/components

This pull request is intended to address issue #642.

Trying to be very explicit about everything I'm doing here since with my first task on the project I feel it's mostly about getting the process right. As mentioned in the commit comment, here's what I did:

In Predictions.js, I added JSX to render the disclaimer `<Message>` between the `<Divider>` and the `<Header>`, as it was in commit 09506a4 per knod's comment in the issue. I also added Message and DownloadFile to the import block in Predictions.js.

In src/components I restored the DownloadFile.js component as it was in commit 09506a4 as well, since it seemed to have been since removed.


src/forms/Predictions.js
```diff
@@ -115,6 +115,7 @@ const PredictionsStep = function ({ updateClientValue, navData, client, snippets
           future       = { client.future }
           time         = { 'future' } />
         <Divider className='ui section divider hidden' />
+        <Message visible warning>This tool is in testing and these numbers might not be right. If they're not, we'd appreciate your feedback. Please <DownloadFile>download the anonymized data</DownloadFile> and email it to <a href="mailto:andrew@codeforboston.org">andrew@codeforboston.org</a> with the answers to these questions: What are the correct numbers? What questions did the form miss asking that could be affecting the situation?</Message>
```
knod: The previous `<Message>` was actually between the header and the table. It seems to me more likely to be read that way, and to be associated with the results, but that's just my gut feeling. I'm definitely open to discussion. For now, I'll mark this as requesting changes until someone brings up a different perspective.

src/components/DownloadFile.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Download file link.
+ * 
+ * @param props {object}
+ * @param props.children {string|null} - Optional link text
+ * @param props.data {object} - Data to be saved
+ * @param props.filename {string}
+ */
+class DownloadFile extends React.Component {
```
knod: My bad - I intended to have the disclaimer as it was later on, where it contained a trigger to the feedback form. I must have been pretty tired last night. I'll find the correct commit and link it - it just uses the same stuff that the other 'Feedback' button uses.

src/forms/Predictions.js
knod: I've adjusted the instructions in the issue. Again, sorry about that.

skudbucket:
I updated this so that the disclaimer is below the header (seemed like solid logic from @knod about why this might be better), and switched it to using the newer version from [9718ebd](https://github.com/codeforboston/cliff-effects/commit/9718ebd3aadd3b58ac9b66ebc6950ebd3e417757), and I removed the src/components/DownloadLinks.js file, which isn't needed with the newer version of the message, and I updated the import block in Predictions.js.

In a minor incidental change, I also changed the links in the README.md to the "github workflow" to a working URL (someone may want to confirm that they're going to what they should be going to) because the old URL wasn't working.

skudbucket:
I've also now updated the snapshot of Predictions.js so that it should pass the automated testing now

README.md
```diff
@@ -42,7 +42,7 @@ Instructions:
 7. Press cltr+c to stop the local server that you have running.
 
 <!-- Further discussion needed. See outdated comments for merge #154 (near the bottom of them). -->
-To get yourself set up to keep up to date with the latest changes to the `dev` branch, make sure you've gone through [Step 3 of the github guide](https://help.github.com/articles/fork-a-repo/#step-3-configure-git-to-sync-your-fork-with-the-original-spoon-knife-repository). To keep your branches from getting tangled as you both make changes and keep up to date with our repo, take a look at our wiki page about a possible [git workflow](https://github.com/codeforboston/cliff-effects/wiki/Git-Workflow).
+To get yourself set up to keep up to date with the latest changes to the `dev` branch, make sure you've gone through [Step 3 of the github guide](https://help.github.com/articles/fork-a-repo/#step-3-configure-git-to-sync-your-fork-with-the-original-spoon-knife-repository). To keep your branches from getting tangled as you both make changes and keep up to date with our repo, take a look at our wiki page about a possible [git workflow](https://github.com/codeforboston/cliff-effects/wiki/Guide:-Git-Workflow).
```
knod: Thanks for correcting this! In future, though, try to keep each commit focused on one problem.

knod:
This is taking too long! We'll merge it and ask for opinions after. It can always be changed!

PR #714, opened 2018-08-23
Change component of Route to a function to elminiate warning

ethanbb:
There was a warning for the `/docs` route  saying that its `component` prop should be a function rather than an object. This makes it into a function.


ethanbb:
@knod What linting messages are you seeing? I'm not getting any...

ethanbb:
Never mind, there were some fixes that happened when I ran the command that I didn't see.

PR #715, opened 2018-08-25
Use snippets in predictions page - Issue #513

pporfilio:


src/localization/en.js
```diff
@@ -262,6 +262,8 @@ export default {
       tabTitleStackedIncomes:  `Stacked Incomes`,
       tabTitleBenefitPrograms: `Benefit Programs`,
       chartsHeader:            `With the new pay, how could your benefits change?`,
+      warningMessage:          `This tool is in testing and these numbers might not be right. If they're not, we'd appreciate your feedback.`,
```
knod: Since these are new ones, it may be worth just starting them out with version numbers instead of going through this intermediate step. In fact, maybe it's worth versioning all the snippets that you're touching here. Would you be up for that?

pporfilio: 👍 Yep I think it makes sense to change all the snippet names I'm touching since this is the direction we're going. Just pushed an update to that effect.

src/localization/en.js
```diff
@@ -255,13 +255,15 @@ export default {
...
-      tabTitleChangesChart:    `Changes Chart`,
-      tabTitleStackedIncomes:  `Stacked Incomes`,
-      tabTitleBenefitPrograms: `Benefit Programs`,
-      chartsHeader:            `With the new pay, how could your benefits change?`,
+      title_v1:                   `What Might Happen?`,
+      futureIncomeQuestion_v1:    `How much money would you get paid in the future? (You can try different amounts)`,
+      tabTitleChanges_v1:         `Changes`,
+      tabTitleChangesChart_v1:    `Changes Chart`,
+      tabTitleStackedIncomes_v1:  `Stacked Incomes`,
+      tabTitleBenefitPrograms_v1: `Benefit Programs`,
```
ethanbb: For the tab titles, maybe it would be better to name the snippets after the component each tab displays, rather than the English text (i.e. `BenefitsTable`, `BarGraph`, `AreaGraph`, and `LineGraph`). That way they wouldn't become outdated if we decided to change the text in the tabs (which is likely to happen more often than changing the component names).

knod: That makes sense, though I don't think @pporfilio was the one who named these. That is, it'd be nice, but I'm not sure it needs to be in the scope of this PR (for the purposes of moving things forward).

Looking forward, though, I'm not sure those generic names are great, though. Is there a way we can name them for their purpose as opposed to by the type of graph they are?

ethanbb: Yeah actually it's probably fine the way it is. Unless you want to come up with more descriptive names for the purpose of each tab. If we change the tab text we'll need new translations anyway though, so I guess we may as well just change the names of the snippets too if that happens.

knod: We can probably shorten them from removing 'Title'. And maybe 'BenefitPrograms' could change to 'BenefitProgramLines/Trajectory/Whatever'. Anyway, it doesn't have to happen here.

knod: Tangential: Maybe we should start commenting in github with @todo when we have an idea we want to come back to later, but don't want to make into an issue, so that it'd be easier to find those later...?

PR #719, opened 2018-08-26
Implements dev HUD

knod:
To see it in action, go to '/dev' and click 'on'. Only has a couple features. Will probably best help with localization/translation spotting right now. Other suggestions welcome!

### DIFF
--- a/src/components/CustomClient.js
+++ b/src/components/CustomClient.js
@@ -1,5 +1,6 @@
 import React from 'react';
-import { Divider, Form, Message } from 'semantic-ui-react';
+import { Form, Message } from 'semantic-ui-react';
+
 
 /**
  * Load previous session from object.
@@ -23,9 +24,7 @@ class CustomClient extends React.Component {
     json:   '',
   };
 
-  submit = (event) => {
-    const { client } = this.state;
-    event.preventDefault();
+  loadClient = (client) => {
     if (client === null) {
       return;
     }
@@ -35,7 +34,18 @@ class CustomClient extends React.Component {
       error:  null,
       json:   '',
     });
+
     this.props.loadClient({ client: client });
+  };
+
+  reset = () => {
+    this.loadClient(this.props.toRestore);
+  };
+
+  submit = (event) => {
+    const { client } = this.state;
+    event.preventDefault();
+    this.loadClient(client);
   };
 
   handleChange = (_event, inputProps) => {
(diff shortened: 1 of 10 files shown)


src/components/CustomClient.js
```diff
@@ -1,66 +1,74 @@
...
 
 /**
- * Load previous session from object.
- * @callback loadClient
- * @param {object} clientContainer - Object containing previous session as `client`.
+ * Propagate data from JSON object.
+ * @callback load
+ * @param {object} toLoad - Parsed JSON object to use.
+ * @todo Change to just 'load'? Change this component
+ *     to a more generic `<Load>`.
```
wacii: I don't think the callback documentation is local to this class, so just `load` may not be clear enough (no matter where you define the callback it can be referenced by any other jsdoc and gets its own entry in the toc). I do agree about changing the name `<CustomClient>`. That one never made sense to me.

src/components/dev/DevHud.js
```diff
@@ -0,0 +1,157 @@
...
+          onChange = { funcs.english } />
+      </Menu.Item>
+      <Menu.Item>
+        <HeadingWithDetail>
+          <Checkbox
+            label    = { `Mark non-English snippets` }
+            checked  = { devProps.nonEnglish }
+            onChange = { funcs.nonEnglish } />
+          <span>
+            Note: text that doesn't have an underline (for reasons) has no snippets.
```
wacii: A single quote is not an apostrophe, even if they are sometimes rendered the same. Not a big deal being that this is a developer tool, but elsewhere you should probably use the html code for an apostrophe `&#39;` or `&apos;`.

knod: Interesting! Is there a apostrophe on the keyboard somewhere? Without triggering special characters?

src/components/CustomClient.js
knod: Hmm, so we'll have to think about that. It may not be a change for this PR to make.

How about something about the input type? `<AcceptJSON>`, `<LoadJSON>`, `<GetJSON>`. Basically, it has an input to get a user string, it only accepts JSON, and then it sends the data to the parent. `GetUserJSON`? `JSONInput`? `JsonImport`?

ethanbb: Just "JSON" seems too generic to me. How about `<ClientJSONImporter>`? Or `<ClientImporter>`?

src/components/dev/DevHud.js
ethanbb: My gut feeling is it's more robust to keep the source files ASCII and use codes for special characters. At the very least it makes it harder for someone to accidentally use the wrong character / easier to notice errors. I think you can probably copy an apostrophe from the character map, though.

src/components/CustomClient.js
knod: Well, I'm thinking of making it more generic - the way it's built, it could import any data, not just client stuff.

src/components/dev/DevHud.js
wacii: @knod There is no apostrophe on the keyboard. Otherwise I agree with what Ethan said.

src/components/CustomClient.js
ethanbb: OK. `<JSONImporter>`? I think it should be a noun, and "Input" seems wrong because that makes me think of an input on a form, while this is a higher-level component.

knod: True, and that might work! I'll merge this for now anyway and we'll work on that later. @todo!

PR #720, opened 2018-08-26
Bring back experimental features, I hope

knod:
This is not through github this time, so I hope the changes don't bleed over into dev somehow...


ethanbb:
Hm, why didn't netlify run on this?

In any case, I have no qualms with merging into a non-dev branch.

PR #721, opened 2018-08-26
Add jsx indentation eslint rules

ethanbb:
I noticed that in some places, there were tabs instead of spaces and the linter wasn't warning about it, so I looked into it. I added the `react/jsx-indent` and `react/jsx-indent-props` eslint rules to catch these instances (and more generally enforce indentation within components).

 This also caused some issues with the ternary operators in CurrentExpenses.js, so to resolve this I put parentheses around the components within the operators where necessary. I think this is more consistent (keeping with the convention of starting whatever comes after the `?` or `:` on the same line), but we could discuss whether this is the style we want.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -42,6 +42,8 @@
     "react/jsx-closing-tag-location": ["warn"],
     "react/jsx-curly-spacing": ["warn", "always", { "spacing": { "objectLiterals": "never" }}],
     "react/jsx-first-prop-new-line": ["warn", "multiline-multiprop"],
+    "react/jsx-indent": ["warn", 2],
+    "react/jsx-indent-props": ["warn", 2],
     "react/jsx-max-props-per-line": ["warn", { "maximum": 1 }],
     "react/jsx-tag-spacing": ["warn"],
     "react/self-closing-comp": "warn",
--- a/src/App.js
+++ b/src/App.js
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import {
   HashRouter,
   Route,
-  Switch
+  Switch,
 } from 'react-router-dom';
 import { Helmet } from 'react-helmet';
 
--- a/src/components/prompts/FeedbackForm.js
+++ b/src/components/prompts/FeedbackForm.js
@@ -21,8 +21,13 @@ class FeedbackPrompt extends React.Component {
   handleInputChange = (event) => {
     // Source: https://reactjs.org/docs/forms.html#controlled-components
     const target = event.target;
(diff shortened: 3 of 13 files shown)


ethanbb:
From what I can tell, the general indent rule ignores anything that's within or a child of a JSX tag, except JS code within braces. This causes it to ignore indentation of both properties and child components.

The issue with ternary operators occurs when they appear within a JS block, so `indent` applies, and an opening tag starts on a new line, so `react/jsx-indent` applies. Without the parens, the general rule sees it as a continuation of a multi-line statement and wants to indent the tag to the same level as the '?' on the previous line, whereas the react rule just sees an opening tag on a new line and wants it indented relative to the previous line, so they conflict. With the parens, both agree it should be indented.

Another solution could be to put the '?' on the same line as the condition, get rid of the parens, and only indent the component once.

knod:
We could also put the '?' on the same line, along with a paren and just indent the component once...?

ethanbb:
@knod yeah, we could do that if we wanted to. That's not my preference but it's an option.

knod:
Eh, I'll still hate ternaries whatever we do, so I'm fine compromising on this one. I'll just go work on a `renderIfTrue()` >.>

knod:
Do you want to go with what we have here, or go with something else that's been suggested?

ethanbb:
OK, before I merge I'd like to make that change in `inlineComponents` that we discussed last night. Also maybe go through the other ternaries and make sure they're all consistent.

ethanbb:
Does anyone else have an opinion on how we should format multi-line ternary operators?

knod:
Heh, if I'm going to approve of a ternary, I think I'd be into that last one. You've got assignments going on, btw, instead of logic operators (`y = cond`). Also, maybe don't fix the `DevHud` ternary till we've decided on the PR I just made. I should have probably tried that out way sooner, though I do think this conversation is an interesting and important one. In fact, I'm liking that last ternary more and more as I look at it...

ethanbb:
OK cool. fwiw I meant that to be an assignment; `cond` is the condition. I will leave `DevHud` alone!

ethanbb:
Here's an update. As we discussed, I used the last style in my comment with mandatory parentheses, and I got rid of any ternary operators outside of JSX.

knod:
Wow, that was a fast turn around! Looks good! I do not rescind my approval :)

ethanbb:
@wacii Any other thoughts before I merge this?

wacii:
Nope, I like the last style as well.

PR #722, opened 2018-08-27
Turns subtle multi-line ternaries into more explicit function call

knod:
There! Thoughts?

I know it'll conflict with changes @ethanbb is planning to make, but I can deal with those.


knod:
Example of the kinds of things that could go wrong with `<RenderIfTrue>`:
https://codepen.io/knod/pen/bxemOj?editors=0011

src/components/renderIfTrue.js
```diff
@@ -0,0 +1,19 @@
+/* If `bool` is `true`, return the given component,
+ *     otherwise returns `null`.
+ *
+ * @param {boolean} bool If `true`, do what is desired.
+ * @param {object} ToRender Component to return if
+ *     condition is met.
+ */
+const renderIfTrue = function (bool, ToRender) {
```
ethanbb: I believe`ToRender` should be camelcase `toRender` since it's a specific (constructed) component rather than a component class.

src/App.js
```diff
@@ -36,7 +38,7 @@ class App extends Component {
     this.state = {
       langCode: `en`,
       snippets: getTextForLanguage(`en`),
-      clients: {
+      clients:  {
```
ethanbb: Tiny linting error here!

src/components/renderIfTrue.js
knod: When we're normally using components in JSX I guess we're using classes to instantiate instances, but I'd never thought of it that way. Interesting!

src/App.js
knod: Dangit! I ran the lint command and got no changes!

src/components/renderIfTrue.js
knod: Wait a sec. Is this component actually still is a class? Has it already been instantiated? If so, wouldn't it have already been rendered and its functions run? I'm thinking of `Route`, where you can give a `component` attribute which is just the name:
``` js
<Route
  path      = { `whatever` }
  component = { ToDo }
/>
```

src/App.js
knod: Oh, no, this is correct. It lines up with the previous key/value pair.

src/components/renderIfTrue.js
wacii: Yes, this has the same problem as described in your `<RenderIfTrue>` example.

React Router does demonstrate a couple solutions. The first, as you have pointed out, is to just pass a component class and have the receiving code be responsible for instantiating it.

The second is to provide a render function: `() => <RenderIfTrue bool={false} />`. This really isn't all the different from the previous. The receiving code is still responsible for "instantiating" the component instance, it just does it by calling a function rather than instantiating directly from a class.

src/App.js
ethanbb: Oh your're right, sorry! I saw I had a lint error there currently and for some reason assumed this was causing the problem instead of fixing it (doesn't really make sense now that I think about it...)

src/components/renderIfTrue.js
knod: So what's the verdict on the name? Caps or camel case? Since I think it may still be a class at that point.

ethanbb: To elaborate: the capitalized-component-class thing comes from how React interprets the element type in a JSX expression, as described here: https://reactjs.org/docs/jsx-in-depth.html#specifying-the-react-element-type. For example, `<Header>` is interpreted as a React element called `Header` (which must be in scope) whereas `<header>` is interpreted as an HTML element.

Since whatever's being passed into `renderIfTrue` is React's representation of a component itself, not an element type that would be used to create a component, it doesn't need to be capitalized. That being said this is a minor issue. But I don't think it should be capitalized.

knod: Thanks for the rundown! Reading that, though, I think I'm finally able to articulate why I've so attached to the capitalization - I think it makes it more clear that that argument is meant to be a `Component` and not something else. Thoughts?

ethanbb: OK, I don't agree but we could argue for days over this minor point. I say it's fine, let's go with how it is now!

knod: Bah, this is not the way I like to leave things! I'll just say that I can go ahead and merge it, but I think it bears further discussion. I'm still not completely clear on the relationships between these things, but I trust your grasp of it, in which case the way I've written it isn't technically correct. On the other hand, it's still easier to see at a glance that this variable is something that React can render. On the other hand, it's confusing to anyone who understands React well enough to know the difference. Bah! @todo

PR #728, opened 2018-09-04
Surrounder - the layout super-component

knod:
@todo Still need that discussion about margins getting in the way.


wacii:
`<HeadingWithDetail>` expanded details text does not wrap. I don't know if the problem was introduced here. See the _Unreimbursed Disabled/Handicapped/Elderly Assistance_ header under the _Expenses_ tab for an example.

It's a little weird to have a wrapper like `<div className = { `top horizontal` } >` when there is no `Top` content, but maybe that is still needed for formatting?

knod:
Was hidden to me because I'm constantly in Dev HUD mode! Fixed!

PR #730, opened 2018-09-05
Parker/issue 513/use snippets in benefits table

pporfilio:


src/forms/output/BenefitsTable.js
```diff
@@ -129,60 +130,60 @@ const BenefitsTable = function (props) {
...
 
-  const TotalsRow = function (props) {
+  const TotalsRow = function ({ snippets }) {
     return (
       <Table.Row style={{ border: 'none' }}>
         <Table.Cell
           textAlign='right'
           width={ 3 }
-          style={ totalsRowHeaderStyle }>Net Total
+          style={ totalsRowHeaderStyle }>{ snippets.rowNetTotal_v1}
```
knod: Looking great to me! May want to `npm run lint` to clean up a couple of things (does that work with JSX?).

Would also appreciate other opinions.

knod:
Do you think you'd also be able to update the google docs when this is finalized?

src/localization/en.js
```diff
@@ -264,6 +264,19 @@ export default {
...
+      columnCurrentBenefits_v1:   `Current Benefits`,
+      columnNewEstimate_v1:       `New Estimate`,
+      columnDifference_v1:        `Difference`,
+      rowSNAP_v1:                 `SNAP`,
+      rowSection8_v1:             `Section 8 Housing`,
+      rowTotalBenefits_v1:        `Total Benefits`,
+      rowIncome_v1:               `Income`,
+      rowNetTotal_v1:             `Net Total`,
+      perMonth_v1:                ` / month`,
+      dollarSign_v1:              `$`,
```
ethanbb: I don't think the dollar sign needs to be a snippet - would it ever be different in another language? Same with the slash before "month".

I don't really have a problem with it, just found it slightly confusing.

src/localization/en.js
```diff
@@ -264,6 +264,19 @@ export default {
...
+
+      benefitsTableTitle_v1:      `Changes`,
+      columnBenefit_v1:           `Benefit`,
+      columnCurrentBenefits_v1:   `Current Benefits`,
+      columnNewEstimate_v1:       `New Estimate`,
+      columnDifference_v1:        `Difference`,
+      rowSNAP_v1:                 `SNAP`,
+      rowSection8_v1:             `Section 8 Housing`,
+      rowTotalBenefits_v1:        `Total Benefits`,
+      rowIncome_v1:               `Income`,
```
ethanbb: Slight nitpick, some of these (like SNAP and Section 8) could be named more generically since they'll probably be reused in other tabs.

src/forms/output/BenefitsTable.js
pporfilio: npm run lint didn't make any changes to this PR.

src/localization/en.js
pporfilio: I think it's generally a good idea to have any string constant in translation files. That way the developer doesn't have to know if there's any language where it would need to be translated.

Now that you mention it though, I think the order can be different in different languages. I'm used to doing something like `translateString("$%1 / month").format(dollarAmount)`.

It's not immediately obvious to me how to do this with our setup (or even in javascript in general without a 3rd party library), and maybe that's better left to a separate PR?

src/forms/output/BenefitsTable.js
knod: Hmm, I guess it doesn't do much with JSX. Maybe you'll have to edit this one manually. I know we were having trouble getting it to behave properly in that context or something.

src/localization/en.js
ethanbb: If we wanted to make it more flexible, we could treat each string like `$20 / month` as an interpolated snippet, so in English it would be like ``[ `$`, { name: `__monthlyAmount__` }, ` / month`]`` - if we had a way to delay interpolation from happening until the actual calculation of the amount occurred. (In this example, the interpolated component would just be `{ __monthlyAmount__: <span>{ amount }</span> }` and then you would call `interpolateText` on the array and that object within `BenefitsTable`.) 

Maybe this would work if we returned the item unchanged in `interpolateText` if the name wasn't found, rather than returning `null`.

knod: If we're going to get this complicated about that, maybe we should really consider adding a lib, like @pporfilio said. I think there was one out there that pluralized stuff, but I'm not sure if it was able to change the order of the text. There must be one that does let you do that, though. Then again, I'm not sure if it would play nice with the interpolation stuff. It does feel a bit like we're putting together Frankenstein right now.

If we were to do that kind of interpolation ourselves, wouldn't we need a way to interpolate in-place? @ethanbb : maybe your proposal would do that, but I wasn't sure how... I wonder if one of the other options I suggested for interpolation in #740 might have a good way to handle this. I've put it in the list for possible discussion/breakout.

knod: Hmm,. maybe `</T>` could have a `component` prop (which would be the snippet) and also children (which would be the changeable text), though I'm not quite sure how that would in a situation, for example, where there were multiple dynamic things going on...

ethanbb: Yeah there are defnitely downsides and simplifications in what I suggested @knod, and I also feel like we should look for a more elegant way to do all this.

Not sure I understood your second comment. What's `<T>` in your example?

Another way to do this would be to have the snippet be a function of the dollar amount. We don't currently support snippets that are functions (and it would probably mess up the language property thing), but it would be cleaner than what I suggested.

knod: `<T>` is mentioned in #740 and might be discussed during the meeting. Functions for snippets was one of the ideas I was going to propose for interpolated snippets, but I'm not yet sure how we'd use them in what we have currently to be used for dynamic input. I have a thought about something like `var interp = function ({ snippets (with .langCode?), children, data/whatever? }) {...}` that's floating around at the moment...

ValerieKenyon: BTW the reason I had made two snippets for "$" and " / month" is because some languages do have a different order.  (We're not doing Thai, but Thai is one of those languages.)

ethanbb:
@knod I agree - my comments were very minor to begin with and I didn't mean for the dollar sign thing to turn into a long discussion! In any case it should be easy enough to make the pre- and post-number snippets whatever they need to be in each language to get the desired result. 

(Although looks like now this has merge conflicts that need resolving.)

pporfilio:
I can clean this up at the end of the afternoon today or during the meeting tonight.

pporfilio:
I rebased on dev and fixed the whitespace mistake you pointed out.

ethanbb:
Looks like there's a problem with the `package-lock.json` (in this and some other PRs). Can you try deleting that file, running `npm install` to regenerate it, and committing the result?

knod:
@ethanbb : With past experiments, that doesn't fix the problem with package-lock.json because of all the npm nonsense. You end up having to basically undo the changes. Usually I'd say just go back to dev and redo the changes, but someone else had this problem and I ended up looking it up. This is approximately what I found, though I haven't gotten a chance to test it yet:

1. Be on your branch and make sure you don't have any other changes right now
2. `git checkout -b test-fix` (to make a new branch to test out whether this will work)
2. `git fetch upstream-name-you-have dev`
3. `git checkout upstream-name-you-have/dev -- package-lock.json`
4. `git add package-lock.json`
5. `git commit`
7. If that works, do it on the actual PR branch.

Got some of this from https://stackoverflow.com/a/215731.

Tell me how it goes!

ethanbb:
@knod Ah makes sense. Although supposedly this PR doesn't modify `package-lock.json`, so couldn't it be the file on `dev` that's broken?

As a data point, my PR that involved installing some new packages is now working in Travis. So hopefully at least it's a transient issue.

knod:
@ethanbb : I did an `npm install` for dev locally and pushed it to the CfB repo. That seemed to fix it. That does need to be pulled in here.

knod:
Oh, there are two approvals! Nevermind, I'll just merge then.

PR #731, opened 2018-09-05
Add maximum value to income/expense text fields

2001zhaozhao:
Adds a $999999.99/yr ($19230.76/wk, $83333.33/mo) maximum value to the fields for income and expenses. Exceeding the maximum value will not update the number and shows a warning, just like how the monthly rent verification works.

To implement this I had to change the CashFlowInputsRow object from a method to a class, and I am not 100% sure that I did the conversion correctly, so please look over what I did in the commits.

### DIFF
--- a/src/forms/cashflow.js
+++ b/src/forms/cashflow.js
@@ -1,5 +1,5 @@
 // REACT COMPONENTS
-import React from 'react';
+import React, { Component } from 'react';
 import { Form } from 'semantic-ui-react';
 
 // PROJECT COMPONENTS
@@ -10,6 +10,7 @@ import { InvalidMessage } from './formHelpers';
 import { toMonthlyAmount } from '../utils/math';
 import { isNonNegNumber, hasOnlyNonNegNumberChars } from '../utils/validators';
 import { toMoneyStr } from '../utils/prettifiers';
+import ChartComponent from 'react-chartjs-2';
 
 
 /** Contains cash flow inputs, their label, and any user feedback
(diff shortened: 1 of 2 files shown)


src/forms/cashflow.js
```diff
@@ -42,68 +43,95 @@ const CashFlowContainer = function ({ children, label, validRow, message }) {
...
+        };
+
+    var cashFlowStoreValidator = this.cashFlowStoreValidator;
+  
+    return (
+      <CashFlowContainer
+        label={ children }
+        validRow={ !this.state.message }
+        message={ this.state.message }>
+        <ManagedNumberField storeValidator={cashFlowStoreValidator(19230.76)}
```
ethanbb: Instead of hard-coding the limits in this and the following lines (magic numbers), it would be better to assign the number that is the source of the others (999999.99) to a descriptively-named variable, and then obtain the others through arithmetic. That way, if the input constraints change, we only have to change one number and it's easier to understand how to do that.

src/forms/cashflow.js
```diff
@@ -42,68 +43,95 @@ const CashFlowContainer = function ({ children, label, validRow, message }) {
...
+        return false;
+      }
+      else if(parseFloat(str) > max) {
+        this.setState({ message: 'The input number exceeds the maximum of $999,999.99/yr' });
+        return false;
+      }
+      this.setState({ message: '' });
+      this.state.message = ''
+      return true;
+    }.bind(this); // bind this too
```
ethanbb: If you return an arrow function instead (`return (str) => { ...`), you don't have to call `.bind(this)` - `this` will automatically refer to the `CashFlowInputsRow` object when the validator is called. We've been preferring the arrow function syntax in this codebase.

src/forms/cashflow.js
```diff
@@ -42,68 +43,95 @@ const CashFlowContainer = function ({ children, label, validRow, message }) {
...
+
+  // Special store validator that handles maximums and sets error message
+  cashFlowStoreValidator(max) {
+    return function (str) {
+      if(!isNonNegNumber) {
+        this.setState({ message: 'Invalid number format' });
+        return false;
+      }
+      else if(parseFloat(str) > max) {
+        this.setState({ message: 'The input number exceeds the maximum of $999,999.99/yr' });
```
ethanbb: It would be nice to use a template string to insert the actual maximum value in the string here, assuming we're going to store it in a variable. For example: `` `The input number exceeds the maximum of $${ maxYearly }/yr` ``

src/forms/cashflow.js
```diff
@@ -42,68 +43,95 @@ const CashFlowContainer = function ({ children, label, validRow, message }) {
...
+    super(props);
+    this.state = { message:'' }
+    // Bind the validator to set message
+    this.cashFlowStoreValidator = this.cashFlowStoreValidator.bind(this);
+  }
+
+  // Special store validator that handles maximums and sets error message
+  cashFlowStoreValidator(max) {
+    return function (str) {
+      if(!isNonNegNumber) {
```
ethanbb: This should be calling `isNonNegNumber` on `str` rather than using it as a value.
I like the idea of adding a message for invalid numbers though!

src/forms/cashflow.js
```diff
@@ -42,68 +43,95 @@ const CashFlowContainer = function ({ children, label, validRow, message }) {
...
+        this.setState({ message: 'Invalid number format' });
+        return false;
+      }
+      else if(parseFloat(str) > max) {
+        this.setState({ message: 'The input number exceeds the maximum of $999,999.99/yr' });
+        return false;
+      }
+      this.setState({ message: '' });
+      this.state.message = ''
```
ethanbb: Looks like this was left in by accident? The line above is the correct way to do it.

2001zhaozhao:
I couldn't get the template string to work, but I've fixed everything else. Thanks for the review!

knod:
Great job! Go ahead and merge dev in again and see if the tests pass!

2001zhaozhao:
Just updated validate function to use arrow function. Thanks again!

ethanbb:
Thanks!

PR #732, opened 2018-09-05
Fixes #725 Link Cliff Effects Tool to github repo

### LINKED ISSUE #725: Edit footer 'Cliff Effects Tool' to link to github repo
I was originally thinking of this as a task for a new coder, but considering it'll involve the interpolating, it may be more complex than that. What do folks think? Will it be straightforward enough to assign it to someone new?

jlangevin:
I took a stab at issue #725.  I added an inline component under the footer section that is an external link component without text.  This allows the settings in the localization files to pass text instead of it being hard-coded.  I went that route since the about page also has a link to the repo, but with text of 'GitHub'.  Perhaps makes sense to standardize and have only one github inline component.  This one could easily be reused by the about page with the appropriate updates to the localization files.

### DIFF
--- a/src/localization/de.js
+++ b/src/localization/de.js
@@ -6,7 +6,12 @@ export default {
   header: {},
 
   footer: {
-    header:    `Cliff Effects Tool`,
+    header: [
+      {
+        name: `__githubRepoLink__`,
+        text: `Cliff Effects Werkzeug`,
+      },
+    ],
     cfbCredit: [
       `Mit `,
       { name: `__heartIcon__` },
--- a/src/localization/en.js
+++ b/src/localization/en.js
@@ -6,7 +6,12 @@ export default {
   header: {},
 
   footer: {
-    header:    `Cliff Effects Tool`,
+    header: [
+      {
+        name: `__githubRepoLink__`,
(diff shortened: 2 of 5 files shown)


knod:
Thanks for this PR!

As for re-usability, right now it does look reusable. Honestly, I think I didn't even notice it there when I made the issue. I'm still not sure that's the way to go. My thinking is that if we change our minds and want it to link to, for example, Code for Boston's home page, we'd want to be able to change that independently. Maybe that's too similar to pre-optimization, I'm not sure. I will admit that if we change our repo address or something, we'd have to change it in both places, which is more fragile. I'm ready to hear opinions.

In fact, I'm leaning more towards re-using it now. Well, shucks. @ethanbb , @pporfilio , @wacii , or anyone else that wants to weigh in?

jlangevin:
You're very welcome.  I'm happy to tweak things further, if folks think other changes are needed.

knod:
@ethanbb , @pporfilio , @wacii , and others - would still appreciate your input on reuse.

wacii:
It only appears twice and is not likely to change all that often. Don't think its worth worrying about.

Issue #708, opened 2018-08-15
Make JSDoc comments more useful


ethanbb:
Currently, there are several places in the (now publishable) JSDoc comments where:

* Things are missing (or "TODO")
* Things that shouldn't be documented (minor local variables) are documented
* A comment refers to the wrong entity
* There are (apparently) several entries on the same object, or objects with the same name

Let's clean these up so that the docs make sense!

knod:
This sounds great! This would be a great place to take note of those locations when we spot them.

Regarding the documentation of minor variables - I remember doing that a couple of times when I knew less about jsdoc and it seemed appropriate to clarify something. Maybe we can just change those to regular comments if they're still relevant.

skudbucket:
I'd be interested in taking this on, but have some questions (I've never used JSDoc before, but I just read up on some on their [documentation ](http://usejsdoc.org/) and feel ready to give this a try, with a little guidance):
1) How do I generate the actual documentation from the comments, so that I can see what the current result looks like and what the problems with it are? Do I just need to download the JSDoc tool locally and manually generate the current documentation? Or is there a fancier way?
2) Are we using the latest version of JSDoc/can I use it in writing description comments? I'm wondering since some of the useful-seeming features of JSDoc 3 (ie support for parsing ES 2015 classes) appear to have support only in relatively recent versions (3.4+).
3) Is it accurate that all classes and all functions in all of the JS files in the whole cliff-effects repo should ideally have JSDoc description comments?
4) Is there anything besides classes and functions that should get JSDoc description comments?

knod:
That would be awesome! Thanks, @skudbucket! I'll answer what I can, and leave the rest to @ethanbb : 
1. @ethanbb should also weigh in on other solutions because I don't quite understand why the docs don't appear locally, but if nothing else works, and you're working on your fork, not our repo, you can do `npm run deploy` and it should create a github page for your repo where you can see the docs page. Other commands, if you're interested in them, are also in 'package.json'.
2. @ethanbb
3. My opinion is that it's not always necessary. If something seems so self-explanatory that the documentation is just redundant, it doesn't need a description. If you're not sure and concerned, though, feel free to add one! Or to ask for input from others, if you're more comfortable with that.
4. First off, just getting the functions and classes is a great start, and there's plenty of them. If we get to a point where we're worrying about that, I think I'll throw us a party :) Second, to answer teh question as best I can - class methods could use descriptions sometimes. I'm not sure about the convention for anything else.

[...]

PR #736, opened 2018-09-06
add jsDOC documentation to App.js

skudbucket:
Aimed at starting to address issue [#708](https://github.com/codeforboston/cliff-effects/issues/708).

PR for my first file that I've done jsDOC documentation for (like @knod said, "commit and PR early and often"), to make sure that I'm doing it correctly before I go and do a bunch of others. Please take a look and let me know if it looks good, any feedback about content, formatting, or whatever else or the documentation comments would be appreciated.

Also, note that I'm using the `@techExpertisePlease` flag in my documentation comments to flag places where I came across something that I couldn't definitively say myself how to correctly document it, and could use someone with a better technical understanding of that piece of the code taking a look at it and maybe clarifying that piece of documentation.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -80,7 +80,9 @@ class App extends Component {
     this.setState({ language: inputProps.value, snippets: snippets });
   };
 
-  /** Set the value of a specified key in the app state's devProps
+  /** Set the value of a specified key in the app state's devProps.
+   * These keys should only be set to boolean values (@todo enforce only allowing boolean values?).
+   * Keys with a value of true are added (later, not in this function) as classes to the app's main element.
    * @param {string} key - The key whose value is to be changed in the app state's devProps
    * @param {boolean} value - The value to be set for the given key in the app state's devProps
    */
--- a/src/App.js
+++ b/src/App.js
@@ -28,8 +28,11 @@ import { CLIENT_DEFAULTS } from './utils/CLIENT_DEFAULTS';
 // LOCALIZATION
 import { getTextForLanguage } from './utils/getTextForLanguage';
 
-/** App component; main top-level component of the app */
-// Change HashRouter tags below to Router tags to turn off hash routing; only used to be compatible with GitHub Pages
+/** App component; main top-level component of the app.
+ * You can hange the HashRouter tags in the App class declaration (below if you are viewing this comment in the source code) to Router tags to turn off hash routing; only used to be compatible with GitHub Pages.
+ * The App component also manages the Dev HUD.
+ * The App component also manages the header and footer that appear on every page except the home page.
+ * The App component sends in the initial client values */
 class App extends Component {
   /**
    * Create App component instance
--- a/src/App.js
+++ b/src/App.js
@@ -35,8 +35,9 @@ import { getTextForLanguage } from './utils/getTextForLanguage';
  * The App component sends in the initial client values */
 class App extends Component {
   /**
-   * Create App component instance
-   * @param {object} props - React props passed to the App component
+   * Create App component instance.
+   * @techExpertisePlease @knod referenced some [confusing state property handling](https://github.com/codeforboston/cliff-effects/pull/736#discussion_r215761120), anyone with the know-how wanna tackle explaining that?
+   * @param {object} props - React props passed to the App component. These props are only used in the `super(props)` call, and not in the rest of the function.
    */
   constructor (props) {
     super(props);
--- a/src/App.js
+++ b/src/App.js
@@ -100,8 +100,9 @@ class App extends Component {
     });
   };  // End setDev()
 
-  /** Load an individual client's data @techExpertisePlease from what source? Where does toLoad come from?
-   * @param {object} toLoad - An object representing the client data to be loaded @techExpertisePlease does this have a specific shape? Or format (ie is it JSON or something?)
+  /** Load an individual client's data.
+   * The client data to load comes from a text input field in the Dev HUD
+   * @param {object} toLoad - A JSON object representing the client data to be loaded. Must match the client data format (See {@link CLIENT_DEFAULTS} for an example of the correct client data format)
    */
   loadClient = ({ toLoad }) => {
     this.setState((prevState) => {
--- a/src/App.js
+++ b/src/App.js
@@ -71,9 +71,9 @@ class App extends Component {
     };
   };  // End constructor()
 
-  /** Set the language of the app
-   * @param {object} evnt - An event object, which is not actually used in the function but is passed in by Semantic UI React input components
-   * @param {object} inputProps - An object representing the properties of the Semantic UI React input component which triggered the language change
+  /** Set the human language of the app (ie the language in which the UI will display text for users to read, NOT the coding language).
+   * @param {object} evnt - An event object, which is not actually used in the function but is passed in by Semantic UI React input components.
+   * @param {object} inputProps - An object representing the properties of the Semantic UI React input component which triggered the language change.
   */
   setLanguage = (evnt, inputProps) => {
     var snippets = getTextForLanguage(inputProps.value);
--- a/src/App.js
+++ b/src/App.js
@@ -28,8 +28,13 @@ import { CLIENT_DEFAULTS } from './utils/CLIENT_DEFAULTS';
 // LOCALIZATION
 import { getTextForLanguage } from './utils/getTextForLanguage';
 
+/** App component; main top-level component of the app */
 // Change HashRouter tags below to Router tags to turn off hash routing; only used to be compatible with GitHub Pages
 class App extends Component {
+  /**
+   * Create App component instance
+   * @param {object} props - React props passed to the App component
+   */
   constructor (props) {
     super(props);
 
@@ -62,11 +67,19 @@ class App extends Component {
     };
   };  // End constructor()
 
+  /** Set the language of the app
+   * @param {object} evnt - An event object, which is not actually used in the function but is passed in by Semantic UI React input components
+   * @param {object} inputProps - An object representing the properties of the Semantic UI React input component which triggered the language change
+  */
   setLanguage = (evnt, inputProps) => {
     var snippets = getTextForLanguage(inputProps.value);
     this.setState({ language: inputProps.value, snippets: snippets });
   };
 
+  /** Set the value of a specified key in the app state's devProps
+   * @param {string} key - The key whose value is to be changed in the app state's devProps
+   * @param {boolean} value - The value to be set for the given key in the app state's devProps
+   */
   setDev = (key, value) => {
     this.setState((prevState) => {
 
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -62,11 +67,19 @@ class App extends Component {
     };
   };  // End constructor()
 
+  /** Set the language of the app
+   * @param {object} evnt - An event object, which is not actually used in the function but is passed in by Semantic UI React input components
```
knod: Nice! I like this description!

knod: Which is weird, because this is _also_ a description of React behavior...

src/App.js
```diff
@@ -62,11 +67,19 @@ class App extends Component {
...
+   * @param {object} inputProps - An object representing the properties of the Semantic UI React input component which triggered the language change
+  */
   setLanguage = (evnt, inputProps) => {
     var snippets = getTextForLanguage(inputProps.value);
     this.setState({ language: inputProps.value, snippets: snippets });
   };
 
+  /** Set the value of a specified key in the app state's devProps
```
knod: Also, currently these keys should only be set to bools (which maybe should be enforced sometime) and will be added as classes to the app's main element if they're value is `true`. This might need to change sometime to allow more flexibility with the dev hud, but I don't currently have a good picture in my head of implementation...

src/App.js
```diff
@@ -81,6 +94,9 @@ class App extends Component {
     });
   };  // End setDev()
 
+  /** Load an individual client's data @techExpertisePlease from what source? Where does toLoad come from?
```
knod: From a text field input available in the Dev HUD.

src/App.js
```diff
@@ -81,6 +94,9 @@ class App extends Component {
...
+   * @param {object} toLoad - An object representing the client data to be loaded @techExpertisePlease does this have a specific shape? Or format (ie is it JSON or something?)
```
knod: It's a JSON object that needs to match the client data format.

knod: Which can be seen in 'src/utils/CLIENT_DEFAULTS.js', though that could move.

src/App.js
```diff
@@ -28,8 +28,13 @@ import { CLIENT_DEFAULTS } from './utils/CLIENT_DEFAULTS';
 // LOCALIZATION
 import { getTextForLanguage } from './utils/getTextForLanguage';
 
+/** App component; main top-level component of the app */
 // Change HashRouter tags below to Router tags to turn off hash routing; only used to be compatible with GitHub Pages
 class App extends Component {
+  /**
```
knod: This should probably be above the class declaration, like the other function descriptions. Unless someone knows otherwise.

src/App.js
```diff
@@ -92,6 +108,9 @@ class App extends Component {
     });
   };  // End loadClient()
 
+  /** Convert an object (ie devProps object) to a... string? @techExpertisePlease not too familiar with backticks and template literals... is this a string being created? Or a template? Or something else?
```
knod: It's creating an array of strings - a list of class names. The strings used are the keys of the object.

src/App.js
```diff
@@ -28,8 +28,13 @@ import { CLIENT_DEFAULTS } from './utils/CLIENT_DEFAULTS';
 // LOCALIZATION
 import { getTextForLanguage } from './utils/getTextForLanguage';
 
+/** App component; main top-level component of the app */
 // Change HashRouter tags below to Router tags to turn off hash routing; only used to be compatible with GitHub Pages
 class App extends Component {
+  /**
+   * Create App component instance
```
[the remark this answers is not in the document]
knod: (Current discussion pending - is this description of React behavior a useful one?)

src/App.js
```diff
@@ -102,6 +121,7 @@ class App extends Component {
     return classes;
   };  // End propsToClasses()
 
+  /** Render the App component */
```
knod: (Current discussion pending - is this description of React behavior a useful one?)

src/App.js
```diff
@@ -28,8 +28,13 @@ import { CLIENT_DEFAULTS } from './utils/CLIENT_DEFAULTS';
 // LOCALIZATION
 import { getTextForLanguage } from './utils/getTextForLanguage';
 
+/** App component; main top-level component of the app */
```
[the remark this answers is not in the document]
skudbucket: My intention was to identify the declaration of the class/component itself as distinct from the declaration of the constructor function within the class. Maybe what I put didn't do that super clearly (open to suggestions about how to be moreso, again pending discussion about whether we even include these basics in the docs) but my goal was to label each as what it is.

General JSDoc thing to note (I'm super new to JSDoc as well, but have been reading up): in JSDoc a documentation comment immediately precedes whatever it applies to. So whenever in doubt about what is being documented in a given comment, the answer ought to be whatever immediately follows the documentation comment.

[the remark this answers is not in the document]
skudbucket: In general, JSDoc guidelines seem to recommend documenting both classes and methods within them (ie in this case the constructor function here). Obvi what we choose to do is up to us tho

src/App.js
knod: Ah, this is the description of the constructor, as your comments above explain. Nvm.

src/App.js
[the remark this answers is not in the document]
knod: Slight change: The first notes should actually go in the actual App description, not the constructor description. The constructor should be about anything special happening with the, obviously, the construction of the component. E.g. What confusing `state` properties it creates and why. And why did they have to be made to be so confusing, I guess.

src/App.js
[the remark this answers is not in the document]
knod: Some things that may be helpful to include:
- The note about `HashRouter` and gh-pages.
- It manages the Dev HUD too.
- It manages the header and footer that appear on every page except the home page.
- It sends in the initial client values.

Not sure all of that needs to be included, but those are some things we can discuss.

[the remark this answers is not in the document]
skudbucket: These seem like potentially useful inclusions to me, enough so that I'll go ahead and include them at least for now so that the PR is reflecting a "current potential most promising" version as we discuss rather than just my initial version until we settle on some things in discussion. Further thoughts from anyone welcome though!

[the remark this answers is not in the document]
skudbucket: Implemented in [2bfee53](https://github.com/skudbucket/cliff-effects/commit/2bfee53a7ea49dea0c13c6ac17387b4f37f135fa)

src/App.js
```diff
@@ -28,8 +28,13 @@ import { CLIENT_DEFAULTS } from './utils/CLIENT_DEFAULTS';
...
 // Change HashRouter tags below to Router tags to turn off hash routing; only used to be compatible with GitHub Pages
 class App extends Component {
+  /**
+   * Create App component instance
+   * @param {object} props - React props passed to the App component
```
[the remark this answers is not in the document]
skudbucket: I don't think I fully grasp the significance of this in terms of its impact on the big picture app architecture, but am gonna err on the side of inclusion for now pending further discussion...

src/App.js
skudbucket: Lol to be fair, it is a description of Semantic UI React behavior, so that's a little more specific than just general React basics!

src/App.js
```diff
@@ -62,11 +67,19 @@ class App extends Component {
     };
   };  // End constructor()
 
+  /** Set the language of the app
```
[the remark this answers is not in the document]
skudbucket: Oh haha that seems like a very worthy clarification to me then. Addressed in [c76418b](https://github.com/skudbucket/cliff-effects/commit/c76418bcba85954942ae84cc459d3c2066b1a7e1)

src/App.js
skudbucket: So, people should correct me if I'm wrong, but I think that by having the "`{boolean}`" in `@param {boolean} value - The value to be set...`, we might already be expressing in the way JSDoc intends that the value should only be a boolean. But perhaps this bears repeating in the description too. I think the part about the keys being added as classes certainly seems like a good idea to include. (I've done so in 65005253efce1890daeeb30c38d74c8ea9b20a16)

src/App.js
skudbucket: Thanks! Updated the JSDoc description for `loadClient` in 269d6dfa6df06d38148aed95d0c563c1dbb3f85b (with even a link to the CLIENT_DEFAULTS that won't break even if it moves! Wow, JSDoc is so fancy! Although this link is to the documentation for the CLIENT_DEFAULTS constant, rather than to the CLIENT_DEFAULTS.js source code, so it will only be as useful as the documentation for CLIENT_DEFAULTS is)

src/App.js
skudbucket: Also, @knod are you sure it's an array of strings and not just a single, space separated, concatenated string that it returns? (That was my reading...)

skudbucket:
Implemented a bunch of these! Didn't realize that doing so would make the conversations @knod started above get marked as "outdated" - still check them out because most of them are not and could still use discussion!

It is my hope that once we've worked out some general guidelines about how specific the documentation should be and how much basic stuff should be included in it, then we won't need to have too much discussion of that going forward as the documentation process continues along!

src/App.js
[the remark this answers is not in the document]
knod: I mean, should we add the info that our code doesn't use those props at all?

src/App.js
knod: Template literals: We're actually hoping to convert as much of the code base as we can to those - it'll mean less escaping of actual quotation marks of any kind. To me personally, it seems like if we only use them when there are quotes around, that inconsistency would be more confusing. Maybe I'm off base there, though. It can definitely be discussed.

Returned value: Sorry, you're right, it's a string of words separated by spaces.

src/App.js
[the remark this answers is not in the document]
knod: Btw, I just meant 'confusing state properties' in general. I'm not sure if there are or aren't any here. I'll have to check it out.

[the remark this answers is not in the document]
ethanbb: In my opinion it's probably worth giving at least a short description for each state property (after all, there are only four), but at least `clients` and `devProps`. Also, maybe these descriptions could be directly before `this.state` is defined, rather than before the constructor?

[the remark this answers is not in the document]
knod: Interesting! How would jsdoc interpret that? I mean, where would it appear visually in the docs?

src/App.js
knod: For the record, I do think that's a great point about the bools already having a type. I sometimes forget how the documentation works :P

src/App.js
knod: Just so you know, there's an `@see` tag in jsdoc, though we're looking for things that will help us actually make the docs visible (talk later) and whatever lib we use may not be that fancy.

src/App.js
knod: So far we've decided that describing React behavior isn't something we want to actively pursue, but that hinting at it in the documentation may be useful. For example, in component classes and functions we may add an `@returns React.node` or `element` or something like that. Not `component` - apparently that's inaccurate because the use of them gets you an instance, not the class itself.

src/App.js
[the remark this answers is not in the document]
knod: @ethanbb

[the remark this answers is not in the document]
ethanbb: I'm not sure; I was looking into whether there's a specific tag to use for class fields defined in the constructor last night, but I didn't find anything. Hopefully if the jsdoc library recognizes ES6 classes, it would know to associate `this.x` definitions within a class with that class - but we can test to make sure that works.

ethanbb:
Taking this over; I'm going to close this and open a PR from my fork.

PR #739, opened 2018-09-06
Fixes big oversight in benefit calculations

knod:
though it doesn't seem to have affected the current cliff example.

Until now SNAP calculations have not been affected by Section 8 benefits. Should be fixed now. Bit embarrassed about this and definitely want another eye on it to make sure it's getting done right this time.

### DIFF
--- a/src/forms/output/BenefitsTable.js
+++ b/src/forms/output/BenefitsTable.js
@@ -39,7 +39,7 @@ const BenefitsTable = function (props) {
     sec8BenefitCurrent     = Math.round(getSection8Benefit(clone, 'current'));
     sec8BenefitFuture      = Math.round(getSection8Benefit(clone, 'future'));
     // Mutate clone for correct SNAP values
-    clone.future.rentShare = sec8BenefitFuture;
+    clone.future.rentShare = (clone.future.contractRent - sec8BenefitFuture);
   }
 
   if (curr.hasSnap) {
--- a/src/forms/output/BenefitsTable.js
+++ b/src/forms/output/BenefitsTable.js
@@ -43,7 +43,6 @@ const BenefitsTable = function (props) {
   }
 
   if (curr.hasSnap) {
-    console.log(clone.future.rentShare);
     SNAPBenefitCurrent = Math.round(getSNAPBenefits(clone, 'current'));
     SNAPBenefitFuture  = Math.round(getSNAPBenefits(clone, 'future'));
   }
--- a/src/forms/output/BenefitsTable.js
+++ b/src/forms/output/BenefitsTable.js
@@ -27,22 +27,25 @@ const getSignSymbol = function (num) {
 
 const BenefitsTable = function (props) {
 
-  var client = cloneDeep(props.client),
-      curr   = client.current;
+  var clone = cloneDeep(props.client),
+      curr   = clone.current;
 
-  var SNAPBenefitCurrent =  0,
-      SNAPBenefitFuture  =  0,
-      sec8BenefitCurrent = 0,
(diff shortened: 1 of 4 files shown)


src/forms/output/BenefitsTable.js
```diff
@@ -27,22 +27,25 @@ const getSignSymbol = function (num) {
...
+    sec8BenefitFuture      = Math.round(getSection8Benefit(clone, 'future'));
+    // Mutate clone for correct SNAP values
+    clone.future.rentShare = sec8BenefitFuture;
   }
 
-  if (curr.hasSection8) {
-    sec8BenefitCurrent = Math.round(getSection8Benefit(client, 'current'));
-    sec8BenefitFuture  = Math.round(getSection8Benefit(client, 'future'));
+  if (curr.hasSnap) {
+    console.log(clone.future.rentShare);
```
wacii: Errant `console.log`.

src/utils/charts/getChartData.js
```diff
@@ -36,32 +75,85 @@ getData.section8 = function (xRange, client, multiplier) {
...
+
+  // Don't loop if there's nothing to do
+  if (benefitDatasets.length === 0) {
+    return benefitDatasets;
+  }
+
+  // Otherwise, loop over incomes
+  for (var incomei = 0; incomei < xRange.length; incomei++) {
+    var income = xRange[ incomei ];
+    clone.future.earned = income / multiplier;
```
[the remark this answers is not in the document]
knod: We're talking it over in real-time since I think we're not on the same page. Regardless, I'm not sure I've expressed this well here, so it obviously needs to be clarified somehow.

src/utils/charts/getChartData.js
```diff
@@ -36,32 +75,85 @@ getData.section8 = function (xRange, client, multiplier) {
...
 };  // End getData.section8()
 
+/**
+ * @param {array} xRange All income values to be included.
+ */
+const insertBenefitData = function (xRange, clone, multiplier, benefitDatasets) {
+
+  // Don't loop if there's nothing to do
+  if (benefitDatasets.length === 0) {
+    return benefitDatasets;
```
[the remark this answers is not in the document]
knod: Fair point, a return value can be confusing in a function that's meant to mutate data. And it should also be documented. I wish there were an `@mutates` in jsdoc or something. That'd be neat.

src/utils/charts/getChartData.js
```diff
@@ -36,32 +75,85 @@ getData.section8 = function (xRange, client, multiplier) {
...
+    var income = xRange[ incomei ];
+    clone.future.earned = income / multiplier;
+
+    // Datasets need to be in correct order for benefits to
+    // affect each other corrrectly.
+    for (let benefiti = 0; benefiti < benefitDatasets.length; benefiti++) {
+
+      let dataset = benefitDatasets[ benefiti ],
+          name    = dataset.benefitName,
+          funcs   = getBenefit[ name ];
```
knod: @wacii wrote this, and it's about to be oudated by an unrelated change:
This inner loop and `getBenefit` are too strongly coupled. I would rather see `getBenefit[name]` resolve to a function that does both the subsidy calculation and the update of the client (when necessary).

knod:
Or maybe that commit didn't fix everything...

src/utils/charts/getChartData.js
knod: I like this idea, and it's one I'll work on, but maybe on a different PR? I'd like to get this fix through, even if it's just a temporary one, so that we have a tool that's working.

knod:
@todo on this one.

PR #740, opened 2018-09-07
Fixes styling for footer link, adds flexibility to `ExternalLink`

knod:
This is making me doubt our way of doing snippets. Someone I was chatting with suggested something I think has been suggested before. I think I was one of the people against it then because I wanted to simplify how snippets would be used in the code, but it may be worth revisiting:
```
<span className={ 'neutral-link' }><T snippet={ snippets.name }></span>
```

Another option is to use functions instead of components in 'inlineComponents.js.'

I dunno. It hasn't caused a lot of trouble, so maybe it's not worth it right now. I'm open to opinions.

### DIFF
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -4,7 +4,6 @@ import {
   // Generic Form stuff
   Label,
   Icon,
-  Container,
 } from 'semantic-ui-react';
 
 // ========================================
@@ -16,12 +15,12 @@ import {
  */
 const ExternalLink = function ({ href, children, ...otherProps }) {
   return (
-    <Container
-      href={ href }
-      target='_blank'
+    <a
+      href   = { href }
+      target = { `_blank` }
       { ...otherProps }>
       {children}
-    </Container>
+    </a>
   );
 };
 
--- a/src/forms/formHelpers.js
+++ b/src/forms/formHelpers.js
@@ -4,6 +4,7 @@ import {
   // Generic Form stuff
   Label,
   Icon,
+  Container,
 } from 'semantic-ui-react';
 
 // ========================================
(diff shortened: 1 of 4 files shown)


pporfilio:
What about this is making you doubt the way we're doing snippets? I don't think I understand what the options are that you're referring to. If you'd like to discuss/think about it more maybe we could discuss in a breakout Tuesday. Otherwise, this PR seems like a reasonable extension of the current snippets to me.

knod:
We can definitely discuss it more another time, though I think it doesn't really apply to this PR. It's because of how we use `.cloneElement` and the dance we needed to do to pass classes down properly (the way we need to use `.cloneElement`, it destroys the contents of the component and overrides its properties). With the other methods, it wouldn't be as big of an issue.

ethanbb:
I see your point, but in my opinion 
>it destroys the contents of the component and overrides its properties
is a bit of an exaggeration. It replaces the `children` only if different text is provided, which hasn't been an issue yet. I imagine it could be a problem though if we wanted to use something other than plain text in the `children` of one of the inline components (like bold text inside a link, maybe). As for the properties, it only overrides `key`, `lang`, and `className`; of those, only `className` would probably ever be a problem, and you've fixed that with this PR.

knod:
Well, when I was trying to do this change, I tried a few different ways. One of those ways was by wrapping what was originally the interpolation component with a `<span>` that contained the class I needed, but then only the span got rendered (with the text). You're right that it doesn't endanger other stuff right now, and come to think of it, I could probably have just put it in the code that's using the snippet (someday I'll double check that). Hmm, maybe that's not a real solution, though, since with arrays you don't really get to do that...

I dunno, maybe it's already solved.

Issue #527, opened 2018-05-23
Make script/test to check whether localization files are valid or how much they are missing


ethanbb:
Come up with some easy way to automatically check the validity and/or coverage of localization files, relative to `en.js`.

Google localization docs: https://drive.google.com/drive/folders/1QDCwxmTWZ8-9nLUg_b0M47UYaHXFR3n0

jlangevin:
The pull request I just submitted affected the localization files.  Assuming that PR is accepted, I'd love to take a crack at this since it relates to what I just worked on.

ethanbb:
@jlangevin Sounds good! This should be a lot easier than when the test was first proposed now that we're putting snippet versions in the JS files.

knod:
Great! I saw this originally and then couldn't track down where I saw it!

I was going to say that it may be possible to use something similar to the code in the google docs' scripts, but I now remember it's a complete hack because the g-docs can't treat the text like code. So nvm! Definitely feel free to run away with that!

Also @pporfilio is doing a lot of the heavy lifting with the snippets, so it may be worth coordinating with him on this localization stuff.

jlangevin:
@knod and @pporfilio Here's the direction I started tinkering with based upon the current info on this thread.  But let me know if you have thoughts/concerns/other ideas.

Import localizations from src/localization/all.js and loop over each, comparing to EN.  This would ensure new localizations automatically get tested as they get added instead of pulling in specific localization files.

Use a recursive compare function which accepts two localization objects (and their names/keys for test output purposes).  Below would effectively walk down the tree of the EN localization object checking ES for the same structure/keys, but would not compare the underlying snippet text which is specific to the locale.

compare( key1, obj1, key2, obj2 ) {
 ...logic and tests here...

}
compare('EN', localizations['en'], 'ES', localizations['es']);

General function logic:

Test the two arguments for same type
  If so, check if we have an array/object
    if Array
      Test length
      if same length, loop over elements calling recursive function on each
   if Object
     Test Object.keys().sort() equal
     if same, loop over object properties calling recursive function on each
   Otherwise, its our localized text, so stop comparing/simply return true

jlangevin:
This would be test/localization/all.test.js.  Test-wise, it would test:

1. localizations object exists
2. localizations has property 'EN'
3. localizations has one or more properties not 'EN'
4. compare each non-EN prop to EN using compare()
5. anything else folks think makes sense

knod:
[...]

PR #745, opened 2018-09-11
[WIP] initial checks for localization reports #527

jlangevin:
This is an initial stab at building out the checks for localization reporting.  Test format is just a convenient way of getting decent output until we decide on report format/functionality.

### DIFF
--- a/src/test/localization/WIP_all.test.js
+++ b/src/test/localization/WIP_all.test.js
@@ -1,0 +1,137 @@
+import _ from 'lodash';
+import { localizations } from '../../localization/all';
+
+/*
+Example:
+
+const pathsArray = getKeyPathsArray(
+  {
+    a: { b: '1' },
+    c: 2,
+    d: [ 1, 2, 3, ],
+    e: {
+      f: {
+        g: { h: '' },
+        i: '',
+      },
+    },
+  }
+);
+
+console.log('pathsArray', pathsArray);
+=> [
+  [ 'a' ],
+  [ 'a', 'b' ],
+  [ 'c' ],
+  [ 'd' ],
+  [ 'e' ],
+  [ 'e', 'f' ],
+  [ 'e', 'f', 'g' ],
+  [ 'e', 'f', 'g', 'h' ],
+  [ 'e', 'f', 'i' ]
+]
+*/
+
+const getKeyPathsArray = function(obj, base = []) {
+  // Array to contain our keys paths (if any)
+  let pathsArr = [];
+  
+  // Arrays in localization are interpolated into translated strings, so don't do anything
+  if (Array.isArray(obj)) {
+    return pathsArr;
+
+  // Return an array of paths to each prop using the base path and the key in each loop iteration
+  } else if (typeof obj === 'object') {
+    const keys = Object.keys(obj).sort();
+    
+    // Handle having an empty object
+    if (keys.length === 0) {
+      return pathsArr;
+
+    // Otherwise, loop through the keys and recursively call compare()
+    } else {
+      for (let key of keys) {
+        // Append this key to the base path array
+        let newBase = [ 
+          ...base,
+          ...[ key ],
+        ];
+        pathsArr.push(newBase);
+
+        // Pass this object back to our function to get paths to any children
+        let childPaths = getKeyPathsArray(obj[ key ], newBase);
+        
+        // Add any child paths to our array before returning
+        pathsArr = pathsArr.concat(childPaths);
+      }
+      return pathsArr;
+    }
+
+  // Otherwise, we've got some sort of primative such as a str, num, etc, so don't do anything
+  } else {
+    return pathsArr;
+  }
+};
+
+
+// Get the list of non-EN localizations to test
+let localizationKeys = Object.keys(localizations);
+const enIndex = localizationKeys.indexOf('en');
+
+// Only run our comparisons if EN localization exists
+if (enIndex !== -1) {
+
+  const keyPaths = {};
+  for (let key of localizationKeys) {
+    // Get the key paths from the localization files
+    keyPaths[ key ] = getKeyPathsArray(localizations[ key ]);
+  }
+
+  // Remove EN from our keys to loop over
+  localizationKeys.splice(enIndex, 1);
+  
+  // Loop through all non-en localizations
+  for (let i = 0; i < localizationKeys.length; i++) {
+    let locKey = localizationKeys[ i ];
+   
+    describe(`localization ${locKey}`, () => {
+    
+      // Test for keys in 'EN' that aren't localization being tested
+      for (let j = 0; j < keyPaths[ 'en' ].length; j++) {
+        let keyPath = keyPaths[ 'en' ][ j ];
+        let keyPathAsStr = keyPath.join('.');
+        let keyExistsInLoc = _.has(localizations[ locKey ], keyPath);
+
+        it(`'en' path ${keyPathAsStr} exists in '${locKey}'`, () => {
+          expect(keyExistsInLoc).toBe(true);
+        });
+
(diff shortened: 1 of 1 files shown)


jlangevin:
I'm not familiar with Travis CI.  Not sure why it's giving a missing package warning when this is using packages already part of the repo.  Or is this because my file contains tests and code coverage hasn't been set up yet in Travis CI?

knod:
@jlangevin : That error was because of something weird going on with the main repo. If you pull dev again, it should go away. Thanks for the work!

src/test/localization/WIP_all.test.js
```diff
@@ -0,0 +1,137 @@
...
+      f: {
+        g: { h: '' },
+        i: '',
+      },
+    },
+  }
+);
+
+console.log('pathsArray', pathsArray);
+=> [
```
knod: This is an interesting idea! I like the concept of 'paths'. I wonder if something more like this could be more succinct and so easier to digest:
```js
[
  a.b,
  c,
  d,
  e.f.g.h,
  e.f.i,
]
```
I assume that if items were missing, those items would be shown in a list of their own, so something were missing part way through an object chain (for example, g.h didn't exist), it could be shown as `e.f`.

Thoughts welcome!

knod: If this is the suggested output format, that is.

src/test/localization/WIP_all.test.js
```diff
@@ -0,0 +1,137 @@
...
+          // Get the value of our key so we can test its type
+          let locVal = _.get(localizations[ locKey ], keyPath);
+          
+          // We'll only compare the values of strings and arrays.
+          // Use a simple deep compare since arrays eventually 
+          // become strings and we want to know if the text elements
+          // of the array have been translated
+          if (typeof locVal !== 'object' || Array.isArray(locVal)) {
+            it(`${keyPathAsStr} translation should not be the same as '${locKey}'`, () => {
+              expect(_.isEqual(locVal, _.get(localizations[ 'en' ], keyPath))).not.toBe(true);
```
knod: Sorry about this - you've talked about this aspect of the check for a while I've felt there was something off about that aspect of it, but I couldn't put my finger on what it was. Now I've remembered.

An easy example is Spanish - 'no' in Spanish is also 'no'. In a more complex example, German is apparently very strict about using a language's own name for its pronouns. So, 'Cliff Effects Tool' stays 'Cliff Effects Tool'. That part never gets changed.

src/test/localization/WIP_all.test.js
jlangevin: Think of this as the EN localization object.

```
{
    a: { b: '1' },
    c: 2,
    d: [ 1, 2, 3, ],
    e: {
      f: {
        g: { h: '' },
        i: '',
      },
    },
  }
```
That fed into the function generates this:
```
[
 [ 'a' ],
  [ 'a', 'b' ],
  [ 'c' ],
  [ 'd' ],
  [ 'e' ],
  [ 'e', 'f' ],
  [ 'e', 'f', 'g' ],
  [ 'e', 'f', 'g', 'h' ],
  [ 'e', 'f', 'i' ]
]
```
I can then easily loop through that and check it against any other localization object for missing keys and values which are the same (not translated).  I can do the reverse, feed in DE for instance, get its keys and compare that to EN for any keys that shouldn't be in DE.

The output would be something like what you see in the tests below, but more friendly.  By default, we'd probably just show failures since this gets really long and allow the user to filter.

✓ 'en' path aboutPage exists in 'de'
✓ 'en' path aboutPage.aboutPageHeader exists in 'de'
✓ aboutPage.aboutPageHeader translation should not be the same in 'de'
✕ 'en' path aboutPage.benefitsLinkText exists in 'de'
✕ 'en' path aboutPage.howToUse exists in 'de'
✓ 'en' path aboutPage.howToUseHeader exists in 'de'
✓ aboutPage.howToUseHeader translation should not be the same in 'de'
✕ 'en' path aboutPage.howToUseNote exists in 'de'
✕ 'en' path aboutPage.quantLinkText exists in 'de'
✕ 'en' path aboutPage.videoLinkText exists in 'de'
✕ 'en' path aboutPage.whatFor exists in 'de'
✓ 'en' path aboutPage.whatForHeader exists in 'de'
✓ aboutPage.whatForHeader translation should not be the same in 'de'
✕ 'en' path aboutPage.whatForImportantNote exists in 'de'
✕ 'en' path aboutPage.whoMadeThis1 exists in 'de'
✕ 'en' path aboutPage.whoMadeThis2 exists in 'de'
✕ 'en' path aboutPage.whoMadeThis3 exists in 'de'
✓ 'en' path aboutPage.whoMadeThisHeader exists in 'de'
✓ aboutPage.whoMadeThisHeader translation should not be the same in 'de'

src/test/localization/WIP_all.test.js
jlangevin: Hmm, I was wondering if that sort of thing would crop up.  I saw things like "Cliff Effects Werkzeug" and "Cliff Effects Herriamienta" and was hoping instances not changed were WIP. This is a tough one then.  We could simply not test values, but then we'd lose a big benefit of this comparison.  Any thoughts on that?  Some indicator like a suffix to a key that indicates not to expect a translation?  Or some other metadata to indicate that?

src/test/localization/WIP_all.test.js
jlangevin: Ah sorry about that.  Looking back at it, I can see how you'd think that now.  Especially if you don't actually run the file via Jest to see what it was dumping out to the console.  I'll try to be more conscious of my comments.

src/test/localization/WIP_all.test.js
knod: That's what version names are. `key_v#`.

knod: In the google docs

knod: We're working on bringing the version numbers into the local docs.

jlangevin: I believe I was tripped up by the app name being in a couple places in the loc files and sometimes it was translated and sometimes not.  So I was trying to do string comparison to catch places where an English phase was incorrectly being used as a placeholder.  Sounds like the manual process of versioning should be catching that and any instances of english should be there.  So I can simply remove my string comparison and we're good on that front?  Just check the structure of the doc?

knod: Yeah, I think that's what I was saying. For some reason the key name check hasn't been clicking in my head. Yup, just don't check for a string-to-string match and that sounds good!

jlangevin: I've made a separate pull request for the localization report since this won't live on in the form of tests.  That includes removing the value comparison, and just leaves the key checks.  I think we can close this without merge at this point and pick things up in that one?  #757

knod: Sounds like a plan! Feel free to close when you feel comfortable!

knod:
Hmm, sorry, the tests weren't passing. I didn't spot that somehow.

knod:
Also, this was supposed to be closed, not merged, right?

knod:
Ok, I think I see what happened - you manually undid these changes on another branch and when that got merged, this got merged too.

jlangevin:
Hmmm... this is weird.  I didn't merge this.  You are correct that I merged a different branch for the report yesterday.  I has deleted the original file with the tests in that branch.  I didn't realize if would affect things in this PR.  Sorry about that.  I assume now that I should have deleted the file from this branch, pushed that so this PR gets updated and then in the other branch only do the replacement work?

jlangevin:
Or should I have just done all the work in this one branch?

knod:
We'll talk about it in more detail on Slack. Easier in real-time.

PR #746, opened 2018-09-12
 Make docs work by putting app on cliff-effects/app, docs on cliff-effects/docs

ethanbb:
OK, it's been a long journey. This is almost working now!

Puts the app itself at https://codeforboston.github.io/cliff-effects/app rather than just `/cliff-effects`. I did this so that we can put the docs at `/cliff-effects/docs` without the service worker interfering. This seems to be working.

I also made a static redirect page to live at `/cliff-effects[/index.html]`. This is working except, for some reason, `/cliff-effects/` (with a trailing slash) seems to trigger a Google search in Chrome rather than just loading the redirect page. Not sure how to fix this.

I changed the documentation package from `documentation` to `jsdoc` because `jsdoc` has prettier output for properties and probably has better support in general. `documentation` was somewhat nice in that the output folder had a single HTML file at the top level, but it still had plenty of assets and that shouldn't matter anyway in this system. Finally, I added a Node script to do the post-build filesystem manipulations (this required adding the plugin `fs-extra` to copy a directory recursively).

I'll be interested to see whether Netlify still works - it might need an update to its config. 

Let me know if this seems reasonable!


ethanbb:
Netlify seems to still work! 🎉

knod:
This is looking awesome and deploys just fine for me! One last shot at avoid /app - @wacii : Have any ideas about how to build without needing those two separate directories?

ethanbb:
Another alternative could be to use another repository called `cliff-effects-docs`, also published to GitHub pages, that includes `cliff-effects` at master as a submodule. Its npm scripts could automatically update the submodule before generating the docs and deploying, and we could have a cron job or something to deploy the docs every 24 hours or whatever interval makes sense.

wacii:
@knod Not really. There's only so much control we have over deployments on Github Pages. And we are already pushing what it was meant for.

PR #750, opened 2018-09-14
Asks user for permission to save their data before sending it - fixes #744

### LINKED ISSUE #744: Add feedback option to not include data?
This is for the feedback form itself. It occurs to me that this is another kind of storing data. In my head, when the person selected to not store data, there'd be a pop-up with a warning that says, 'If you're trying to tell us about a problem, we'll have a harder time fixing it if we don't have this data.' What do folks think?

knod:
I know we haven't actually discussed it yet, but it seems pretty much required.

It may be worth abstracting `AskPermission` at some point.


knod:
@todo Abstract `AskPermission`.

Issue #752, opened 2018-09-17
Bring version numbers from en google doc into en.js and update snippet names accordingly


pporfilio:

knod:
This is great to finally have as an issue! And some PRs! Also, maybe press pause for a sec!

To put it in a place where others can find it, I'm repeating what I said in #755:
@jlangevin and I just started talking about ways to dynamically strip out version numbers when the objects are created so that we don't have to constantly change these. There are some drawbacks - mainly that, in its current incarnation in my head, it would be invisible to anyone unfamiliar with the code base.

Sorry for the roller coaster ride. I think it may still be worth pulling in those PRs since they're something that's actually done while this thing is just an idea.

knod:
I think everything has version numbers now!

PR #755, opened 2018-09-17
[WIP] Parker/issue #752/add version numbers to snippets visit page

pporfilio:


knod:
A way of tackling the problem just became apparent, btw. This wasn't something we were holding back or anything, so this PR was honest work.

knod:
@pporfilio : This has one approval, which means you can merge it whenever you want 👍

pporfilio:
I think this is no longer correct because we've switched to the i_snippetName naming scheme so I'm closing it.

PR #756, opened 2018-09-18
Localization report

jlangevin:
Here is the localization report.  A button has been added to the DevHud menu which opens the report in a modal.  Happy to make appropriate changes if folks want the button in a different place/a different means of accessing the report.

Due to the length of the report, it only compares one localization file to EN at a time.  Instructions on how to use the report can be found in the report, though it's pretty self-explanatory.


jlangevin:
Oh, also, I moved my function that gets the "key paths" from the localization object into /utils/objectHelper.js.  Open to suggestions if folks have a better name for that.

jlangevin:
Closing and will create a new pull request.

knod:
How about `collectLocalizationPaths.js`?

PR #759, opened 2018-09-19
[WIP] Redirect Layout Prototype

dylanesque:
-Created basic layout prototype for redirect page
-Refactored class names in HomePage.js to lessen ambiguity.


redirect.html
```diff
@@ -3,14 +3,19 @@
...
     <meta charset="utf-8">
     <title>Redirecting to Cliff Effects</title>
-    <link rel="canonical" href="./app"/>
-    <meta http-equiv="refresh" content="0;url='./app'" />
+    <link rel="stylesheet" href="redirect.css">
+    <!-- <link rel="canonical" href="./app"/>
+    <meta http-equiv="refresh" content="0;url='./app'" /> -->
   </head>
   <body>
+    <header>Standard Cliff Effects Header goes here</header>
```
knod: I'm not sure it's a great idea to try to do the standard header here. If we're not using React, then we'd have to build this part by hand and would have to change stuff here independently if we changed stuff in that header. It would be a pain to maintain. Maybe we could just do something unique to this page? A link to the github repo or something in case the redirect doesn't work? Or no link. Just not something we'll have to keep matching up by hand.

redirect.html
```diff
@@ -3,14 +3,19 @@
...
+    <div class="redirect-text">
     <h1>Redirecting...</h1>
     <a href="./app">Click here if you are not redirected.</a>
-    <script type="text/javascript">
+  </div>
+    <!-- <script type="text/javascript">
       window.location = './app';
-    </script>
+    </script> -->
+    <footer>Standard Cliff Effects Footer goes here</footer>
```
knod: Same as comment for header

src/containers/HomePage.js
```diff
@@ -23,7 +23,7 @@ class HomePage extends Component {
             </Link>
           </div>
 
-          <p>*This is a prototype and should not be used to make financial decisions.</p>
+          <p className="home-disclaimer">*This is a prototype and should not be used to make financial decisions.</p>
```
knod: Generally speaking, we try to keep each PR to one purpose. This looks good to me, but in future, it'd be helpful to keep this kind of stuff separate.

ethanbb:
Not sure this was ready to be merged. We want to wait for at least one other person to explicitly "approve" it. @knod reviewed it, but it looks like wasn't completely satisfied with everything yet based on the comments.

knod:
Yeah, we generally do wait until a PR's been approved by at least one person. Also, we might discuss not approving or merging WIP PRs until they're not WIP anymore.

knod:
This prevents redirects from happening automatically.

PR #760, opened 2018-09-19
Update Chinese localization

2001zhaozhao:
Update zh.js localization file to match current format.

### DIFF
--- a/src/localization/zh.js
+++ b/src/localization/zh.js
@@ -1,6 +1,5 @@
 export default {
 
-  version: '2',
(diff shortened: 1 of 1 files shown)


src/localization/zh.js
```diff
@@ -1,61 +1,61 @@
 export default {
 
-  version:  `0.1`,
+  version: '2',
```
pporfilio: We're not using file-level version numbers any more, so could you update this pull request and remove this line from the google doc?

pporfilio: Nevermind, you beat me to it : )

PR #766, opened 2018-09-19
Fix StackedAreaGraph not adjusting to future income and not redrawing

ethanbb:
This fixes a crash in `VerticalLine.js` caused by not taking the future income into account when calculating the x-axis range for the `StackedAreaGraph`. Basically I just changed it so that the axis maximum is equal to or greater than both the max specified in `PROGRAM_CHART_VALUES` and the input income. 

I also cleaned up the math a little, adjusting the minimum according to the timescale the same way as the maximum (which doesn't matter now since it's 0, but just in case), and I added `redraw: true` to the `stackedAreaProps`. The latter is because I noticed that the graph wouldn't update if just the income and not the xRange changed. I believe this is because only the plugin cares about `income` (not the chart itself), so it doesn't know to update if only `income` changes. Not totally confident about this, since the graph seems to update many times for any change, but at least for me the performance is still good.

Finally, if you enter a ridiculously high value, you will eventually get a crash, so we also need to pull in #731 for this to be totally bug-free.

### DIFF
--- a/src/forms/output/StackedAreaGraph.js
+++ b/src/forms/output/StackedAreaGraph.js
@@ -47,22 +47,22 @@ class StackedAreaGraph extends Component {
     const { client, timescale, activePrograms } = this.props;
     const multiplier = multipliers[ timescale ];
 
-    // Adjust to time-interval, round to hundreds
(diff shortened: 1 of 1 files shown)


wacii:
I would have to spend more time with this to figure out the redraw part of it, but otherwise it looks fine.

PR #768, opened 2018-09-19
Increase timeout for Jest on big regression test

ethanbb:
Making this PR now so I can see how this works in Travis.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,0 @@
-{
-  "presets": ["es2015", "react"]
-}
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Instructions:
 8. Press cltr+c to stop the local server that you have running.
 
 <!-- Further discussion needed. See outdated comments for merge #154 (near the bottom of them). -->
-To get yourself set up to keep up to date with the latest changes to the `dev` branch, make sure you've gone through [Step 3 of the github guide](https://help.github.com/articles/fork-a-repo/#step-3-configure-git-to-sync-your-fork-with-the-original-spoon-knife-repository). To keep your branches from getting tangled as you both make changes and keep up to date with our repo, take a look at our wiki page about a possible [git workflow](https://github.com/codeforboston/cliff-effects/wiki/Guide:-Git-Workflow).
+To get yourself set up to keep up to date with the latest changes to the `dev` branch, make sure you've gone through [Step 3 of the github guide](https://help.github.com/articles/fork-a-repo/#step-3-configure-git-to-sync-your-fork-with-the-original-spoon-knife-repository). To keep your branches from getting tangled as you both make changes and keep up to date with our repo, take a look at our wiki page about a possible [Github workflow](https://github.com/codeforboston/cliff-effects/wiki/Guide:-Github-Workflow).
 
 <!-- Further discussion needed. See outdated comments for merge #154 (near the bottom of them). -->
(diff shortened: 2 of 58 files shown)


knod:
I think this was the kind of thing I was talking about, but I'm no longer sure: https://travis-ci.org/codeforboston/cliff-effects/builds/430572084

knod:
If I recall correctly now, it wasn't about a time limit for the whole test suite, it was about having too long a pause between responses sent to Travis.

jlangevin:
Could this be the issue.  I pushed some code earlier and a test unrelated to my work failed.

https://api.travis-ci.org/v3/job/430648409/log.txt

"getSNAPBenefits() matches saved results

Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL."

ethanbb:
@jlangevin Yeah, that's the issue that I can replicate with a sleep call (if I don't override the timeout interval like I'm doing here). I think the best way to address when this happens is to identify tests that take longer and allow more time for just those tests (the 3rd argument to `test` is timeout in ms). We can't override the global timeout interval without ejecting from create-react-app, it seems.

ethanbb:
OK, I just looked at it too and I see what you're talking about - the test cases in `test/programs/federal/snap/test-cases.txt`, right? I think this is related to what Michelle was telling me yesterday about generating tests via `setupTests.js`. @knod, were you involved with writing this test, and if so (or not) do you think we ought to keep it this long, or break it up in some way?

ethanbb:
@knod Interesting. I'm confused about what happened with that build - there doesn't seem to be any output at all from the test script. Did that just happen randomly?

No output for 10 minutes seems like a pretty reasonable fail condition, but if that's also happened when the tests were actually running correctly, we should change it.

knod:
It may just be something that happens with Travis sometimes - maybe a connection issue - but I'm not sure. Giving Travis a longer time may help or it may not. Still, this branch has a fix for the Jest stuff, so maybe we can merge it anyway and have that fix available. It's shame we can't make it a global change :/

knod:
Let us know when it's not WIP anymore and we can do the approval thing!

ethanbb:
Hey just to be clear, this shouldn't be merged since the only change I made was artificially increase the time required for a test and tell Jest to allow a longer time for it. The real fix would be to find the tests that actually take a long time (like that SNAP regression test) and increase the time allowed for them. And we should write down how to do that in the future since it'll probably happen again with new tests.

knod:
So you mean chunk it like @jlangevin suggested?

ethanbb:
We could do that, or if it's easier in the short term and it's causing issues with PRs now, we could increase the time limit and put in an issue to refactor it later.

knod:
My vote is for the latter, since there's so much to do, but it might be a quick fix and I won't object if folks want to look into it. @ethanbb , do you want to take a look to see if it can be quick?

ethanbb:
OK, I can take a little time looking into how easy it would be to refactor the test before just giving it extra time.

ethanbb:
No longer WIP

wacii (approved):
Seeing jasmine referenced here is a little jarring, but it looks good. I forgot you could pass a timeout as a third argument.

turnerhayes:
I'm curious if you tried running tests with `--runInBand` passed? I had issues with Jest tests timing out with a project of mine in Travis, and passing that flag (or `-i` for short) resolved it. Don't know if that's the issue here, but figured I'd bring it up.

knod:
May be worth some research!

PR #770, opened 2018-09-19
[WIP} Header and Footer refactoring

dylanesque:
-Refactored inline styles out of Header and Footer components


src/components/Footer.js
```diff
@@ -10,10 +10,9 @@ import {
 
 const Footer = ({ snippets }) => {
   return (
-    <Segment
+    <Segment className="footer_segment"
```
knod: Just to let people know this is being discussed, we know this doesn't have to do with the redirect page and we're working out how to keep branches untangled.

ethanbb:
Same as #759, this was probably not ready to be merged.

PR #774, opened 2018-09-20
Localization report util function tests and jsdocs - #527

jlangevin:
This is a first stab at jsdocs.  Feedback welcome.  Once I have a better feel for the docs I'll add them to the localization report along with tests.

I also added two tests to cover missing cases for getKeyPathsArray().


src/test/utils/objectKeyPaths.test.js
```diff
@@ -30,6 +34,25 @@ describe('getKeyPathsArray()', () => {
       ],
     ]);
   });
+
+  test('prop w/ version should have version removed when stripVersions flag is set', () => {
```
knod: Not a blocker, but does this test that '_v' followed by non-digits or periods don't get stripped?

jlangevin:
@knod,  You raise an interesting point.  So getKeyPathsArray() is assuming we've standardized key names in camelCase_v{#.#}.  So I'm doing a simple key.split('_')[ 0 ] to separate the key name from it's version.  However, I just played around and could switch to using regex.  The following would only treat it as a version num if the key ended with _v{#.#}.  See a few comments in the matches output below.
```
const reg = /(.+)(?:_v(?:\d+\.*?\d*?))$|(.+)/;
const keys = [
  'foo',
  'foo_v',
  'foo_v1',
  'foo_v1.',
  'foo_v1.2',
  'foo_v12.34',
  'foo_v1.2lkjfsd',
  'foo_v1.2_lkjfsd',
  'foo_somethingElse',
  'foo_something_else',
  'foo_something_else_v',
  'foo_something_else_v1',
  'foo_something_else_v1.',
  'foo_something_else_v1.2',
  'foo_something_else_v12.32',
];
keys.forEach((key) => {
  let matches = reg.exec(key);
  let match = matches[ 1 ] || matches[ 2 ];
  console.log(key, match);
});

/*
**** Matches output ****
foo                       foo
foo_v                     foo_v
foo_v1                    foo
foo_v1.                   foo <== perhaps this should not be allowed?
foo_v1.2                  foo
foo_v12.34                foo
foo_v1.2lkjfsd            foo_v1.2lkjfsd <== perhaps something like key_1.0a should be allowed?
foo_v1.2_lkjfsd           foo_v1.2_lkjfsd
foo_somethingElse         foo_somethingElse
foo_something_else        foo_something_else
foo_something_else_v      foo_something_else_v
foo_something_else_v1     foo_something_else
foo_something_else_v1.    foo_something_else <== perhaps this should not be allowed?
foo_something_else_v1.2   foo_something_else
foo_something_else_v12.32 foo_something_else
*/
```

knod:
I suggest disallowing all those, but allowing things like `foo_value` (which you already do, but which isn't an example). I'm not sure anything with a decimal should be allowed.

knod:
Again, though, I don't think that should be blocking things. Is there any other reason to keep this at WIP status?

knod:
Hmm, maybe just count `_v\d+$` - no no alphabetical letters allowed.

knod:
That is, if someone did `foo_v1a` then that doesn't have a version number and you shouldn't treat it as if it does.

PR #775, opened 2018-09-21
Put redirect files into dir and copy them all after building

ethanbb:
This allows having a separate `redirect.css` and `redirect.js` for example, or something more complicated. Since putting everything inline in one file probably isn't great style, and there's already a `redirect.css`. Everything in the `redirect` top-level directory now gets copied into `build` after the main app is built.


knod:
Hmm, is there a reason netlify doesn't automatically trigger a redirect?

ethanbb:
@knod Yeah, currently the redirecting parts of the redirect page have been commented out by the recent merge.

knod:
Ah, it's confusing because it just happens on deployed versions.

knod:
Feel free to merge!

PR #779, opened 2018-09-21
Removes last minor version numbers from en.js

knod:


knod:
Will do! Though not here.

PR #783, opened 2018-09-22
Inline style Refactoring

dylanesque:
-Refactored inline styles out of PageContent.js and AboutContent.js pages

-Corrected mistake from earlier PR where inline styles were not removed from Header.js file.


src/components/Header.js
```diff
@@ -10,7 +10,6 @@ class Header extends React.Component {
       <Segment className="header_segment"
         inverted
         textAlign='center'
```
knod: Not sure it's a blocking change, but it might be worth checking if this kind of stuff could be abstracted too. Not sure if it does something complicated css-wise in semantic-ui-react.

src/components/AboutContent.js
```diff
@@ -10,9 +10,8 @@ const AboutContent = function ({ snippets }) {
   return (
     <div>
 
-      <Header
+      <Header className="ac-header"
         as='h1'
-        style={{ fontSize: '4em' }}>
```
knod: You've gotten rid of the closing angle bracket here, which is probably why the build is failing.

dylanesque: Good catch, just fixed that

src/components/Header.js
dylanesque: Good point, I was wondering the same thing myself. I'll have a look at the semantic-ui docs tomorrow and experiment, see if this can also be ported over the the stylesheet, Most of it seems pretty common sense.

dylanesque:
Ok, this conflict should be resolved now

Issue #784, opened 2018-09-22
Abstract benefit calculations so all chart things are getting them the same way


knod:

ethanbb:
When looking through PR #785, I had some misgivings about the way we're reusing the `future` client object for each hypothetical income when constructing the stacked area or line graph. I'll summarize what @knod and I were talking about here.

We recently realized that we had a bug in our calculation of SNAP benefits when Section 8 benefits also applied. (@knod correct me if I get the details wrong here.) The issue was that the Section 8 benefit change would change the client's housing expenses, which in turn affects the SNAP calculation. So we had to make sure to modify the `future` `rentShare` value after calculating Section 8, and then use that new value when calculating SNAP.

This made us think a little about how things could look if we eventually added a larger number of benefit programs, each of which could affect different pieces of the `future` client situation. The `applyAndPushBenefits` function introduced in #785 provides a systematic way for each benefit calculation to also mutate `future` after returning its result to reflect the updated situation. This works if we always calculate benefits in a canonical order (such that changes always precede the calculation of benefits that depend on them) and only read from `future` rather than `current` within each calculation.

My issue was with reusing `future` for each of several hypothetical incomes (the values on the x-axis) rather than resetting it to `current` each time. My concern is that, if benefit A is to be calculated before benefit B, while we may know that A does not need to know about the value of B or any side-effects of B, that does not necessarily imply that A does not depend on the value of something that B touches _before_ B modifies it. 

But that probably should be the case, and maybe I'm off-base about this. I was thinking that B could modify something general in `future` whose old value might have been relevant to A, and this could mess up the A calculation on the next iteration (when the modified value is still present in `future`). However, it should probably be the case that anything B touches is not relevant to and not even read by A (or anything else that comes before B), so it won't matter that the change is carried over. That wasn't obvious to me until just now, so maybe we should either express that clearly somewhere (on the wiki?) or somehow work that requirement into the code (without over-engineering it, hopefully).

Sorry for rambling on! I'm interested to hear other people's thoughts!

knod:
Closed by #785 unless someone else has some more notes. If I'm not mistaken, folks discussing this seemed to have come to the conclusion that the last paragraph is correct. Not sure how to express this clearly yet, so not sure what to put on the wiki :/

knod:
Still worth thinking about. Should we make a new issue? One including the idea to make some requirements in the code? It seems to me that any requirements that would be in the code would make things less maintainable, but maybe we can figure something out.

PR #785, opened 2018-09-22
Abstract calcs - #784

knod:
First attempt (really the third attempt...)

Still have to implement in other charts/output.

### DIFF
--- a/src/forms/output/BenefitsLineGraph.js
+++ b/src/forms/output/BenefitsLineGraph.js
@@ -13,7 +13,7 @@ import {
   formatLabel,
   formatBenefitLinesTitle,
 } from '../../utils/charts/chartFormatting';
-import { getDatasets } from '../../utils/charts/getChartData';
+import { getChartData } from '../../utils/charts/getChartData';
 
 // DATA
 // In future, graphs will control their own aspect ratio,
@@ -49,7 +49,7 @@ class BenefitsLineGraph extends Component {
 
     var xRange      = _.range(limits.min, max, interval),  // x-axis/income numbers
         extraProps  = { snap: { fill: false }, section8: { fill: false }},
-        datasets    = getDatasets(xRange, client, multiplier, activePrograms, extraProps);
+        datasets    = getChartData(xRange, multiplier, client, activePrograms, extraProps, extraProps);
 
     // If there's no data to show, don't show the table
     if (datasets.length === 0) {
--- a/src/forms/output/BenefitsLineGraph.js
+++ b/src/forms/output/BenefitsLineGraph.js
@@ -49,7 +49,7 @@ class BenefitsLineGraph extends Component {
 
     var xRange      = _.range(limits.min, max, interval),  // x-axis/income numbers
         extraProps  = { snap: { fill: false }, section8: { fill: false }},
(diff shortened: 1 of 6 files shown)


knod:
@wacii , @justinqtaylor , @ethanbb : This one's off of WIP status.

src/utils/charts/getChartData.js
```diff
@@ -2,175 +2,89 @@ import _ from 'lodash';
...
+  // Existing `name` prop would be changed to `label`
+
+  // Active benefits should include 'income' if want it
+
+  // Shorter name now that we've clearly expressed what it is
+  var benefits = activeBenefitsInOrder;
+
+  var benefitDatasets = [],
+      allData         = {},  // each active benefit will have data in here
+      clone           = _.cloneDeep(client),
```
ethanbb: Would it be better to re-clone the client separately on each iteration below, so that the benefit calculation starts afresh from the current situation for each income? Currently I don't think it makes a difference, but I could see a situation at some point where benefit A depends on the current value of benefit B (and how it affects expenses, etc.), so we wouldn't want benefit B's lingering value/side effects from a previous calculation to affect a second one. (And now my head hurts...)

src/forms/output/BenefitsLineGraph.js
```diff
@@ -49,7 +49,7 @@ class BenefitsLineGraph extends Component {
 
     var xRange      = _.range(limits.min, max, interval),  // x-axis/income numbers
         extraProps  = { snap: { fill: false }, section8: { fill: false }},
-        datasets    = getDatasets(xRange, client, multiplier, activePrograms, extraProps);
+        datasets    = getChartData(xRange, multiplier, client, activePrograms, extraProps, extraProps);
```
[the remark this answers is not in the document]
knod: Yup!

src/utils/charts/getChartData.js
ethanbb: Hm OK, I see what you're saying and I want to agree with your points about why mutating `future` values should be safe, but I'm not fully convinced. For calculating SNAP `future` benefits for example, only the `future` client is used, so changes to `future` values could potentially affect that calculation. At the same, time, we need it to be able to see the change to future `rentShare` from the Section 8 calculation. 

I think it would be better if there were a way to guarantee that each benefit calculation only uses the `current` client, except for changes to future values that were made in the process of the current hypothetical situation. Cloning `future` each time is one way to do that. Or another way, and this is a radical proposal, could be to drop the idea of parallel `current` and `future` clients, and instead replace `future` with just a structure of all the differences from `current` in the future situation - call it `delta`. If `delta` doesn't have a certain key, that means it's the same as `current` (i.e.  the delta is implicitly 0). Then it would be a much smaller structure, and we could completely clear it before starting the benefit calculation process for a new hypothetical income.

Anyway that's clearly beyond the scope of this PR, and this should be merged!

As for the point about changing `current` values as we move forward or backward through incomes - I guess that depends on whether the benefit you get at a certain income depends on your previous income (or income history in general). If so, that makes things more complicated for sure...we'd have to decide whether the graph is meant to be showing what would happen if your income suddenly changed to each future income, or how benefits would change if your income slowly increased or decreased over time. In the latter case I think doing what you're saying makes sense. We should consider what effect it would have, though.

knod:
Closes #784, unless someone thinks there's more to do.

PR #656, opened 2018-07-12
Upgrade to most recent react-scripts

zhujinxuan:
1. Upgrade react-scripts and babel for compatibility of node 10 
2. Enable eslint in travis test
3. Enable prettier as linter for markdown.

PR #789, opened 2018-09-23
Update react-scripts, remove unnecessary babel packages, audit fix

ethanbb:
Part 1 of splitting up PR #656. Here I upgraded `react-scripts` and `gh-pages` and removed some `babel`-related dependencies and config files that were unnecessary (they don't show up in a brand-new CRA project and it seems to run fine without them). I'm going to check to make sure the build/preview still work and resolve any conflicts.


ethanbb:
Also added `ajv` version 6 since npm kept complaining about an unmet peer dependency.

ethanbb:
I forgot to mention, it also includes significant changes in `package-lock.json` from running `npm audit fix` (i.e. upgrading packages with security issues). I believe the upgrade to `gh-pages` was from that as well (it must've been the only one of our top-level dependencies that required an upgrade).

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: node_js
 node_js:
-  - 8
+  - lts/* # latest LTS version
+  - node  # latest stable version
```
knod: Do we want to automatically use the latest node version? Instead of keeping the version steady?

knod:
Other than that question, it looks good to me! Though I definitely have no idea what side effects could happen.

.travis.yml
ethanbb: That was what I was thinking. After all, people are probably going to update their Node gradually as new versions come out, and you can't control which version is used on the developer side, right?

The node version is mostly relevant for the person who runs `npm run deploy`, correct?

knod: Well, there are node version managers, as far as controlling what developers use. Can it work to just be more diligent about staying up to date with node versions so we can check the changelog for each version to make sure nothing's going to get broken?

I'm not sure what all the node version can affect. Seeing how this is for Travis, it looks like this will affect Travis tests, and if there's a problem with that, it may be a failure that's a bit tough to track down.

ethanbb: OK, I know I currently just update my Node whenever there's an update available. I think I'm on the LTS track currently.

Should we add an `.nvmrc` to the project and tell people to install Node using `nvm`? Looking at this: https://github.com/creationix/nvm

ethanbb: Hm, it doesn't look like n supports something like `.nvmrc` (or maybe I just missed it). Both these tools also don't support Windows, although I'm not sure how many people that's a problem for.

Maybe the better thing to do is to manually advance our "official" version number when a new major LTS version comes out, to make sure it doesn't break anything, but otherwise just tell people to use the version we're on (Node 8 currently). I don't think updates other than major version changes are anything to worry about. Do you think we should also test the latest version in Travis (maybe allow that one to fail), or is that a waste of time?

knod: If we do stay diligent about updating node, then it'd be almost the same, no?

ethanbb: The same as what?

knod: Did not see your latest comment before posting mine!

knod: That was partially because of that Node 10 version, true. I will defer to your judgement, but I'd prefer us to increment node versions by choice, though diligently, instead of doing it automatically.

ethanbb: OK. To be clear, I wasn't trying to say we increment versions too slowly and we should do so automatically, I was pretty much assuming people would update on their own, gradually, and just wanted to make sure Travis would keep up. But I'm hearing that we want to make it more of a thing to stay on a certain version until we decide it's safe to update, so I'll revert it back to how it was.

knod: Oh, ok. Sorry about that. Maybe I'm just terrible and incrementing my version of node.

PR #790, opened 2018-09-25
Replace split with regex when stripping version num from key

jlangevin:
Updated with the regex we discussed the other day.

### DIFF
--- a/src/utils/objectKeyPaths.js
+++ b/src/utils/objectKeyPaths.js
@@ -72,8 +72,9 @@
  */
 const getKeyPathsArray = (obj, stripVersions, base = []) => {
   // Array to contain our keys paths (if any)
-  let pathsArr = [];
-  
+  let pathsArr = [],
+      versionRegex = /_v\d+$/;
+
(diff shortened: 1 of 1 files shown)


knod:
@ethanbb : Hmm, does this need to account for that? Snippets aren't supposed to have minor versions now.

ethanbb:
@knod, yup, that's why I suggested changing the test. Changing the regex would only be if we think we might use minor versions again in the future, which I doubt. On the other hand it's probably harmless. But let's just change the test.

knod:
@ethanbb: Ohh, my bad - I forgot that this report isn't a series of tests anymore and that it has its own tests.

jlangevin:
Thanks, good catch!  I've split that set of tests into two.  One ensures that decimal versions do not get stripped anymore.  Now it's passing again locally.

jlangevin:
So we had a discussion about version numbers and key_v#.# is not considered a valid version.  Only key_v# is valid now, hence the simple regex in place now.

PR #796, opened 2018-09-26
Refactoring Inline Styles

dylanesque:
-Refactored inline styles out of VisitPage.js and Predictions.js

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -48,6 +48,29 @@ body {
   padding: 8em 0em;
 }
 
+/* Predictions.js */
+.prediction-button {
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: 10px;
+  max-width: 400px;
+}
+
+.predictions-message {
+  text-align: center;
+}
+
+/* VisitPage.js styles */
+
+.visit-container {
+  padding: 42px 0;
+}
+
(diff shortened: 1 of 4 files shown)


src/index.css
```diff
@@ -48,6 +48,29 @@ body {
   padding: 8em 0em;
 }
 
+/* Predictions.js */
+.prediction-button {
```
ethanbb: Am I missing something or did `prediction-button` not get added as a class on the Button it's supposed to apply to?

dylanesque: Ok, snapshot has been updated this should hopefully be all good now?

knod:
As well as the button class, I think you'll also need to resolve the conflicts. Let us know if you need a hand. If you go it alone, remember to incorporate the changes dev made.

dylanesque:
I'll definitely need help untangling these, I wouldn't have touched predictions if I knew somebody else was currently making changes to it at the same time.

knod:
Considering the... interesting... code I've been producing lately, I'm a little reluctant to throw stuff around in this room. I'll take a look anyway, but if anyone else wants to take a stab at it, that would also be great.

knod:
Made a PR on your repo's branch. Do with it what you will!

PR #797, opened 2018-09-26
Removing deprecated code

jlangevin:


jlangevin:
This had already been merged before you told us about the other issue.  I was also told that removing that code and risking breaking other fields was okay since it would reveal other code that needed to be refactored.

knod:
Ok, got it. Yeah, if other folks were ok with it, that's fine. I will say that there isn't a need to merge just to reveal other code - the changes are in that branch and everyone call pull it and look for it there.

knod:
Which isn't your fault, I'm just talking to them _through_ you because that's the most effective way :P

wacii:
I did pull the changes and check manually. But without relevant tests or a central implementation to work on, bugs were bound to slip through.

knod:
Sorry for the snark, it wasn't your fault either, or the fault of anyone to do with this particular PR. When I wrote that I thought I was making a lighthearted joke, but it doesn't read back very well. The code needs some serious tests, and the architecture needs rethinking.

PR #799, opened 2018-09-26
Add/fix JSDoc comments for App.js

ethanbb:
Continuation of PR #736 - see there for previous discussion.


knod:
@ethanbb : Y'know, I don't remember anymore. If you want to remove it, or deprecate it or whatever, I'd be ok with that.

As for notation, I'm happy with whatever. If that's a convention we decide to adopt, though, we should put it in the wiki somewhere.

src/App.js
```diff
@@ -81,6 +125,14 @@ class App extends Component {
     });
   };  // End setDev()
 
+  /** Load an individual client's data.
+   * The client data to load comes from a text input field in the Dev HUD
```
knod: Maybe 'currently only comes from' or something.

src/App.js
```diff
@@ -102,6 +160,7 @@ class App extends Component {
     return classes;
   };  // End propsToClasses()
 
+  /** Render the App component. */
```
knod: It may be worth asking if this comment is needed. i.e. does it clarify anything that the name doesn't already tell us?

src/App.js
```diff
@@ -62,11 +92,25 @@ class App extends Component {
     };
   };  // End constructor()
 
+  /**
+   * Set the human language of the app (ie the language in which the UI will display text for users to read, NOT the coding language).
```
knod: This was a small typo from before - 'ie' should be 'i.e.'

src/App.js
ethanbb: Yeah, probably not. I'll get rid of it.

ethanbb:
OK, I got rid of `loadClient`, at least for now.

ethanbb:
Great! I'll merge then.

PR #802, opened 2018-09-26
Fixes all radio buttons - fixes #792

### LINKED ISSUE #792: "Current Benefits" step - can't change radios back to "No"
### Replicate:
1. From home page, click Get Started button
2. On Current Benefits step, change either radio (Do you have Section 8 Housing? or Do you have SNAP?") to "Yes". 
3. Now try to change radio back to "No".

### Result I'd Expect:
- Radio should change back to "No"

### Actual Result:
- Radio stays set to "Yes"

### Environment:
Operating system and version: OSX 10.13.6
Browser type and version: Tested in FF 62.0 and Chrome 69.0.3497.100

App version or commit id at the time: Latest commit  4e84bcb

knod:


src/forms/CurrentExpenses.js
```diff
@@ -109,43 +109,46 @@ const EarnedFrom = function ({ hasExpenses, CashFlowRow, label, propData }) {
...
-      electricity = current.nonHeatElectricity,
-      phone       = current.phone,
-      fuelAssist  = current.fuelAssistance;
+  let hasClimate     = current.climateControl,
+      hasElectricity = current.nonHeatElectricity,
+      hasPhone       = current.phone,
+      hasFuelAssist  = current.fuelAssistance;
 
   let setChecked = function (evnt, inputProps) {
+    // Why is `checked` not always true here? Not a radio?
```
ethanbb: Makes sense to me - you'd want to know if a checkbox got unchecked, right? But I guess a radio button only sends an event if it's the one that got selected (is this true?)

src/forms/ShowOnYes.js
```diff
@@ -27,7 +27,6 @@ class ShowOnYes extends React.Component {
   }
 
   handleChange = (evt, inputProps) => {
-
     if (inputProps.value === 'Yes') {
```
ethanbb: Instead of changing `value` in `ControlledRadioYesNo` to `checked`, could we keep the `handleChange` that you deleted (maybe changed to an arrow function) and change this line to check whether `inputProps.value` is true? Then I think we wouldn't need the special case in `convertForUpdate` to handle `checked`.

(I guess either way it's kind of a special case, but at least this way we'd only have to handle converting 'Yes' or 'No' to a boolean in one place.)

knod: I think what you're saying is to change `ControlledRadioYesNo` so that it converts the value of `value` instead of doing that in `ShowOnYes`. I thought I had tried that and run into a problem, but I'm having some trouble remembering the details right now.

Given that, I'll go ahead and try it and see what happens!

src/forms/CurrentExpenses.js
knod: I think it's true, but I think in semantic-ui-react `Radio` button are just a `Checkbox` under the hood. I understand how it's trying to mimic HTML, but I always found that annoying in HTML too...

src/forms/ShowOnYes.js
knod: Also, I may have thought that keeping `Radio`s behaving in the expected manner all the way through, even though I don't like it, was the way to go. Who can know what yesterday-knod was thinking!? No one, that's who.

knod:
Hmm, I'm open to suggestions and re-working of the code. My intention was to get rid of the checked as a value property and switch it over to always ending up as a boolean value of whether something is checked/selected or not. At the end, in `convertForUpdate()` I was using it to tell when a non-standard `value` was coming in and dealing with it there. I figured since everything was going to get funneled there, why not do it there instead of at the source of every radio button. Then something didn't get funneled through there and it became an issue. I'm not sure where along the line to attack the problem. Thoughts? If you want to re-work the code, I'm open to that.

ethanbb:
I guess the two best options I see are: either fix the `value` of radio buttons as soon as possible by putting back in that `handleChange` listener under `ControlledRadioYesNo` (and then using `value` everywhere else rather than `checked`), or remembering that radio buttons (and probably checkboxes, if we're going this route) are special and have to be handled differently from other inputs in things like `convertForUpdate`. Personally the first way seems a little easier and cleaner.

knod:
I've thought a bit and maybe part of the problem is that when `value` gets passed down and it's `true` or `false`, I've found it really confusing to remember what's going on when usually for radio buttons you'd be handing down the actual value - 'Yes' or 'No'. Because the client data doesn't find 'Yes' or 'No' useful, though, `true` and `false` is what we end up with. I can still switch to the old way, but does anyone have ideas for that? Do we change that value of `true`/`false` to 'Yes'/'No' as soon as possible (which would be similar to changing 'Y' and 'N' to `t` and `f` as soon as they come out of the gate) or do we let `t` and `f` funnel down all the way to the `Radio` itself and then change it (like letting everything filter to `convertForUpdate()`)?

knod:
The actual prop on `Radio` itself that decides whether it will be selected or not is called `checked`, btw.

knod:
So I switched it up so that `convertForUpdate()` is not handling the `value` change, the radio component is. I kept `checked` because the `<Radio>` itself uses the `checked` property and uses 'Yes' as the `value` value, so it seems more consistent to make that clear from the start, but I'd appreciate input on that if anyone has some.

I also changed the radio's `updateClientValue()` to `onChange` since it's not always updating a client value. Didn't spot that before, but I think it's worth doing, though I'd love input on that one too.

What other changes does this need?

PR #805, opened 2018-09-28
Vietnamese translation - #513, #500

hmtranae:
Vietnamese translation of en.js; not familiar with JSON format but all of the objects keys are surrounded with single quotes which is not the case in en.js. Would this cause a problem?

### DIFF
--- a/src/localization/all.js
+++ b/src/localization/all.js
@@ -2,11 +2,13 @@
 
 import DE from './de';
 import EN from './en';
+import VI from './vi';
 import ZH from './zh';
 
 var localizations = {
   de: DE,
   en: EN,
+  vi: VI,
   zh: ZH,
 };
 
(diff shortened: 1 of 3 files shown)


knod:
It's strange that it copied that way, but it won't break anything. I'll have to check out the google en.js doc.

knod:
Hmm, forgot we had the localization checker. There seem to be a couple other things amiss, but I think we can take care of those too just fine. Do what you will!

knod:
Ha! I think it's something wrong with en.js! Eh, let's go with this and we'll see what we see when everything _else_ is up to snuff.

knod:
It looks like you've got it covered, thanks! A PR maker needs one approval to be able to merge, so after the tests pass, feel free to merge anytime!

hmtranae:
Okay great, thanks!

PR #806, opened 2018-09-29
Implements 'Overview' (you can see it in the wiki too)

knod:
Hasn't been tested with a deploy since my repo name is weird. I'll see it in the preview!

I'd love everyone's input first, but I think it would be good to merge this soon and tweak it after.


OVERVIEW.md
```diff
@@ -0,0 +1,124 @@
...
+
+The tool is not meant to be used to make financial decisions. We didn't have access to SMEs (**subject matter experts**), so we couldn't double-check our logic and could not fully account for all regulations. There also wasn't a way to fully test the tool for accuracy.
+
+
+
+## A. Input/Output Flow
+
+### Summary
+
+The client puts in their information. Valid answers ([see the definition of valid input](#d1-valid-input)) is processed into the [`client`](#d2-the-client-object) object. Calculations are made, and the result of the calculations are displayed.
```
ethanbb: Should be "are," not "is"

OVERVIEW.md
```diff
@@ -0,0 +1,124 @@
...
+- User-friendly icons and colors to more clearly visually represent what's going on.
+
+
+
+
+## G. Code and Code Cautions
+
+### G1. WARNING: Values the user shouldn't change
+
+The options for what circumstances the user can try to predict can be expanded. That said, never allow users to change certain kinds of future values. For example, a value like `future.rentShare` - they're not in charge of their subsidy. They can change `future.contractRent`, which is out of their control, but still something that the calculations aren't changing themselves.
```
ethanbb: This is a little confusing - I think I know what you mean, but you could elaborate on why some things shouldn't be inputs. Inputting `rentShare` wouldn't make sense because that's what the app is trying to calculate, right? So a basic rule of thumb would be that a value can't be both an input and an output.

On the other hand, what if the client knew for sure what their future rent share would be, but wasn't sure about some of their other benefits? Could we have a way to override what the tool would normally output for a benefit, to affect the calculation of other benefits? Maybe that's too hypothetical a situation though.

ethanbb:
Just looked at it on the docs landing page - looks great!

knod:
It's not replacing the readme. That's why I called it 'overview' instead.

knod:
I guess it could replace the readme if we put what's in the readme at the top.

OVERVIEW.md
knod: People do put in their current rent share - it's the only way we can have a chance at calculating their future rent share.

I do think that it should be possible for a client to put in exceptions and such because those do happen, but I do agree that that's a hypothetical right now.

knod: Changed the text a bit. Let me know what you think.

ethanbb:
Sorry but currently it looks like it *is* replacing the readme - at least, I see changes in both OVERVIEW.md and README.md on Github.

knod:
@ethanbb : There are changes in both, but the change in the README.md should just be to add a link to the overview.

knod:
Wait... wat?!

knod:
Ok, _that_ should be better. Weird.

OVERVIEW.md
ethanbb: ^ right sorry, I meant future rent share.

knod: So the reason I put this section in is because it the rent subsidy is sort of an input and an output. It's an input for current values, but can't be an input for future values. Did that come through in the new text?

knod: Hmm, actually maybe I should put it more like that in the text of the overview...

ethanbb: Right, I agree it's confusing and should be explained. Didn't mean to imply that it's obvious. I was just saying we should explain here that you can't put something as an input if it's already a calculated result from the app.

knod: I'll play around with the wording a bit more.

knod:
Right now, working from these comments (mine and others'):
> the rent subsidy is sort of an input and an output. It's an input for current values, but can't be an input for future values.

> you can't put something as an input if it's already a calculated result from the app.

ethanbb:
OK, I think the new version is very clear, thanks!

Issue #616, opened 2018-06-25
'New Client' button in 'Predictions' section to orange color


knod:
May be better to just make a class for 'dangerous' and give that class to both 'New Client' buttons, then style in css.

knod:
Fixed by a `ButtonReset` PR recently

PR #807, opened 2018-09-29
One idea for a different abstraction for buttons - more clear? #616

knod:
The `...moreProps` thing is about possibly handing down class names or something. Premature feature? Other thoughts?


knod:
👍

ethanbb:
Agreed about the other being better. Also, I think we should only factor out the reset button if we do also reuse it for the side button (I'm sure we should be able to, though).

knod:
This closed in favor of #808

PR #808, opened 2018-09-30
Another, similar, option for buttons - only one button file

knod:
The file is left abstracted so that it can be used for the fixed 'New Client' button on the left.


knod:
I have your approvals, so I'm going to merge, but I will put it out there that there may be other solutions too and if we find something better that'd be good too.

PR #810, opened 2018-09-30
[WIP] Inline Styles

dylanesque:
-Refactored inline styles out of Predictions.js

### DIFF
--- a/src/forms/Predictions.js
+++ b/src/forms/Predictions.js
@@ -57,7 +57,7 @@ const TabbedVisualizations = ({ client, snippets }) => {
     <Tab
       menu={{ color: 'teal',  attached: true, tabular: true }}
       panes={ [
-        { 
+        {
           menuItem: (
             <Menu.Item key="tab1">
               { snippets.i_tabTitleChanges }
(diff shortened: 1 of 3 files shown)


knod:
Feedback wanted, or not yet?

dylanesque:
Sorry, yes, just a review would be good.

knod:
The first thing that jumps out at me is that the 'submit feedback' button isn't centered. I'm not sure how much to worry about that, though, because that button's going to be a bit different after the 'text result' changes.

dylanesque:
> The first thing that jumps out at me is that the 'submit feedback' button isn't centered. I'm not sure how much to worry about that, though, because that button's going to be a bit different after the 'text result' changes.

It wouldn't be, because the display is set to block. I can still change that to inline-block, or experiment so that it's centered, for the time being?

knod:
I wouldn't be against it! It's good to always have a working version as much as is possible, even when making a transition.

dylanesque:
> I wouldn't be against it! It's good to always have a working version as much as is possible, even when making a transition.

Ok, this has been adjusted to center that button.

knod:
Seems to work for me! Let us know when it's not WIP anymore!

PR #811, opened 2018-09-30
Adds some tab-ability and visual feedback - starts #816

knod:
It will definitely need some conflicts resolved.

Problems that remain:
- Income inputs are waaaaaay too much to tab through. We need to figure out something with arrow buttons... somehow... Really not sure how. Maybe we need to break them up into different sections and be able to skip to the next section somehow? :/
- Spouse/member drop down sometimes doesn't drop
- Spouse/member drop down not clear which choice is currently being chosen
- Spouse/member drop down sometimes doesn't switch when it should.
- Still not enough contrast in app in general

Probably other things too. For example, I'm not sure you can press arrow buttons to navigate to different graph tabs.

### DIFF
--- a/src/components/details.js
+++ b/src/components/details.js
@@ -29,6 +29,14 @@ class HeadingWithDetail extends Component {
     });
   };
 
+  // For keyboard access
+  // 'Icon' won't take `onKeyDown` event handler
+  onKeyDown = (evnt) => {
+    if (evnt.key === `Enter`) {
+      this.toggleDetails();
+    }
+  };
+
   render () {
     var { showDetails } = this.state;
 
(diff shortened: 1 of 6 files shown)


ethanbb:
Oh yeah this focus style is a lot more noticeable than the browser default! This is a good starting point.

Couple things I noticed:
* The buttons on the home page seem to get tabbed over twice.
* When a modal is open, things underneath still get tabbed over (not sure if that's fixable).
* The focus style isn't working within modals.

knod:
@ethanbb : Can we save those improvements for a later PR?

ethanbb:
@knod Just noticed that this also adds the extra highlighting when just clicking on buttons and other widgets. Not sure I like that - was that intentional? Do you think we could/should have it show up only when using the keyboard?

knod:
@ethanbb : From what I've researched, I haven't found a way to do that. There's no different selector when focus happens with the mouse vs. the keyboard.

knod:
Does that change this choice?

ethanbb:
No, I think the keyboard accessibility takes priority. So if it's not possible to separate the two, I'm OK with how it is.

PR #817, opened 2018-10-02
Adds HUD off button in HUD itself. Got too long, made some changes

knod:
Pros - you won't notice it if you keep your window tall enough
Cons - you have to refresh your window if you change window
heights to be able to see the hud at the right size/to see all
the columns :/

### DIFF
--- a/src/components/dev/DevHud.js
+++ b/src/components/dev/DevHud.js
@@ -4,6 +4,7 @@ import {
   Checkbox,
   Button,
   Dropdown,
+  Header,
 } from 'semantic-ui-react';
 
 // PROJECT COMPONENTS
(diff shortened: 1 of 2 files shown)


ethanbb (changes requested):
Hm, the whole HUD seems to get wider and go off the page when I click the question mark sometimes. Also, the second checkbox seems like it's not working (always stays checked). I like the appearance tweaks, though.

jlangevin (commented):
I'm seeing what Ethan is with the width, but all the time FF (62.0.2).  As for the height, perhaps make the dev-menu have a max height of the viewport minus 30px for the close button and make it scroll y if taller than the screen?

knod:
@jlangevin , @ethanbb : The weirdness with the width and not adjusting and such is because of my attempts to flex column and wrap at the same time. It seems flexboxes won't take on the combined width of their children in that case. It's apparently a bug in all browsers. The examples linked here seem to work: https://github.com/philipwalton/flexbugs#flexbug-14. Unfortunately, `position: fixed` still seems to interfere with that in some ways and my attempts to fix it have been fruitless. The HUD will change to the right size, but you have to refresh the page to get it to behave again, which isn't useful. For some reason, that codepen doesn't work that way.

As for scroll y, currently you can click through the body of the HUD to the stuff behind it, which was intentional. The only way to make it scrollable is to take away that feature. That might be worth it, though, especially considering you can minimize the HUD. Thoughts?

@ethanbb : I'll examine that second checkbox more closely.

@jlangevin : Hmm, any reason to not have the `max-height: 100vh`? Why minus 30px? Maybe you're talking about the `.dev-menu`, but putting the height on the `dev-hud` is, I think, just as effective. That isn't what's causing the weird behavior.

knod:
Sounds good! I'll work on that change.

knod:
Decided not to untangle this mess and just do the more simple thing from scratch: #830

Issue #821, opened 2018-10-02
Clean out useless jsdoc comments


knod:
In:
- [x] /components (knod)
- [x] /forms (knod)
- [x] /data
- [x] /programs
- [x] /utils
- [x] Any others

[knod, 10/23/18: I think the are done at this point, though I'm not sure. I'm pretty sure /utils hasn't been worked on.]

knod:
I think this may have been completed, or mostly so, in subsequent cleanups. We'll keep our eye out for more and keep working on updating the docs in the previous issue.

PR #823, opened 2018-10-02
Cleans up docs a bit for Benefits section. Not comprehensive. #821

knod:
And not meant to be comprehensive.


jlangevin (commented):
Is there a reason this exports LocalizedRadioYesNo and CurrentBenefitsContent?

knod:
@jlangevin : So they can be tested more easily without making a bigger overhaul to the code.

knod:
You're right that that's not quite related to docs, now that I think on it. That's more of a testing thing. I can remove that change.

dylanesque (approved):

PR #824, opened 2018-10-03
Specifies node version

alvincrespo:
## Changes

- Adds a `.node-version` file to ensure consistent node version among devs

## Developer Experience

If you're using [nvm](https://github.com/creationix/nvm), you can run:

```
nvm use $(cat .node-version)
```

This will set the node version to be used with this project.

If you do not have this version installed, you can run:

```
nvm install $(cat .node-version)
```


alvincrespo:
Not sure if this is the version of node we want to stabilize on - but thought I'd create a PR 😄

ethanbb (commented):
I think it might make sense to use the LTS version of Node instead of the latest, unless there's something we need in Node 10. No strong opinion from me though.

knod:
Thanks for the PR, @alvincrespo , it's probably a good idea! I agree about the latest stable version as opposed to just the latest version.

Also, just curious, but with a .node-version file present, isn't the command just `nvm use`? I've never seen the `$(...)` syntax before, so maybe I'm misreading something.

knod:
By latest stable version, I do mean an actual version number, not just a general 'LTS' command. Is that what you meant, @ethanbb ?

ethanbb:
Yup.

turnerhayes:
We might consider using an [.nvmrc](https://github.com/creationix/nvm#nvmrc) file so that we don't have to `cat` files

knod:
That looks like an attractive option.

turnerhayes:
Can we close this due to #979 ?

knod:
Closed due to #979.

PR #825, opened 2018-10-03
Adding developer docs to centralize info

alvincrespo:
Rendered Docs:

- [README](https://github.com/alvincrespo/cliff-effects/tree/ac-docs)
- [Installation and Setup](https://github.com/alvincrespo/cliff-effects/blob/ac-docs/dev-docs/INSTALLATION-AND-SETUP.md)

## Changes

- Creates a dev-docs directory to consolidate technical information around the project (e.g. installation, code styling, testing, etc...)

### DIFF
--- a/dev-docs/INSTALLATION-AND-SETUP.md
+++ b/dev-docs/INSTALLATION-AND-SETUP.md
@@ -1,0 +1,62 @@
+# Installation and Setup
+
+## Preparing Yourself
+
+To follow these instructions, it'd help if you:
+
+- have mentally prepared for experimentation and learning and/or
+- know how to open your terminal/command prompt
+- know a bit about how to navigate to different folders in your terminal
+- know a bit about installing software for your terminal, like nodejs
+
+## Github Setup
+
+If you're new to github look at [How to: fork a repo](https://help.github.com/articles/fork-a-repo/). Otherwise, fork and clone the repo, creating a link (remote branch) to the Code for Boston repo.
+
+In your termainal/command prompt, do `git checkout dev`.
+
+_Do not touch the `master` branch._
+
+## Dependency Installation
+
+### [NVM](https://github.com/creationix/nvm)
+
+```
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
+```
+
+### [Node/NPM](https://nodejs.org/en/)
+
+```
+nvm install $(cat .node-version)
+```
+
+## Project Dependency Installation
+
+First, make sure you're using the correct node version
+
+```
+nvm use $(cat .node-version)
+```
+
+Now, run:
+
+```
+npm i
+```
+
+## Setup
+
+Running the app is as easy as:
+
+```
+npm start
+```
+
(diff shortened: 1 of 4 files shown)


dev-docs/INSTALLATION-AND-SETUP.md
```diff
@@ -0,0 +1,62 @@
...
+In your termainal/command prompt, do `git checkout dev`.
+
+_Do not touch the `master` branch._
+
+## Dependency Installation
+
+### [NVM](https://github.com/creationix/nvm)
+
+```
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
```
alvincrespo: I added nvm here since I've found this to be extremely helpful to ensure that dev setup is consistent across machines. More than happy to hear what everyone thinks.

knod:
These are some awesome additions! Is it possible we could add these to the docs, but also keep them in the README? I've noticed people have a hard enough time actually reading these instructions when they're in the README, but it does give them a chance to spot the node version (or some other useful thing) out of the corner of their eye. If they have to click on a link before they see the text, they're much less likely to spot anything.

The other thing is that I can't see them in the netlify preview deploy docs. To see that you can click on 'Show all checks' next to 'All checks have passed', and then click on 'Details' next to 'netlify'. I wanted to check in with @ethanbb to check on how to get jsdoc to include these. I think they'd be considered 'guides' in there, but I've never tried to add guide documents.

knod:
If you do want to bring back the README, one way to avoid a merge conflict is to just look at the README in the current `dev` branch, delete the text in your README, and copy and paste the `dev` branch stuff into your README file.

ethanbb:
@knod, hm not sure, but it looks like you can use `-u` or `--tutorials` to indicate a directory with tutorials. Maybe that's what we're looking for?

(For the record, I'm not an expert on `jsdoc`, I just have it installed globally and am running `jsdoc --help`)

dev-docs/INSTALLATION-AND-SETUP.md
jlangevin: I think this is a great idea.  I'd add the wget command as well, just in case they don't have curl installed.  

```
To install or update nvm, you can use the install script using cURL:

curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash

or Wget:

wget -qO- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
```

knod:
@ethanbb : I guess these do kind of fall into the category of tutorials. I'd seen something called 'guide', though, and these might fit better there - they don't go into as much detail as I'd expect from a tutorial. Not sure though. Or maybe 'guide' is just someone's other way of saying 'tutorial'.

ethanbb:
OK, interesting. Do you remember where you saw this thing about "guide"? I don't see anything like that built into `jsdoc`, but maybe it's a plugin or part of a different documentation parser.

knod:
@ethanbb : Things like https://docs.videojs.com/

knod:
Haven't heard from this person. We'll deal with this issue in a separate place.

PR #826, opened 2018-10-03
Adds linting precommit hook

alvincrespo:
## Changes

- Adds [husky](https://www.npmjs.com/package/husky), [prettier](https://prettier.io/) and [lint-staged](https://github.com/okonet/lint-staged) to ensure consistent code styling upon commit

## Developer Experience

When you commit, you should see something like:

```
➜  cliff-effects git:(ac-lint-staged) ✗ git commit
husky > pre-commit (node v10.11.0)
 ✔ Running tasks for *.{js,jsx}
[ac-lint-staged 93caa19] Adds linting precommit hook
 3 files changed, 1309 insertions(+), 244 deletions(-)
 rewrite src/App.js (62%)
```

On success, files will be linted and auto-formatted.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,15 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@samverschueren/stream-to-observable": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/@samverschueren/stream-to-observable/-/stream-to-observable-0.3.0.tgz",
+      "integrity": "sha512-MI4Xx6LHs4Webyvi6EbspgyAb4D2Q2VtnCQ1blOJcoLS6mVa8lNN2rkIy1CVxfTUpoyIbCTkXES1rLXztFD1lg==",
+      "dev": true,
+      "requires": {
+        "any-observable": "^0.3.0"
+      }
+    },
     "@types/node": {
       "version": "10.0.10",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-10.0.10.tgz",
@@ -150,6 +159,12 @@
           }
         }
       }
+    },
+    "any-observable": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/any-observable/-/any-observable-0.3.0.tgz",
+      "integrity": "sha512-/FQM1EDkTsf63Ub2C6O7GuYFDsSXUwsaZDurV0np41ocwq0jthUAYCmhBX9f+KwlaCgIuWyr/4WlUQUBfKfZog==",
+      "dev": true
     },
     "anymatch": {
       "version": "1.3.2",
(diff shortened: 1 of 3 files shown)


ethanbb (commented):
So, using prettier almost definitely conflicts with our current usage of eslint. That being said, *I'm* not 100%  opposed to prettier, but we should choose one or the other (for code style issues, at least).

knod:
I'll admit I'm partial to the code styles we've established so far. I know other folks feel differently, though, and would love to hear where other folks are at.

It might be good to keep in mind that even if we do use prettier, it won't take away the need to make coding style choices that prettier doesn't cover, or choices that no linter _can_ cover, so it's not a complete solution whichever way we go.

dylanesque (dismissed):
I'm definitely in favor of using Prettier or something that auto-formats more strictly, though we should probably discuss this. But I don't feel like eslint has been adequate for consistency, I've had to resolve some merge conflicts due to whitespace.

knod (changes requested):
Sorry, I'm going to request changes to give the current conversation a chance to play itself out.

jlangevin:
I agree this needs more discussion before any switch.  My $0.02 is that many of these changes are fine, but some I'd argue against, and some for.

Prettier forces double quotes - I have to say I'm pretty strongly on the no side of this argument.

The only other place where I really don't care for Prettier is removal of too much white space.  I find it MUCH easier to read long lists of key value pairs when they're lined up.  I know Google argues for no extra spacing, but what the heck do they know!?  ;-)

```
Current:
devProps: {
  dev:        false,
  devHidden:  false,
  english:    true,
}

Prettier:
devProps: {
  dev: false,
  devHidden: false,
  english: true,
}
```

Again, too much white space removal.  Code should be written for humans imho.  Computers do just fine ignoring white space.  I prefer line wrapping when destructuring vars to be required (or optional).  I used to put long lists on separate lines (find it easier to read) and short lists on the same line.  If I had to choose just one, definitely separate lines.  This gets really ugly when you have lots of vars.
```
Current:
var {
  langCode,
  snippets,
  devProps,
  clients,
} = this.state;

Prettier
var { langCode, snippets, devProps, clients } = this.state;
```

And below adds inconsistency with space at the opening { but removes it on the closing }
```
Current:  return { clients: { ...clients, loaded: newData }};
Prettier:  return { clients: { ...clients, loaded: newData } };
```

Okay with either:
```
Current:  if (obj[ key ] === true) {
Prettier:  if (obj[key] === true) {
```

Arrow functions technically do not require ( ) when there is only one parameter.  So I prefer to leave them out for less typing, but okay with current rules since always having them makes for consistency.
```
Current:
this.setState((prevState) => {
     ...
}

Prettier:
this.setState(prevState => {
      ...
}
```
Current rules simply don't force either of these.  I generally do not put a space.
```
propsToClasses (obj) {}
propsToClasses(obj) {}
```

jlangevin:
Okay, that might have been more like $0.20...

ethanbb:
Maybe a good intermediate step would be to set up a hook to automatically run `eslint --fix` when pushing (or making a PR? however prettier does it) since that generally fixes all the style issues according to the rules we've decided on after that long discussion. There were some weird edge cases like ternary operators without parentheses where rules conflicted and eslint couldn't do an automatic fix, but I think we've figured out how to avoid those at this point.

If people want to use prettier because they prefer the style to our current style, that's another issue. I've seen the argument that prettier ends these discussions by taking away everyone's control over the style, but that's kind of a moot point if we first have to decide whether we like its style well enough to switch. So currently, since our `eslint` style has been fixed for a while and was based on people's preferences, the most conflict-free and time-efficient route would be to stick with what we've got, at least in my view. I agree with a lot of @jlangevin's points about prettier's style. But if enough people really don't like our style, as in a majority of everyone (including people who don't care), then maybe we should switch.

knod:
Yeah, if people want to discuss switching we can do that.

knod:
It sounds like folks are most interested in an automated lint fix before PRs. We'll work on that instead. Thanks for the push in the right direction!

Issue #816, opened 2018-10-02
Some accessibility items


knod:
Some accessibility items that need addressing:
- [ ] More contrast needed for hover and other interactions, including:
   - [ ] General nav bar
   - [ ] Form section nav bar
   - [ ] Spouse vs. other member drop down in 'Household' form section
- [ ] More visible `:focus` styling.
- [ ] Income inputs are waaaaaay too much to tab through. We need to figure out something with arrow buttons... somehow... Really not sure how. Maybe we need to break them up into different sections and be able to skip to the next section somehow? :/
- [ ] Spouse/member drop down sometimes doesn't drop when user navigates to it.
- [ ] Spouse/member drop down chosen item sometimes doesn't switch when it should.
- [ ] The buttons on the home page seem to get tabbed over twice.
- [ ] When a modal is open, things underneath still get tabbed over (not sure if that's fixable).

knod:
I'm doing some work on the focus visibility part. Either this issue can be for multiple people, or we can break it into individual issues (possibly as folks come along and are interested in moving it forward).

dylanesque:
I'll start looking for contrast issues and knock those out.

knod:
Do labels need to have a `for` attribute that'll connect them to their ui element in order to be read properly by a screen reader?

dylanesque:
> Do labels need to have a `for` attribute that'll connect them to their ui element in order to be read properly by a screen reader?

**"Whenever possible, use the label element to associate text with form elements explicitly. The for attribute of the label must exactly match the id of the form control."**, from the W3C Accessibility Tutorial, so yes, we absolutely should be using the ```for``` attribute.

PR #829, opened 2018-10-04
Focus styles on the feedback form work, styles tweaked - #816

knod:
tweaked to put a single border around form elements and double borders
around non-form elements. Should the bottom buttons should be
taken out of the form?

Another question is do 'Cancel' and 'Submit' buttons on modals count as form elements?

The difference in style is because things like check boxes and radio buttons get overwhelmed if they have the double border. Should we just not have a double border on anything? Should we have a double border around everything except check boxes and radio buttons?

What this doesn't do is fix the fact that stuff behind the modal still gets focus while the modal is open. It's a bit of a puzzler. First off, I haven't found a way to do it with CSS - I just can't access previous siblings or anything like that. That means that we'd have to manage it from the previous sibling side of things, which would be kind of messy. Possibilities are a class and `user-select: none;` CSS rule, an attribute of 'unselectible' or whatever (for Safari?), or disabling the inputs somehow (does disabling a parent disable everything in it?).

Anywho, them's the breaks. Not sure what we want to do about that.

### DIFF
--- a/src/components/prompts/FeedbackForm.js
+++ b/src/components/prompts/FeedbackForm.js
@@ -176,6 +176,8 @@ class FeedbackPrompt extends React.Component {
       };
     };
 
+    // Without the #root selector, we can't get the accessibility
+    // focus styles to win 
     return (
       <Modal
         mountNode = { document.getElementById(`root`) }
@@ -227,6 +229,7 @@ class FeedbackPrompt extends React.Component {
         </Modal.Actions>
 
         <Modal
+          mountNode = { document.getElementById(`root`) }
           size               = { `large` }
           open               = { this.state.ready }
           onClose            = { this.closeAskPermission }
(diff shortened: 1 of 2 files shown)


knod:
Those questions don't have to be answered for this PR, they're just @todo questions.

ethanbb:
Interesting. So the `#root` helps the selector take precedence because it's an id selector, even though everything is under the root? I don't want to officially review this since it's CSS, but looks pretty safe.

One thing I noticed is that the focusing style on the home page is back to the default - was that intentional? I think it would be really nice if we could get the same big color change on keyboard focus as we have on mouseover currently, instead of the black outlines like everywhere else.

knod:
@ethanbb : Hmm, I'll check out home page. I will say that the styles for button colors in general don't respond to keyboard input. I think that may be for a different PR. I think all those styles are on hover, etc.

knod:
@ethanbb : Does it make sense for that to be a different PR?

ethanbb:
Yeah they could be. I wasn't trying to request changes for this, just informally giving some suggestions.

knod:
I've already prepped a PR that takes care of that for the home page :)

dylanesque (approved):
Looks good, every time you remove the universal selector from a codebase, an angel gets it's wings

knod:
lol

PR #830, opened 2018-10-04
Adds size flexibility and scrollability to hud menu

knod:
Based off of comments and discoveries from #817

### DIFF
--- a/src/components/dev/DevHud.js
+++ b/src/components/dev/DevHud.js
@@ -46,7 +46,7 @@ const DevMenu = function ({ devProps, funcs, data, state }) {
         </Button>
       </Menu.Item>
 
-      <Menu.Item header>> Snippets</Menu.Item>
+      <Menu.Item header> Snippets</Menu.Item>
       <Menu.Item>
         <Checkbox
           label    = { `Mark English snippets` }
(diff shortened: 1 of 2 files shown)


dylanesque:
Where's the toggle for the dev hud? I'm not seeing an obvious one in the app.

knod:
@dylanesque : It's hidden so it won't be accessed by non-devs. Go to /dev (or /app/dev on the deployed version).

dylanesque:
> @dylanesque : It's hidden so it won't be accessed by non-devs. Go to /dev (or /app/dev on the deployed version).

Got it. I had been thinking on and off if there was a simple way to ditch the fixed positioning and get the same effect, but all that I'm coming up with at the moment is "make its parent a Grid container", which is like using a battle-ax to kill an ant in this instance.

knod:
Thanks for giving it some thought. I also haven't been able to come up with another solution, especially if we want it to be flexible.

The tests look like they're pending in the github page, but when I go to the details, Travis says they've passed, so I'm going to merge.

@ethanbb and @jlangevin : Thanks for the input. If this doesn't match up with what you were thinking, let me know. It should be pretty easy to undo/tweak these changes.

ethanbb:
@knod Sorry didn't get to this, but it looks good!

PR #832, opened 2018-10-04
Cleans docs a bit for Expenses section. Not comprehensive. #821

knod:


src/forms/CurrentExpenses.js
```diff
@@ -243,19 +243,16 @@ const HousingRadio = function ({ currentValue, label, time, updateClientValue })
...
  * @function
  * @param {object} props
- * @param {object} props.current - Client data of current user circumstances
- * @param {string} props.type - 'expense' or 'income', etc., for classes
- * @param {string} props.time - 'current' or 'future'
- * @param {function} props.updateClientValue - Sets state values
+ * @param {object} props.current Client data of current user circumstances
+ * @param {string} props.type 'expense' or 'income', etc., for classes
+ * @param {string} props.time 'current' or 'future'
+ * @param {function} props.updateClientValue Sets state values
```
ethanbb: Hmm, I was using the dashes for params in `App.js`. Do you want to not use the dashes?

knod: I'm not sure how the dashes show up in the docs, but it seems to me like they might turn out looking weird. I think a syntax highlighter like the github one would help at least _me_ not to need the dashes to visually separate the description visually from the var name.

I'll go ahead and merge!

knod: But we can keep talking here if you still have thoughts.

knod: Nice find!

PR #833, opened 2018-10-04
Cleans docs a bit for Predictions section. Not comprehensive. #821

knod:


src/forms/Predictions.js
```diff
@@ -14,26 +14,25 @@ import { BenefitsLineGraph } from './output/BenefitsLineGraph';
...
+ *
+ * @function
+ * @param {object} props
+ * @property {object} props.future Client future/predictive data.
+ * @property {string} props.time Used in class names. Meant to make
+ *     this more easily decoupled in future.
+ * @property {function} props.updateClientValue Update client state
+ *     value.
+ * @property {object} props.snippets Language-specific text
+ * @note As per Project Hope's input, for the first prototype
```
ethanbb: These should all be @param, not @property. That way they show up as sub-objects of `props`.

ethanbb:
@knod any thoughts on using `@module`?

knod:
Sorry, I put something about modules in a comment somewhere, but maybe it was in the wrong place. I'd really hoped to have dropdowns and searching working in our docs before implementing modules, but I'd be fine starting in on that if needed. That said, though, I'm not sure that needs to be in the scope of these PRs.

You and I are on the same wavelength with module names 👍

knod:
The reasoning is that without drop downs or searching, when I've tried it and then looked through the docs I've just found it harder to find things I'm looking for when I can't see function/object/whatever names. For example, I'd have to remember that what I wanted to find was in 'input.js'.

Honestly, though, I'm fine with it. I can look in the code easily, so that's not a problem for me.

The reason for not doing it _right now_ specifically is just to not delay merging these cleanups. Theses have just been running into merge conflicts and some of them are holding up other work. That's part of why these are doing so little - they're a light touch meant to be quick and easy to merge. That hasn't been as effective as I'd hoped it would be, but for good reasons. Still, I think they're faster than they would be otherwise.

ethanbb:
Oh OK, that all makes sense. Thanks for the explanation!

If we just look through some jsdoc templates, we can probably find one has those features and get it working pretty soon.

PR #838, opened 2018-10-06
Help wanted. Tweaks feedback wording. Goals: more plain, resemble other langs?

knod:
To explain that last one - there are certain words that are similar in a lot of romance languages. For example, the English word 'finish' is more similar to 'fin' in French and 'fim' in Portugese. I had trouble finding much for that, unfortunately, and 'done' or 'ready' might be more plain. Thoughts?

Also, I'm not too sure about 'Send it with my information', though I'm not sure 'data' is a useful word.

### DIFF
--- a/src/components/prompts/FeedbackForm.js
+++ b/src/components/prompts/FeedbackForm.js
@@ -69,13 +69,13 @@ class AskPermission extends React.Component {
         disabled = { this.state.submitType === null }
         color    ='teal'>
         {renderIfTrue(this.state.submitType === `withData`,
-          `Send With Data`
+          `Send it with my information`
         )}
         {renderIfTrue(this.state.submitType === `withoutData`,
-          `Send Without Data`
+          `Send it without my information`
         )}
         {renderIfTrue(this.state.submitType === null,
-          `Send`
+          `Send it`
         )}
       </Button>
     </Modal.Actions>,
@@ -189,13 +189,13 @@ class FeedbackPrompt extends React.Component {
           <Form>
             <Form.Input
               { ...inputProps('currentSnap') }
-              label={ 'If amount for the CURRENT SNAP subsidy was wrong, what\'s the correct amount?' } />
+              label={ 'If amount for the CURRENT SNAP amount was wrong, what\'s the right amount?' } />
             <Form.Input
               { ...inputProps('futureSnap') }
-              label={ 'If amount for the FUTURE SNAP subsidy was wrong, what\'s the correct amount?' } />
+              label={ 'If amount for the FUTURE SNAP amount was wrong, what\'s the right amount?' } />
             <Form.Input
               { ...inputProps('futureS8') }
-              label={ 'If amount for the FUTURE Section 8 voucher was wrong, what\'s the correct amount?' } />
+              label={ 'If amount for the FUTURE Section 8 amount was wrong, what\'s the right amount?' } />
             <Form.TextArea
               { ...inputProps('otherCircumstances') }
               label={ 'What else could be going on that could affect your benefit amount? ' +
@@ -220,7 +220,7 @@ class FeedbackPrompt extends React.Component {
           </Button>
           <Button
             onClick={ this.onReady }
-            color='teal'>Ready
+            color='teal'>Finished
           </Button>
         </Modal.Actions>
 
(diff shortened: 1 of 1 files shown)


src/components/prompts/FeedbackForm.js
```diff
@@ -69,13 +69,13 @@ class AskPermission extends React.Component {
...
-          `Send With Data`
+          `Send it with my information`
         )}
         {renderIfTrue(this.state.submitType === `withoutData`,
-          `Send Without Data`
+          `Send it without my information`
         )}
         {renderIfTrue(this.state.submitType === null,
-          `Send`
+          `Send it`
```
ethanbb: Not sure the "it" really adds anything - seems more awkward and I don't see how it would aid comprehension. I like "with/without my information" though.

src/components/prompts/FeedbackForm.js
```diff
@@ -189,13 +189,13 @@ class FeedbackPrompt extends React.Component {
           <Form>
             <Form.Input
               { ...inputProps('currentSnap') }
-              label={ 'If amount for the CURRENT SNAP subsidy was wrong, what\'s the correct amount?' } />
+              label={ 'If amount for the CURRENT SNAP amount was wrong, what\'s the right amount?' } />
```
ethanbb: OK, putting "amount" in the sentence 3 times sounds pretty weird. How about "If the CURRENT SNAP amount was wrong, what's the right amount?"

src/components/prompts/FeedbackForm.js
```diff
@@ -220,7 +220,7 @@ class FeedbackPrompt extends React.Component {
           </Button>
           <Button
             onClick={ this.onReady }
-            color='teal'>Ready
+            color='teal'>Finished
```
ethanbb: I'm neutral on this, but you would probably have to ask a non-native speaker if one is really easier than the other. I think it's clever to try to use a cognate! But "ready" is also a pretty simple word, so in reality they might be equivalent for most people who know any English.

Aside from that question though, "Finished" probably makes more sense here anyway, since "Ready" kind of implies that there's something that comes next, rather than just being "finished" with the form.

src/components/prompts/FeedbackForm.js
```diff
@@ -235,7 +235,7 @@ class FeedbackPrompt extends React.Component {
             submitting         = { this.state.submitting }
             closeAskPermission = { this.closeAskPermission }
             submit             = { this.submit }
-            question           = { `Is it ok if we save the data you've filled in? We won't save your name or any other information that could give away your identity and it'll help us understand your situation better.` } />
+            question           = { `Is it ok if we save the information you've filled in? We won't save anythign else about you. If you had problems, it'll help us solve them.` } />
```
ethanbb: Typo here, "anything" is misspelled.
Not sure we should offer to "solve" their problems. Or at least we could clarify "If you had problems using the app" (as opposed to problems with their life in general).

ethanbb (changes requested):
Good idea! I had a couple issues, though.

src/components/prompts/FeedbackForm.js
knod: I just thought it would mesh better with the other two because they had 'Send it' at the start. But I'm not attached. I'll change it back.

src/components/prompts/FeedbackForm.js
knod: lol, I completely missed that. Yeah, I can do that.

src/components/prompts/FeedbackForm.js
knod: There is something that comes next - that question about sending data or not. I agree that what we really need is some user testing. I wonder if we could do that kind of testing more easily than the other? On mechanical turk somehow or something?

src/components/prompts/FeedbackForm.js
knod: Hehe, thanks for catching those. Hmm... 'help us work on them.'?

src/components/prompts/FeedbackForm.js
ethanbb: Oh, I meant for all three of them actually.

src/components/prompts/FeedbackForm.js
ethanbb: (and for the other similar sentences below.) Thanks!

src/components/prompts/FeedbackForm.js
ethanbb: OK, either way is good for the "help..." part. But I really think it should be "problems using the app" or similar.

src/components/prompts/FeedbackForm.js
knod: Oh, hmm. I see what you're saying. Well, without 'it', it's not a full sentence. I'm not sure how much of a problem that would be for people. Still, I can take it out for now and if we can get our hands on some testing, we can just see how it goes.

ethanbb (approved):

knod:
@justinqtaylor , @jlangevin : If you were holding off because this is a WIP, it no longer is. I'll hold off merging a bit to give you a chance to add your two cents.

PR #839, opened 2018-10-06
Turns result text summary into snippets - mentioned in #795

knod:
Also fixes 'Summary' tab to be accessible too.


knod:
Also addressed #513.

knod:
@todo! Even may be worth making an issue.

PR #840, opened 2018-10-06
Add more tests

wacii:
Sets up JSDOM for use with Enzyme. Adds some tests for the form components. Haven't touch the charts or `<ShowOnYes>`. ~Made a few minor refactors and documentation fixes as I went.~

Still more I want to do, just getting this up there so people know what I'm up to.


src/forms/inputs.js
```diff
@@ -13,25 +13,20 @@ import {
...
-    ...buttonProps,
-  };
-
   return (
-    <Button { ...overriddenDefaults }>{ children }</Button>
+    <Button { ...buttonProps }>{ children }</Button>
   );
-};  // End <BigButton>
+};
+BigButton.defaultProps = {
```
knod: So many awesome tests! Could you break out these kinds of component changes into separate PRs, though. I'd like to be able to accept them because there are other changes I'd like to make.

dylanesque:
> Sets up JSDOM for use with Enzyme. Adds some tests for the form components. Haven't touch the charts or `<ShowOnYes>`. Made a few minor refactors and documentation fixes as I went.
> 
> Still more I want to do, just getting this up there so people know what I'm up to.

Here's a question: Does the ShowOnYes component even need to be tested if the functionality it provides is passing tests on all of the front-end components it's being imported to (CurrentExpenses.js and VisitPage.js, if I'm catching all of these)?

src/forms/inputs.js
wacii: Yeah, I'll do that.

knod:
Sorry, @wacii , I had to steal the jsdom stuff to get a test working in FeedbackPrompt after an edit to fix the modals - #861. Couldn't figure out another way to do it.

wacii:
Perhaps a bit rough, but I wanted to improve test coverage quickly before the codebase deviated too much.

The tests are for components and containers. I think I did all of them except from the visit page and dev hud.

I used snapshots in places I am not entirely happy with. Think it might have been better to use `expect(() => mount()).not.toThrow()` as the default sanity check, and just use snapshots when I wanted to test default props. Or something similar.

Anyways, if any of the snapshots get in the way, feel free to get rid of them.

I'll add tests for the visit page, but I want to extract pieces out of that first. And that really doesn't belong in this PR.

**edit:** Ah, I didn't do all the components. I left `<ShowOnYes>`. Thought someone else wanted to work on that.

knod:
`.not.toThrow()` sounds pretty useful for some situations actually. Above all else, it'd be easier for new members to add their names to the `About` page if they didn't have to fix snapshots at the same time.

knod:
Ask and ye shall receive! 🎉

wacii:
@knod Yeah, I added it. Although on second thought, I am mocking the snippets, so I don't think it would have been a problem. But we didn't need the snapshot test there, so this is better anywho.

knod:
Ah, didn't catch that, but it makes sense.

src/test/forms/rentFields/ContractRentField.test.js
```diff
@@ -0,0 +1,54 @@
...
+  it('when input negative, sets state to invalid', () => {
+    const field = buildField();
+    expect(field.state()).toEqual({ valid: true, message: null });
+
+    field.instance().storeValidator(-1);
+    field.update();
+    expect(field.state()).toEqual({ valid: false, message: null });
+  });
+
+  it('when input less than rent share, sets state to invalid', () => {
```
knod: Would it be accurate to describe this as 'when contract rent less than rent share'?

wacii: I think so. It might be slightly different in that you have the contract rent in the client as well as the potential contract rent in the input. But just saying contract rent is probably clearer.

src/test/forms/rentFields/RentShareField.test.js
```diff
@@ -0,0 +1,54 @@
...
+  it('when input negative, sets state to invalid', () => {
+    const field = buildField();
+    expect(field.state()).toEqual({ valid: true, message: null });
+
+    field.instance().storeValidator(-1);
+    field.update();
+    expect(field.state()).toEqual({ valid: false, message: null });
+  });
+
+  it('when input greater than contract rent, sets state to invalid', () => {
```
knod: 'when rent share greater than...'?

knod (approved):
This is a crazy amount of work. I'm not sure if my assessment is enough by itself since my eyes were bugging out by the end, but this is looking good to me. The test name changes I suggested aren't a blocking change as far as I'm concerned.

src/test/setupJSDOM.js
```diff
@@ -0,0 +1,23 @@
...
+    })
+    .reduce((result, prop) => {
+      return {
+        ...result,
+        [ prop ]: Object.getOwnPropertyDescriptor(src, prop),
+      };
+    }, {});
+  Object.defineProperties(target, props);
+};
+copyProps(document.defaultView, global);
```
ethanbb, line 23: To follow up on my comment during the meeting: although it seems like this might go against `jsdom`'s recommendations, I can see from some googling that it seem like standard boilerplate for using `jsdom` with `enzyme` (and is indeed what `enzyme` recommends). So this does follow the convention for our use case and I don't think we should be too concerned.

src/test/forms/inputs/ControlledRadioYesNo.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+const yesRadio = (control) => {
+  return control.find('Radio[label="Yes"]');
+};
+const noRadio = (control) => {
+  return control.find('Radio[label="No"]');
+};
+const selectYes = (control) => {
+  control.instance().handleChange({}, { value: 'Yes' });
+  control.update();
```
ethanbb, line 15: I'm curious why this couldn't be written similar to `selectNo` (I'm sure there's a good reason...)

ethanbb (approved):
OK, I looked through everything and I think it all looks good! Thanks very much!

src/test/forms/inputs/ControlledRadioYesNo.test.js
knod: I think this has to do with how semantic-ui-react changes the props of their input components in ways that aren't useful. I'm actually surprised that `selectNo()` is different than _this_. I'm curious about that instead.

wacii: Yeah, they should be the same. But for some reason simulate doesn't work on the yes radio. Thought I had figured it out at one point, but I can't seem to remember now. I'll take another look.

src/test/setupJSDOM.js
knod: @todo This might be useful to have in our documentation/wiki somewhere.

knod:
It looks like we can probably merge this and continue the conversation as time allows.

PR #845, opened 2018-10-08
Cleans up last docs in /forms. Very small. Not comprehensive. #821

knod:
Dipped into actually editing content, which I should do for this really quick stuff.


dylanesque (commented):
I'm curious about the type changes in this file that show as from 'string' to 'string | object'. Is that because of some default params or because they're optional values?

knod:
@dylanesque : It's because the argument is allowed to be a string or an object.

dylanesque:
> @dylanesque : It's because the argument is allowed to be a string or an object.

That's what I thought, just wanted to be 100%

dylanesque (approved):

PR #846, opened 2018-10-08
[WIP] ShowOnYes tests

dylanesque:
Initial test and setup for ShowOnYes component

### DIFF
--- a/src/components/AboutContent.js
+++ b/src/components/AboutContent.js
@@ -4,8 +4,7 @@ import { ExternalLink } from '../forms/formHelpers';
 
 // CUSTOM COMPONENTS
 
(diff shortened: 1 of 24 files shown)


src/components/AboutContent.js
```diff
@@ -10,8 +10,7 @@ const AboutContent = function ({ snippets }) {
     <div>
 
 
-      <Header
-        className="ac-header"
+      <Header className="ac-header"
```
knod: I think it'd be best to avoid unrelated changes.

knod: If you do want to make some other changes, though, we can discuss them later.

dylanesque:
corrupted repo

PR #849, opened 2018-10-09
ShowOnYes test

dylanesque:
-Added render test for ShowOnYes
-Test PR for new, uncorrupted fork

### DIFF
--- a/src/forms/ShowOnYes.js
+++ b/src/forms/ShowOnYes.js
@@ -64,7 +64,9 @@ class ShowOnYes extends React.Component {
 
     var show = this.state.show;
     return (
-      <div className = { `show-on-yes` }>
+      <div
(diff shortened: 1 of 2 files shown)


src/forms/ShowOnYes.js
```diff
@@ -64,7 +64,9 @@ class ShowOnYes extends React.Component {
 
     var show = this.state.show;
     return (
-      <div className = { `show-on-yes` }>
+      <div
+        className = { `show-on-yes` }
+        data-test="component-showonyes">
```
ethanbb, line 69: Not sure I like adding a new attribute to this div just so it can be located during testing. Is there a way you can find the `show-on-yes` class instead?

dylanesque:
Squashing this because it came from a corrupted fork. May revisit in the future.

PR #850, opened 2018-10-09
Prints results summary plainly - fixes #844

### LINKED ISSUE #844: Add option to print text summary

knod:
I wanted to add pixel comparisons test for this too, but it took me down too many rabbit holes and those tests will need discussion.

Also, not a huge deal, but for some reason the styles in the print dialogue are slightly different than the styles I get when I use Chrome's tools to emulate printing. Don't know how to troubleshoot those. [Edit: At least, it does when there's a summary there for an actual cliff.]

### DIFF
--- a/src/forms/output/Summary.js
+++ b/src/forms/output/Summary.js
@@ -424,7 +424,7 @@ const Summary = function ({ client, openFeedback, snippets }) {
           className = { `print` }
           color     = { `teal` }
           onClick   = { print }>
-          Print
+          { sn.i_printButton }
         </Button>
       </div>
 
--- a/src/forms/output/Summary.js
+++ b/src/forms/output/Summary.js
@@ -342,6 +342,8 @@ const Summary = function ({ client, openFeedback, snippets }) {
   }  // ends for each benefit
 
   // Ask for feedback
+  // Stays put when printing. A take-home hint
+  // that the tool is still a prototype
   var feedbackAsk = (
     <p>
       <span key = { `pre-ask` }>
@@ -393,6 +395,10 @@ const Summary = function ({ client, openFeedback, snippets }) {
     );
   }  // ends if there's a cliff end
 
+  const print = function () {
+    window.print();
+  };
+
   return (
     <div>
 
@@ -412,6 +418,15 @@ const Summary = function ({ client, openFeedback, snippets }) {
       </div>
       
       { endOfCliffContent }
+
+      <div className={ `print-row` } >
+        <Button
+          className = { `print` }
+          color     = { `teal` }
+          onClick   = { print }>
+          Print
+        </Button>
+      </div>
 
     </div>
   );
--- a/src/index.css
+++ b/src/index.css
@@ -84,6 +84,10 @@ Used to fill vertical space.
   white-space: nowrap;
 }
 
+#cliff-effects-tool {
+  padding: 42px 0;
+}
+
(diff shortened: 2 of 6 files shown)


src/forms/output/Summary.js
```diff
@@ -342,6 +342,8 @@ const Summary = function ({ client, openFeedback, snippets }) {
   }  // ends for each benefit
 
   // Ask for feedback
+  // Stays put when printing. A take-home hint
+  // that the tool is still a prototype
```
ethanbb, line 346: Hm, I'm not seeing this in the print preview. Not sure why.

src/index.css
```diff
@@ -478,3 +489,46 @@ ACCESSIBILITY - Also see any other focus styles
...
+  #root #App .ui.segment {
+    margin: 0;
+  }
+
+  @page {
+    size: portrait;
+  }
+
+  #App {
+    filter: grayscale(100%);
```
ethanbb: I feel like we don't necessarily have to make it black-and-white; people can print in black and white if they want to anyway. And some printers might sneakily still use color ink even if printing B&W content in color mode. But not a big deal.

ethanbb (commented):
Neat! One more thing I noticed is that in the Chrome dialog, if you check "background graphics" (which is unchecked by default), the footer appears in a block of gray (well, green converted to gray). We might want to change the colors on that on our end, in case some browsers don't have that feature.

src/forms/output/Summary.js
knod: Weird, I'll double check later today.

src/index.css
knod: If the setting is automatically set to print in black & white, then that's fine and we can get rid of a style. I'll do a bit of research on it to make sure.

src/forms/output/Summary.js
knod: I did it now instead, and I do see the 'let us know' line when there's a cliff. The button just doesn't have a background.

knod: If you don't see it, could you post an image here of what you do see?

knod:
@ethanbb : Oh, I think I misunderstood the note. Should the footer have no background color at all? What are your intentions, sir!?

src/forms/output/Summary.js
ethanbb: Oh, I didn't realize it was only supposed to show up if there's a cliff?

ethanbb: Oh never mind, I wasn't reading closely and I thought this was the yellow bar asking for feedback above the tabs, for some reason.

ethanbb:
@knod I was saying we should remove the background color from the footer with CSS if possible, rather than relying on Chrome doing it automatically.

knod:
@ethanbb : Got it. Is that something that we can push off for a future PR? Is it acceptable as it is as a working feature?

knod:
You're right, though - it looks like not all browsers give choices about backgrounds: https://answers.microsoft.com/en-us/windows/forum/apps_windows_10-msedge/microsft-edge-does-not-allow-users-to-print/f1631753-cfe6-44c0-be34-b1dd10ae9b46.

knod:
Question still stands!

knod:
Because of interesting issues with Travis, it looks like I have to close this and re-do the PR, which means we'll lose the comments. I'll try to remember to put another comment in here with the link to the new one.

PR #854, opened 2018-10-10
Test button reset

rob111:


knod:
@rob111 : This is some good work! There's just the problem of package-lock.json. Ping me on Slack if you're not sure what I'm talking about!

knod (commented):
Just have to solve the issue with the package-lock.json and then it'll be ready!

knod:
You might look into how to reset a single file to an older commit.

knod:
This is being moved to #889

PR #855, opened 2018-10-10
Add ability to disable warning modal to devhud

jlangevin:


knod (changes requested):
Thanks for getting this ball rolling so quickly! The HUD option's looking good. When you refresh, though, the setting's gone and the modal comes back. May be because it's not tied into the `devProps` object. That's the only data that ever survives reloading. Never articulated it like that before, so maybe I'll add a bit of documentation about it.

jlangevin:
Ah gotcha.  I was unaware this needed to survive reloads.   I'll update to leverage localStorage.

knod:
@jlangevin : Just so you don't have to re-invent the wheel, the `devProps` and `setDevProp()` do that already, so you can just hook into those.

knod:
That's how those dev hud setting persist with reload.

jlangevin:
Okay, thanks.  Started looking at that, but will take a closer look when I return to this.

knod:
Y'know what, that's my bad. That's an extra feature and not a blocking change. I'll approve and you can decide whether to add that on this PR or merge it as is. I'll be happy to tie it into `devProps` afterwards myself.

knod (approved):
A solid PR that does exactly what was requested!

jlangevin:
No worries.  Thanks for the approval.  I'll merge this to reduce the chance of a conflict.  Keep an eye out for another PR either later today or this weekend on the local storage.  Should be a pretty straight forward tweak.

PR #857, opened 2018-10-10
Improve readability of SNAP calcs

knod:
And adds documentation


src/programs/federal/snap.js
```diff
@@ -49,140 +51,170 @@ const getSNAPBenefits = function (client, timeframe) {
...
+ *     to mess around with actual client values.
+ * 
+ * @namespace
+ *
+ * @note Almost all require the `current` or `future`
+ *     prop of the typical `client` object as an
+ *     argument.
+ */
 var SNAPhelpers = {},
     hlp         = SNAPhelpers;
```
ethanbb: While we're making this file clearer: could we move these declarations up above `getSNAPBenefits`? I was scratching my head trying to figure out where `hlp` was coming from.

src/programs/federal/snap.js
```diff
@@ -12,30 +12,32 @@ import { getLimitBySize } from '../../utils/getGovData';
...
+      netIncomeTestResult     = hlp.passesNetIncomeTest(client),
+      maxSnapAllotment        = getLimitBySize(SNAPData.SNAP_LIMITS, householdSize),
+      percentageOfNetIncome   = hlp.getNetIncome(client) * SNAPData.PERCENT_OF_NET,
       maxClientAllotment      = Math.max(0, maxSnapAllotment - percentageOfNetIncome);
 
   if (grossIncomeTestResult === true && netIncomeTestResult === true) {
 
     if (maxClientAllotment <= SNAPData.SMALL_HOUSEHOLD_MIN_GRANT) {
-      if (hlp.householdSize(client) <= SNAPData.SMALL_HOUSEHOLD_SIZE) {
+      if (householdSize <= SNAPData.SMALL_HOUSEHOLD_SIZE) {
```
ethanbb, line 54: Currently, with this nested if statement, a household larger than `SMALL_HOUSEHOLD_SIZE` but with a max client allotment of less than `SMALL_HOUSEHOLD_MIN_GRANT` will get a `finalAmount` of 0. If that's a mistake, we should change this to a single if statement with an &&.

src/programs/federal/snap.js
```diff
@@ -12,30 +12,32 @@ import { getLimitBySize } from '../../utils/getGovData';
...
-  getUnder13OfHousehold,
+  getYoungerThan,
+  getOlderThan,
 } from '../../utils/getMembers';
 
 
-/** Based on https://www.masslegalservices.org/SNAPCalculator */
-const getSNAPBenefits = function (client, timeframe) {
+/** Based on https://www.masslegalservices.org/SNAPCalculator
+ * @note Bay State CAP not included as this prototype only deals with
```
ethanbb: I don't think `@note` is a real thing in `jsdoc` - do you see it in the output?

ethanbb (commented):
OK this is pretty long. I just looked at the top of the first file so far. Will come back.

src/programs/federal/snap.js
knod: Good idea!

src/programs/federal/snap.js
knod: Dagnabit, it doesn't. I'll have to remove that wherever it is in the comments in general. It should be! And @warning should be too!

src/programs/federal/snap.js
knod: Not sure, but that's more of a problem with the logic and not this PR, right?

ethanbb: Yeah true, I've just never read through this file before and thought it seemed important.

knod: I don't remember any particular one of these, but I do remember finding several places where the logic didn't make sense to me and ended up matching the mass legal logic. We can take another look at that later for sure. @todo.

Other than that, are more changes needed?

knod:
Because of issues with Travis and doing PRs from the same source branch onto multiple upstream branches, have to close and re-open a new PR for this. Will mention this one in that one to keep them linked, but the comments will not carry over.

PR #858, opened 2018-10-10
Improving Color Contrast on Selected Site Elements #816 and fixes #852

### LINKED ISSUE #852: Improve Color Contrast on Selected Site Elements
More contrast needed for hover and other interactions, including:
 - [x] General nav bar
- [x] Form section nav bar
- [x] Spouse vs. other member drop down in 'Household' form section

dylanesque:
-Starting with the main nav bar, I'll move on to the next item when there's a basic consensus. I tried this out on the site, and didn't love the look of the hover state manifesting as the background color, but that was with the text being white and I'm open to suggestions. I feel like the transition is too rushed and would like to slow it down, but again, I want to hear from others.


knod:
As long as this is enough contrast between the hovered text and its background, this looks good to me. I like the speed of the transition. It tells me what's going on without me having to stick around for the length of time a longer transition might take.

dylanesque:
Changes are complete and ready for review (I see, and will take care of the merge conflicts).

src/components/MainMenu.js
```diff
@@ -15,9 +15,9 @@ const MainMenu = function (props) {
         inverted
         secondary
         size='large'>
-        
-        <Menu.Item><Link to="/">Home</Link></Menu.Item>
-        <Menu.Item><Link to="/about">About</Link></Menu.Item>
+
+        <Menu.Item><Link id="main-nav" to="/">Home</Link></Menu.Item>
+        <Menu.Item><Link id="main-nav" to="/about">About</Link></Menu.Item>
```
knod: I apologize if I didn't notice this before, but I believe html id's have to be unique on a page.

src/index.css
```diff
@@ -478,3 +484,13 @@ ACCESSIBILITY - Also see any other focus styles
   outline: 2px solid rgba(0, 121, 116, 1);
   outline-offset: 1px;
 }
+
+/* Semantic UI Default Override */
+
+  .mini.ui.steps a.step:hover {
```
knod: Unfortunately, the little diamonds between the steps are still the same color on hover as before and show up pretty starkly when you hover over a step. Could you add their selectors to this so they can blend in like they did before?

src/index.css
```diff
@@ -478,3 +484,13 @@ ACCESSIBILITY - Also see any other focus styles
...
+  background: #dcddde;
+}
+
+    div[role="option"].active.selected.item:hover {
```
knod: This only changes the hover color for the active and selected item. The non-active/selected item(s) don't get the same hover color. Is there a specific reasoning behind that?

src/containers/VisitPage.js
```diff
@@ -360,6 +360,7 @@ class VisitPage extends Component {
             minWidth='874.5'
             style={{ padding: '14px 0' }}>
             <StepBar
+              className="testorr"
```
knod: Can probably remove this at this point, though I like the name. May have to steal it for myself sometime!

knod (changes requested):
Thanks for the work so far! A bit more and I think we'll be there!

knod:
Also, I don't think we can approve until the conflicts have indeed been resolved, though that doesn't mean we can't discuss stuff!

src/components/MainMenu.js
dylanesque: That's correct, they can only appear once per page.

src/containers/VisitPage.js
dylanesque: Whoops!

src/index.css
dylanesque: This should be fixed now.

src/index.css
dylanesque: The effect seen here is my interpretation of the description as described in the issue, namely that the hover effect only needs to be added to the currently selected item on hover, with no other changes.

src/components/MainMenu.js
knod: Here, there are two id's with the same name.

src/index.css
knod: If you're talking about what @2001zhaozhao described, I can understand the confusion. If you want, we can merge this based on that.

I think his intention, though, was the 'hover' state. He also only described form nav bar menu styles that way, but I think that was meant to be on 'hover' as well, which is the way you implemented it.

The confusion in terms might be because he's not super familiar with semantic-ui classes or CSS. @2001zhaozhao : Whatever your intention was, no worries, you input has been really useful in moving this forward 👍

src/components/MainMenu.js
knod: Here, there are two id's with the same name.

src/index.css
knod: If you're talking about what @2001zhaozhao described, I can understand the confusion. If you want, we can merge this based on that.

I think his intention, though, was the 'hover' state. He also only described form nav bar menu styles that way, but I think that was meant to be on 'hover' as well, which is the way you implemented it.

The confusion in terms might be because he's not super familiar with semantic-ui classes or CSS. @2001zhaozhao : Whatever your intention was, no worries, you input has been really useful in moving this forward 👍

src/components/MainMenu.js
knod: Here there are two id's with the same name.

src/index.css
knod: Ah, you've changed that form nav bar interaction now. Well, if you want, we can accept both of them and we can discuss the idea of applying it to all hovered elements.

knod: Hmm, no, something different happened. I think the styles aren't being applied anymore at all.

src/components/MainMenu.js
dylanesque: Oh yeah, I see, those are def supposed to be classes

src/index.css
dylanesque: We probably should discuss this in person Tuesday night, I'm definitely seeing the changes I applied still working.

src/index.css
```diff
@@ -42,6 +42,12 @@ body {
   padding: 1em 0em;
 }
 
+/* MainMenu.js styles */
+
+.ui.inverted.menu .item, .ui.inverted.menu .item>a:not(.ui):hover {
```
knod, line 47: Note for later discussion: not seeing these styles happening anymore.

knod: I think I put this on the wrong item. Forget this.

knod:
Unfortunately, once the conflicts are handled, the tests'll fail. We'll have to pull in that test fix to be able to move forward on anything new.

PR #859, opened 2018-10-12
Adds outlines back into the home page buttons

knod:


ethanbb (approved):
Looks good. Is it necessary now that we have the color change though? Sorry, I think I requested this, but before the color change on tab was working...

Or maybe the outline is easier to see for some people. The color change is pretty drastic though.

knod:
It's true, but if nothing else the outlines stay consistent with the style of the rest of the app, which seems useful. Also, if we don't do that, then the buttons get that usual fuzzy blue outline. The last thing is that our button styles may change, though I hope it's always to something visible, so maybe that's not a big deal.

PR #861, opened 2018-10-12
Fixes one modal from blocking other modal

knod:
When the feedback form was opened, the modal that was asking for feedback stayed in front of the feedback form.


dylanesque (approved):

knod:
Some scripts stolen from  #840.

PR #864, opened 2018-10-13
Removes carats from deps. Fixes #856

### LINKED ISSUE #856: Fix package-lock.json issues by removing carats from dependencies

knod:
Sorry, folks, not sure what the ultimate plan originally was for fixing this, but maybe this can be a good temporary measure.


knod:
Also, there were two sort-of discrepancies - in package-lock.json ajv was at 5.5.2, despite the fact that our package.json said `"ajv": "^6.5.4"`. Maybe we hadn't updated package-lock for a while or something.

dylanesque (approved):

PR #865, opened 2018-10-13
[WIP] Removes `install` fix, no longer needed. Numbers don't matter.

knod:
Also tweaks wording a bit.


dylanesque (changes requested):
Not quite there, note the spelling and spacing errors on line 44

knod:
Thanks! Nice catch!

knod:
Apparently this isn't fixed yet... so... the quest continues!

PR #872, opened 2018-10-14
More robust styles and structure for left-fixed buttons

knod:


knod:
Also prep for combining the reset buttons into one reset button

dylanesque (dismissed):
Almost there, the render method of VisitPage.js would be a lot more readable with any and all state and props items destructured.

knod:
@dylanesque : I do love that idea! Is there a reason it's appropriate as part of this PR? We try to keep PRs focused on one task at a time.

knod:
@dylanesque : Just pinging you again on this question.

knod:
The change requested wasn't related to the PR, no other changes have been requested, and this PR has been sitting here for a while, so it's being merged.

PR #876, opened 2018-10-14
Header structural adjustments

knod:
May conflict with other header PR, not sure.


dylanesque (approved):

ethanbb (approved):
I like the reorganization and the preview looks good to me. Trusting you on the CSS.

knod:
The CSS is slightly different, but I don't think it's a terrible thing. [That is, I think this may be preferable in several ways and no one has seemed attached to the CSS in the past, so I went ahead and made a call. We can pretty easily restore the one style difference that was lost.]

PR #877, opened 2018-10-16
Updates status of the project

knod:


dylanesque (commented):
How pedantic do you want me to be? I think this is mostly very good, but I think some of the wording underneath the initial cliff effect can be strengthened.

knod:
@dylanesque : lol, not too pedantic for now. If this is good enough we can start with this and debate more language later. I may also point you at other folks to discuss wording who are more in tune with that kind of thing.

dylanesque:
> @dylanesque : lol, not too pedantic for now. If this is good enough we can start with this and debate more language later. I may also point you at other folks to discuss wording who are more in tune with that kind of thing.

The suggestions, which I think are subjectively better, but am not married to:

"you're taking home less in all" --> "you're taking home less overall"

"you don't make the cutoff for some of those benefits."  --> "you have gone past the cutoff..."

"Some benefits drop to nothing, or almost nothing." --> "Some benefits get drastically reduced, or disappear entirely." ( I feel like this is stronger if the consequences of the cliff effect are escalated upwards through the passage.)

knod:
@dylanesque : Those sound like great discussion items. I'll add @todo in here. If you want, you can add them as an issue and it'll probably be easier to keep track of them there. In the meantime, since those aren't crucial, is this ready for merging?

knod:
It sounds like the change is a blockage to merging. I'll work on that and we'll see where we get.

dylanesque (approved):

PR #882, opened 2018-10-17
Fix reset

knod:


knod:
Since Travis was having a bunch of trouble, I went in a different direction than PRs...

PR #884, opened 2018-10-21
Improve readability of SNAP calcs

knod:
See #857 for discussion.


knod:
These particular test failures aren't related to this PR and are being fixed in a different one. Let's not merge it till they're fixed, though.

knod:
Got some informal feedback and it seems this is an improved version. It also doesn't change functionality and doesn't need to sit here forever. Merging will happen!

PR #888, opened 2018-10-23
Abstract calcing so each state can have its own. Fixes #887

### LINKED ISSUE #887: Make calculations US State-specific
Abstract data for calculations so that it can be specific to each US State.

This doesn't include abstracting federal regulations. We need an SME for that.

knod:


wacii:
Unless you want to allow different executions of `applyAndPushBenefits()` to use different state's regulations within the same app, I would not provide the ~current~ state as an option to the function. That is just creating a potential source of errors that doesn't need to exist.

One solution~s~ is that you could instead have a setter in the `applyAndPushBenefits()` module.

```js
var USState = 'MA';
export const setUSState = state => USState = state;
```

This way you would set it once, and then all uses of the function would not have to worry about ~setting~ the correct state themselves ~properly~.

There are other ways of doing it, I imagine most of which involve finding ways to store them in the react component tree, but just wanted to give an example.

dylanesque (approved):

knod:
As per our discussion, this one item isn't going to deal with @wacii 's comment, though state will be a thing.

Issue #446, opened 2018-03-21
Add regression tests for user interface components


ethanbb:
Inspired by #443. Use enzyme (including click simulation) to verify that things like editing/focusing/blurring each type of inputs, clicking buttons, navigating, and throwing errors do what we want them to. Start with ManagedNumberField, CashFlowRow, MonthlyCashFlowRow, etc.

Branch: https://github.com/ethanbb/cliff-effects/tree/test-interactions

shp54:
https://docs.google.com/document/d/13BtVvZi2C6lusC5xUeTKxYP9LpLWW5geuaW903UE7ts/edit

Writeup so far on UI testing

knod:
This is that same document, but in the google shared folder for Cliff Effects/Technical/Test: https://docs.google.com/document/d/1IjLVvwEg0RRn5bz2ArswMZ6ERp2q0AjO9xvjLv1luxc/edit?usp=sharing

knod:
There's some stuff in there that may be worth pursuing, btw, like the SNAP stuff. Could swear I've never seen it before in my life.

knod:
This is still ongoing

knod:
Hmm, I think the 'testing' project is a good place to keep the link to the doc and issues could contain more specific tests.

PR #889, opened 2018-10-24
Tests ButtonReset component #446

rob111:


src/test/components/ButtonReset.test.js
```diff
@@ -0,0 +1,19 @@
...
+import { ButtonReset } from '../../forms/ButtonReset';
+
+import { CLIENT_DEFAULTS } from '../../utils/CLIENT_DEFAULTS';
+
+
+describe('Test ButtonReset component', () => {
+  it('Test click event', () => {
+    let mockCallBack = jest.fn();
+
+    let button = shallow(<ButtonReset onClick={mockCallBack} snippets={{}}></ButtonReset> )
```
knod, line 14: Just a note for the future - if you wrap this in a `<div>` you can then do `.find('ResetButton')` instead.

knod (approved):
This looks good to me! Feel free to merge it when you want. I added a note that might be helpful in the future, but you don't need to change anything right now.

src/test/components/ButtonReset.test.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { shallow } from 'enzyme';
+
+import { ButtonReset } from '../../forms/ButtonReset';
+
+import { CLIENT_DEFAULTS } from '../../utils/CLIENT_DEFAULTS';
```
turnerhayes, line 7: It doesn't look like `CLIENT_DEFAULTS` or `renderer` is used in this file. Can these be removed?

turnerhayes (approved):
This branch has some lint errors, too (although they don't seem to be at all related to this PR--don't know how we generally treat things like that. Separate ticket? @knod )

knod:
@turnerhayes : Yeah, linting is separate.

src/test/components/ButtonReset.test.js
knod: Wonder why the tests aren't catching that. It's supposed to catch warnings like 'unused vars' last I checked. Wonder if/when that changed. Regardless, it doesn't break anything.

turnerhayes (approved):

PR #894, opened 2018-10-24
Reuses 'ResetButton' for all 'New Client' buttons. Fixes #893

### LINKED ISSUE #893: Use `<ButtonReset>` for all 'New Client' buttons
May require changes in other components and tests.

knod:
Also renames test to more accurate name and adjusts to new
component requirements.


src/containers/VisitPage.js
```diff
@@ -374,7 +373,12 @@ class VisitPage extends Component {
         </Container>
 
         <Container id={ `alwaysLeftButtons` }>
-          <ResetAnytime askToResetClient={ this.askToResetClient } />
+          <ButtonReset
+            onClick   = { this.askToResetClient }
+            snippets  = { snippets }
+            overrides = {{ id: `resetFixed`, size: `medium` }}>
+            New Client
```
turnerhayes: This should be localized into a snippet

knod: True. It's new code. It should be snippetized if someone experienced is working on it. Thanks!

src/forms/ButtonReset.js
```diff
@@ -21,8 +21,9 @@ const ButtonReset = function ({ onClick, snippets, overrides }) {
 
   return (
     <BigButton
-      onClick = { clickHandler }
-      { ...overrides }>
+      className = { `caution` }
+      onClick   = { clickHandler }
+      overrides = { overrides }>
       { snippets.i_newClient }
```
turnerhayes: It looks like `ButtonReset` doesn't use `props.children`; can we have this render `{ children }` instead of the snippet? Or if this will never be used with any other text, not pass children in `VisitPage`?

src/forms/inputs.js
```diff
@@ -19,15 +19,25 @@ import {
...
+    overrides = {};
+  };
+
+  // Don't override classes, combine them
+  var allClasses = `big-button`;
+  if (className) {
+    allClasses += ` ` + className;
+  }
 
   // Combine class names?
```
turnerhayes: Looks like we can remove this comment now!

turnerhayes (approved):

turnerhayes:
Looks good; a couple of minor points but nothing blocking.

src/forms/ButtonReset.js
knod: lol, that occurred to me late last night after I'd already made the changes to the `VisitPage` changes. In general, I kind of dislike having that kind of silent default. Obviously it's worse than a silent default right now, though. It'll have to change and I'll go ahead and do it for this PR.

src/forms/inputs.js
knod: True!

src/forms/inputs.js
```diff
@@ -19,15 +19,23 @@ import {
 *
 * @returns {object} React element
 */
-const BigButton = function ({ children, ...buttonProps }) {
+const BigButton = function ({ children, className, overrides }) {
```
turnerhayes, line 22: This actually seems to have broken the BigButton because some props, such as `onClick` are no longer passed to the underlying button. I'll open a new issue for it.

knod: Thanks!

knod: I'm curious if you have other thoughts about how to handle this. I'm not a huge fan of passing on any random prop unfiltered, but `overrides` seems like a pain to maintain.

PR #904, opened 2018-10-25
Add linting pre-commit hook via husky and lint-staged packages. Fixes #897

### LINKED ISSUE #897: Add pre-commit hook to lint project
[Husky](https://github.com/typicode/husky) is a good solution for adding git hooks; we can add a pre-commit hook that will lint, fix, and commit any changes (assuming any violations are fixable).

turnerhayes:
Test plan:

- Change some file in such a way that it fails linting.
- Run `git commit -am 'test commit'`
- The commit should run a lint check first
- Look at the commit diff and make sure it's changed to pass lint

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,15 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@samverschueren/stream-to-observable": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/@samverschueren/stream-to-observable/-/stream-to-observable-0.3.0.tgz",
+      "integrity": "sha512-MI4Xx6LHs4Webyvi6EbspgyAb4D2Q2VtnCQ1blOJcoLS6mVa8lNN2rkIy1CVxfTUpoyIbCTkXES1rLXztFD1lg==",
+      "dev": true,
+      "requires": {
+        "any-observable": "^0.3.0"
+      }
+    },
     "@types/node": {
       "version": "10.0.10",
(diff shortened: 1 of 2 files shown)


knod (approved):
Tested and it's looking good! I like that it only fixes staged changes.

knod:
You can go ahead and merge whenever you're ready!

@todo - a couple thoughts to discuss:

Carats in our versions is something we're staying away from. We were hoping it would solve our package-lock.json problem and it hasn't, but until we figure it out, we may remove them from the version numbers so that further efforts don't get confounded by them.

It's verbose, which might make newer folks nervous, but that's a conversation for another time. There's currently no way to turn that off. I do generally go for explicit over implicit. I'm not sure about this specific situation, but since there's [currently no way to turn off logs](https://github.com/typicode/husky/pull/300), it's a moot point for now.

Lastly, it's strange how it can make a commit even if the linting is the only thing that happened. I ruined some code format and made a commit just for that. After linting, there were no changes and it still committed. I wonder how it does that.

knod:
Hmm, it doesn't quite manage to limit `npm lint` to just staged files, though it doesn't add the unstaged files, just leaves them changed and un-added. I'm not sure we can leave things like that. Considering this will be in everyone's code and all files will be linted when committed and there shouldn't be leftover files to lint, it shouldn't be a problem. In that case, though, we might consider getting rid of 'lint-staged'.

knod (approved):
Eh, this is an improvement regardless. We can refine it later.

knod:
@todo - 'lint-staged' and carats

PR #905, opened 2018-10-25
Change var to const/let, and other eslint recommended rules, fixes #899

### LINKED ISSUE #899: Convert var to const/let

turnerhayes:


knod:
It doesn't seem broken to me, but if it were broken, it could be in a really subtle way. Since I'm the one who's probably the biggest offender with `var`, I will say that I take great care about `for` loops and other similar `var` issues, so I hope we're in a reasonable position.

dylanesque (dismissed):

dylanesque:
Whoops, hit approve prematurely. This looks great, just needs those merge conflicts sorted.

knod:
Hmm, too many things changing at the same time. @turnerhayes : How about we merge some PRs, figure out a date and time for you to make the PR, make no changes in that half hour, and merge your PR before moving on?

knod:
It's got some changes now that bear examining, so let's hold off for a sec while we explore these.

PR #906, opened 2018-10-25
Changes var name to reflect purpose (not terms of service)

knod:


dylanesque (changes requested):
I think this warrants a consensus discussion, the new var name strikes me as counter-intuituive and confusing.

knod:
Yeah, there's probably a better name. It just needs a different one.

knod:
`distrustConfirmed`?

knod:
If you're working on the code, you should be aware that a user should not trust the site.

dylanesque:
> `distrustConfirmed`?

Yeah, that's much better IMHO

knod:
Updated the name and renames the toggling function too.

dylanesque (approved):

PR #910, opened 2018-10-25
[WIP] Replace var with const/let, add eslint recommended rules. Fixes #899

### LINKED ISSUE #899: Convert var to const/let

turnerhayes:


knod:
Hmm, didn't realize it was introducing a lot of new rules. Making this a WIP until we have time to take it in.

knod:
WIP's should not be merged.

knod:
Sorry that I didn't notice the other rule changes before.

.eslintrc.json
```diff
@@ -48,6 +51,10 @@
     "react/jsx-tag-spacing": ["warn"],
     "react/self-closing-comp": "warn",
     "space-in-parens": ["warn", "never"],
-    "space-infix-ops": ["warn"]
+    "space-infix-ops": ["warn"],
+    "no-var": ["error"],
+    "no-undef": ["error"],
+    "no-unused-vars": ["error"],
```
knod, line 57: Can the linter automatically fix `no-unused-vars`? Same question for the other config options here. Our decision currently is that linting issues shouldn't block a commit or PR.

turnerhayes: It looks like it can fix some (most)? `no-var`s, but not `no-undef` or `no-unused-vars`.

PR #913, opened 2018-10-26
Inline Style Reafactoring

dylanesque:


knod (approved):
Visual check complete and it looks solid! Thanks for the well-focused refactor!

turnerhayes (approved):
This makes me so happy! I love getting rid of inline styles.

PR #915, opened 2018-10-27
Replace var with const/let; fixed some existing lint errors

### LINKED ISSUE #899: Convert var to const/let

turnerhayes:
Fixes #899

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -46,10 +46,10 @@ class App extends Component {
   constructor (props) {
     super(props);
 
-    var defaults = cloneDeep(CLIENT_DEFAULTS);
+    const defaults = cloneDeep(CLIENT_DEFAULTS);
 
     // Development variables are the only things stored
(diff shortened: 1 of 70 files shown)


src/App.js
```diff
@@ -258,14 +258,14 @@ class App extends Component {
...
-            null
-          )
+          <DevHud
+            devProps = { devProps }
+            funcs    = { devFuncs }
+            data     = {{ default: clients.default }}
+            state    = { this.state } />
+        ) : (
+          null
+        )
```
knod, line 268: `var/let/const` stuff looking good so far! This doesn't happen when I do `npm run lint`, though. Where is this coming from?

knod: Or maybe I'm just wrong... I think my editor's stopped showing me lint errors.

knod: Regardless, that's not what I wanted that to look like. That said, it's also not this PR's problem. [Meaning that it shouldn't block this PR]

PR #919, opened 2018-10-27
Test improvements

wacii:
There were inconsistent helpers in the `<ControlledRadioYesNo>` test that I wanted to investigate before fixing.

Also some improvements to the snapshot testing. I made sure each has a meaningful name, broke up tests that contained multiple snapshot comparisons, and converted the ones that were basically smoke tests to test for errors explicitly using `.not.toThrow()`.

The intent of these tests should be clearer now, a general improvement but it also makes it easier to convert them to proper assertions, if that is what people want to do.

This only involved the tests I wrote.


src/test/components/GraphTimeButtons.test.js
```diff
@@ -11,7 +11,12 @@ describe('<GraphTimeButtons>', () => {
     expect(wrapper.find('#Yearly').exists()).toBe(true);
   });
 
-  it('matches snapshot', () => {
-    expect(mount(<GraphTimeButtons />)).toMatchSnapshot();
+  it('renders with provided props', () => {
+    const props = {
+      activeID:   'not sure what this should be',
```
knod: iirc, that's supposed to actually be a time interval value - weekly, monthly, or yearly.

wacii: Whoops. Meant to look that up.

src/test/components/GraphTimeButtons.test.js
```diff
@@ -13,7 +13,7 @@ describe('<GraphTimeButtons>', () => {
 
   it('renders with provided props', () => {
     const props = {
-      activeID:   'not sure what this should be',
+      activeID:   'monthly',
```
knod: Hmm, it might need to be 'Monthly'.

knod: @todo Change `activeID` to `timeInterval`

knod (approved):
That looks good to me! I have questions about 'why this and that', but that's just curiosity and we can chat about it over candy corn and apples or whatever.

turnerhayes (approved):
Looks good to me!

PR #920, opened 2018-10-27
[WIP] extract client code from view

wacii:
Putting this up here more for discussion than anything. Is this the sort of code we want in the app?

The client is extracted into a class, which the view components can subscribe to for updates.

### DIFF
--- a/src/ClientModel.js
+++ b/src/ClientModel.js
@@ -1,0 +1,105 @@
+// DATA MANAGEMENT
+import { setNestedProperty } from './utils/setNestedProperty';
+import { cloneDeep } from 'lodash';
+import { convertForUpdate } from './utils/convertForUpdate';
+
+// Data
+// import { clientList } from '../config/dummyClients';
+import { CLIENT_DEFAULTS } from './utils/CLIENT_DEFAULTS';
+
+export default class ClientModel {
+  constructor() {
+    this.listeners = [];
+    this.data = cloneDeep(CLIENT_DEFAULTS);
+    this.default = cloneDeep(CLIENT_DEFAULTS);
+    this.userChanged = {};
+    this.housing = this.data.current.housing;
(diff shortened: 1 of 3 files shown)


knod:
Paraphrasing from Slack:
I like the idea of extracting into a class! It sounds like we'll save Redux for where we need it, partly to help show its clear use cases.

src/ClientModel.js
```diff
@@ -0,0 +1,105 @@
...
+import { CLIENT_DEFAULTS } from './utils/CLIENT_DEFAULTS';
+
+export default class ClientModel {
+  constructor() {
+    this.listeners = [];
+    this.data = cloneDeep(CLIENT_DEFAULTS);
+    this.default = cloneDeep(CLIENT_DEFAULTS);
+    this.userChanged = {};
+    this.housing = this.data.current.housing;
+    this.isBlocking = false;
```
turnerhayes, line 17: What does `isBlocking` mean? I had thought it had something to do with blocking navigation from the visit page. Does it mean something in the context of the client?

knod: Not sure if it specifically needs to be in here. When the client has been changed, the navigation is blocked, that's all. We could possibly change this to `.isDirty` or something if we're going to put it in the client object.

We could possibly just check if `.userChanged` has any properties. I'm not sure which is worse, lol. Though I could see a future in which we check if the current values are different than the default values and set `.isDirty` based on that.

turnerhayes: I see. Yeah, I think the name `isBlocking` makes sense in VisitPage but not in the ClientModel. If it's in the ClientModel, it should probably be `.isDirty`. That could be a member property that's updated every time updateClientValue is called.

knod: That said, I personally don't count that as a blocking change here.

PR #922, opened 2018-10-27
Fixes #545, Form routes

### LINKED ISSUE #545: Back button in browser behavior unexpected, `Forms` to pages/routes instead

turnerhayes:
Test plan:

- Mark one or more of the benefits in the first step.
- Click on steps in the step bar. The URL should change and you should not see any dialog asking you to confirm navigation.
- Click on a link outside the visit workflow (e.g. Home, About)
- You should see a navigation confirmation dialog

- Reload the page while on one of the later steps (not the first step)
- Before the page changes, you should see a navigation confirmation dialog
- After reloading, the visit should go back to the first step

- If you have not selected any benefits, you should not see any confirmation dialog, even when going outside the visit workflow.


src/containers/VisitPage.js
```diff
@@ -62,37 +52,19 @@ class VisitPage extends Component {
...
-      {
-        form:              PredictionsStep,
-        key:               'predictions',
-        updateClientValue: this.changeFuture,
-      },//,
-    //  { title: 'Graphs', form: ResultsGraph }
-    ];  // end this.steps {}
+  componentDidMount() {
+    // If the user has refreshed the page while they were on a later step, then we will
+    // go back to that step, but since the client data is currently stored in memory, the
```
knod: 'isn't stored in memory', I think

knod: This also makes it sound like returning to the current step is the currently expected behavior.

knod: Hmm. 'If the user... later step, they'd usually be automatically sent back to that step...'?

src/containers/VisitPage.js
```diff
@@ -221,55 +193,55 @@ class VisitPage extends Component {
...
   };
-
+  
   previousStep = () => {
-    this.setState((prevState) => {
-      return { currentStep: prevState.currentStep - 1 };
-    });
+    const prevStepIndex = this.getCurrentStepIndex() - 1;
+
+    if (prevStepIndex < 0) {
```
knod, line 206: What would call this function that would go below 0? The first form section doesn't have a 'Previous Step' button, so unless the navigation 'back' button does it, I'm not sure what would.

src/containers/VisitPage.js
```diff
@@ -62,37 +52,19 @@ class VisitPage extends Component {
...
-    ];  // end this.steps {}
+  componentDidMount() {
+    // If the user has refreshed the page while they were on a later step, then we will
+    // go back to that step, but since the client data is currently stored in memory, the
+    // client data will be lost. So if we are loading on a step that is not the first step,
+    // go back to the first step to start again
+    const firstStepURL = `${this.getPathPrefix()}/${STEPS[ 0 ].key}`;
 
-  };  // End constructor()
+    if (this.props.history.location !== firstStepURL) {
```
knod: Is this needed? If you're on the first step, you can still go to the first step without difficulty. Is it heavy to replace history and navigate to a step? Are there other reasons to avoid it?

knod (commented):
I'm pausing here to say great work! This is some awesome looking stuff! I'm not sure anything I've found so far is blocking, though I do have to take another go to look more thoroughly and I need to test it out. Most of these comments are just clean up notes and don't necessarily need to be addressed right now.

src/containers/VisitPage.js
```diff
@@ -345,7 +320,7 @@ class VisitPage extends Component {
          * point. Perhaps a user's page should be a route
          * in VisitPage? Like our form sections will be? */}
```
knod: If you feel like it, you can probably update this comment ;)

src/containers/VisitPage.js
turnerhayes: The intent was to say "the way we currently store client data is in memory, so if you leave the page, it'll be gone". So the intent of this code is to redirect the user to the first step so that they can re-enter the data. I can try to reword it if it's not clear enough.

src/containers/VisitPage.js
turnerhayes: I had some issues with React Router throwing errors to the console when transitioning to the same page.

src/containers/VisitPage.js
turnerhayes: True; this is basically a sanity-check. Same with the next-step bit.

src/containers/VisitPage.js
knod: It's not blocking, so we can deal with it another time, but I found it unclear and rewording was what I was trying to describe.

src/containers/VisitPage.js
knod: Futureproofing, but maybe worth it. 👍

src/containers/VisitPage.js
knod: Ah, got it! A comment to that effect would probably be useful to someone trying to tighten up the code in the future.

src/components/StepBar.js
```diff
@@ -1,15 +1,17 @@
 import React from 'react';
 import { Step } from 'semantic-ui-react';
 
-const StepBar = ({ steps, currentStepIndex, goToStep, snippets }) => {
+import { STEPS } from '../constants';
```
knod, line 4: @todo for a later refactor - the name and location of `constants` is a big ambiguous. We have plenty of them around - a bunch for the benefit program logic, etc. These constants are specifically for form sections. They might find a better home with the form sections themselves and maybe a more specific name until we have something we need to add to the file.

src/components/MainMenu.js
```diff
@@ -13,8 +13,16 @@ const MainMenu = function ({ snippets }) {
       secondary
       size='large'>
 
-      <Menu.Item><Link className="main-nav" to="/">{ snippets.i_homeNav }</Link></Menu.Item>
-      <Menu.Item><Link className="main-nav" to="/about">{ snippets.i_aboutNav }</Link></Menu.Item>
+      <Menu.Item><Link
```
knod: Just noting that 'lint-staged' isn't working well here because I suspect this wasn't one of your staged files.

src/components/prompts/ReactRouterLeaveListener.js
```diff
@@ -40,14 +40,29 @@ class ReactRouterLeaveListener extends React.Component {
...
+      ) {
+        return true;
+      }
+    }
+
+    return `If you leave you'll lose this data. Are you sure you want to leave?`;
+  };
+
   render() {
     const { isBlocking } = this.props;
```
knod, line 58: Another option to think about might be to leverage this prop. The parent could determine whether the prompt should show beforehand, since it knows what step it's on anyway.

knod: @todo To consider for possible later refactor

src/containers/VisitPage.js
```diff
@@ -345,7 +320,7 @@ class VisitPage extends Component {
...
         { this.state.redirect ? (
-          <Redirect to={ `/detail/${this.state.clientInfo.clientId}` } />
+          <Redirect to={ `/detail/${this.state.clientId}` } />
```
knod: This is `this.props.clientId` now, right?

src/containers/VisitPage.js
```diff
@@ -360,17 +335,28 @@ class VisitPage extends Component {
             minWidth='874.5'
             style={{ padding: '14px 0' }}>
             <StepBar
-              currentStepIndex={ this.state.currentStep }
-              steps={ this.steps }
+              currentStepKey={ this.props.stepKey }
```
knod, line 334: From what I can tell, this could be `step.key` here, which might be more consistent with how it's being used further down.

knod: Non-blocking, @todo

src/containers/VisitPage.js
```diff
@@ -360,17 +335,28 @@ class VisitPage extends Component {
...
-            { this.getCurrentStep(navData) }
+            <StepComponent
+              component={ step.form }
+              snippets={ snippets[ step.key ] }
+              updateClientValue={
+                step.time === 'current' ?
+                  this.changeCurrent :
+                  this.changeFuture
+              }
+              currentStep={ stepIndex }
```
knod, line 348: @todo Is this still being used somewhere or can we get rid of it?

src/containers/VisitPage.js
```diff
@@ -360,17 +335,28 @@ class VisitPage extends Component {
...
+              currentStepKey={ this.props.stepKey }
               goToStep={ this.goToStep }
               snippets={ this.state.snippets.stepBar } />
           </Responsive>
           <div
             className="flex-item flex-column"
             style={{ padding: '14px 0' }}>
-            { this.getCurrentStep(navData) }
+            <StepComponent
+              component={ step.form }
```
knod, line 341: @todo Do the sections use their own component?

src/containers/VisitPage.js
turnerhayes: Yes! Good catch!

src/components/MainMenu.js
turnerhayes: I believe I may have run `npm run lint` which changed this. Not 100% sure.

src/components/StepBar.js
turnerhayes: Good point. We could have a single master constants file, but you're right that it might be better to have more specific constants modules.

knod (dismissed):
I've only looked through this thoroughly about four times, so I can't be sure I've caught everything, but it's looking good to me! Though some buttons can't be tested because of that bug.

Sorry I ended up doing so much thinking aloud in here. I will aim for better in the future.

knod (dismissed):
Lookin' good to me!

knod (approved):
Looking good other than a visual bug introduced earlier!

src/components/prompts/ReactRouterLeaveListener.js
knod: @turnerhayes : Did you have any thoughts on this?

turnerhayes: This is the only way I could find to get the destination location on navigation.

Issue #912, opened 2018-10-26
Implement a Redux store


turnerhayes:
If it ends up being too much of a barrier to new project members, we may later remove it.

Other alternatives:

- [Context API](https://reactjs.org/docs/context.html)
- [Hooks](https://reactjs.org/docs/hooks-intro.html)
  - Interesting bleeding-edge feature (currently in alpha); we'll have to see where it goes

knod:
Since it looks like I haven't put this down in writing yet, my suggestion is we start small and see how it goes.

PR #927, opened 2018-10-30
[WIP] Redux store

turnerhayes:
Addresses #912 

Still todo:
- [ ] Make visualizations use Immutable collections whenever possible
- [ ] Integrate Dev HUD with Redux


knod (commented):
Haven't gone through the rest of this yet, but is `immutable` a peer dependency of Redux? I think that lib was still up for discussion, though I may have forgotten a conversation.

package.json
```diff
@@ -5,15 +5,23 @@
   "dependencies": {
     "ajv": "6.5.4",
     "chart.js": "2.7.2",
+    "connected-react-router": "^4.5.0",
```
knod: Is this used?

wacii: It doesn't appear to be. It's used so that you can control react router with redux. The integration is kind of awkward otherwise.

wacii:
_Immutable.js_ isn't a peer dependency. But it's often used with it. And even if you don't use an immutability library, you end up programming in a way that accomplishes much the same thing.

wacii:
This is a tremendous amount of progress for a first pass.

package.json
turnerhayes: Ah yeah, sorry; I copied the store creation stuff from a project that does use it and `npm install`ed before removing that code. I'll get rid of this.

turnerhayes:
Like @wacii said, `immutable` is not *technically* a peer dep; immutability is kind of required for Redux. We *can* do immutability manually by deep cloning, but it's
a) easy to forget, especially for newcomers/people not familiar with Redux
b) leaves state objects open to (inadvertent) manipulation in code; anyone can push onto an array/set/delete a property, in ways that could potentially break code

It is a different way of coding and can take some getting used to, so we can definitely discuss this more if you're not comfortable with it.

PR #932, opened 2018-10-31
Replace var with let

### LINKED ISSUE #899: Convert var to const/let

turnerhayes:
Fixes #899.


knod (approved):
I like it! Should we press pause on other things while these conflicts get resolved, or shall we keep forging ahead?

PR #934, opened 2018-10-31
Change import path of <ExternalLink> Fixes #701

### LINKED ISSUE #701: Move `<ExternalLink>` into src/components
It may be good to double check this, but I think form components aren't the only ones that use that component.

MRLmic:
ExternalLink imports moved from formHelper export to src/components


knod (approved):
Looking great! Feel free to merge when you want or someone will merge when we get a second eye on it.

PR #944, opened 2018-11-01
Moves 'contants' and renames file/var. Fixes #943

### LINKED ISSUE #943: Move 'constants.js' to more specific location
From https://github.com/codeforboston/cliff-effects/pull/922#pullrequestreview-169101048

knod:
Previous name sounded like list of components to me. Open to other thoughts.

### DIFF
--- a/src/forms/STEP_PROPS.js
+++ b/src/forms/STEP_PROPS.js
@@ -1,10 +1,10 @@
-import { CurrentIncomeStep } from './forms/CurrentIncome';
-import { CurrentExpensesStep } from './forms/CurrentExpenses';
-import { PredictionsStep } from './forms/Predictions';
-import { HouseholdStep } from './forms/Household';
-import { CurrentBenefitsStep } from './forms/CurrentBenefits';
+import { CurrentIncomeStep } from './CurrentIncome';
+import { CurrentExpensesStep } from './CurrentExpenses';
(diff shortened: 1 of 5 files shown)


dylanesque:
This is definitely clearer than the previous version.

src/forms/STEP_PROPS.js
```diff
@@ -1,10 +1,10 @@
...
-import { HouseholdStep } from './forms/Household';
-import { CurrentBenefitsStep } from './forms/CurrentBenefits';
+import { CurrentIncomeStep } from './CurrentIncome';
+import { CurrentExpensesStep } from './CurrentExpenses';
+import { PredictionsStep } from './Predictions';
+import { HouseholdStep } from './Household';
+import { CurrentBenefitsStep } from './CurrentBenefits';
 
-export const STEPS = [
+export const STEP_PROPS = [
```
turnerhayes: I would avoid using "props" here since it has a specific meaning in React-land, and suggests these are meant to be passed as component props. Why do you feel `STEPS` is insufficient?

knod: Fair point. `STEPS` feels like a list of components to me, but it's more than that. Any suggestions? `STEP_VALS`? `STEP_PARTS`?

turnerhayes (approved):

PR #949, opened 2018-11-01
Change localization data to object clone instead of object reference

### LINKED ISSUE #947: .langCode incorrectly displaying in each section or localization report
Report contains an entry like the following:
"aboutPage.langCode should exist."

`.langCode` exists as a key at the top of the localization object only.  It's being treated as if it's a key in every object.

jlangevin:
Fixes #947 .  Also moves List.Item key to ReportListItems to fix duplicate keys warning.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -88,10 +88,6 @@ class App extends Component {
       },
       distrustConfirmed: false,
     };
-
-
-
-    console.log(getTextForLanguage(`zh`));
   };  // End constructor()
 
   /**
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -88,6 +88,10 @@ class App extends Component {
       },
       distrustConfirmed: false,
     };
+
+
+
+    console.log(getTextForLanguage(`zh`));
```
knod: I've done this a million times myself

jlangevin: Ah, that was leftover from our call this morning.  Sorry about that.  Good catch.  Thanks!

knod (approved):
A bit of a heavy PR, but the changes are lookin' good to me!

PR #962, opened 2018-11-03
Fixes #961 - different abstraction for user message, take 1

### LINKED ISSUE #961: Different abstraction for user error/invalid messages?
Looking at `CashFlowRow` in src/forms/cashflow.js, the place 'invalid' message has ended up seems tangled unnecessarily (mostly my doing, I suspect). I'd like to look at other ways to abstract it that make more sense.

knod:
#961


src/forms/formHelpers.js
```diff
@@ -1,33 +1,75 @@
...
+ * @param {string|object|null} [message] To be rendered above
+ *     the children.
+ * @param {boolean} [isUserError] Is this an error message?
+ *     Another option would be `type` instead - descriptor(s)
+ *     for the message that will currently be turned into
+ *     classes. Things like 'invalid'.
+ *
+ * @returns React element
+ */
+const WithMessageAbove = function ({ children, ariaName, isUserError, message, className }) {
```
pporfilio: I like that you've added the ariaName to make it easy to associate with another node.

src/forms/formHelpers.js
```diff
@@ -1,33 +1,75 @@
...
-    result = (
+  let messageElem = null;
+  if (message) {
+
+    let messageClasses = `message-above `;
+    if (isUserError) {
+      messageClasses += `invalid`;
+    }
+    // `id` is for aria description hook
+    // tabIndex is for
```
pporfilio: I think you meant to add a note about internet explorer here.

pporfilio (approved):

src/forms/formHelpers.js
knod: lol, thanks for catching that

turnerhayes:
Looks good, but I wanted to put a discussion question out there; is this premature abstraction? It seems it's only currently being used in one file; are we sure this is something that's going to be reused?

turnerhayes (approved):

knod:
We do plan to re-use it as we move forward. Also, It seemed like a more concrete abstraction to me. We also discussed that it could use a better name.

PR #963, opened 2018-11-04
Fixes #945 #6 last item, switch var name

### LINKED ISSUE #945: Resolve/clean VisitPage questions a bit
Specifically things in:
- [x] 1. https://github.com/codeforboston/cliff-effects/pull/922#pullrequestreview-169109576 (use step.key)
- [x] 2. https://github.com/codeforboston/cliff-effects/pull/922#pullrequestreview-169109728 (`currentStep` attr needed?)
- [x] 3. https://github.com/codeforboston/cliff-effects/pull/922#pullrequestreview-169109853 (give component to itself?)
- [x] 4. https://github.com/codeforboston/cliff-effects/pull/922#pullrequestreview-169069284 (explanatory comment about why first step can't go to first step) (seems to have been changed)
- [x] 5. Scroll in `.goToStep()` instead of in prev/next step funcs
- [ ] 6. `step.form` name to `step.component`

knod:
It's not a big thing, I'm just fiddling around brainstorming about other stuff and it was low-hanging fruit. Not all the sections are really forms. One is a form and results, and we're still figuring out if we want to keep it that way.

### DIFF
--- a/src/containers/VisitPage.js
+++ b/src/containers/VisitPage.js
@@ -277,7 +277,7 @@ class VisitPage extends Component {
 
     const step = STEP_VALS[ stepIndex ];
 
-    const StepComponent = step.form;
+    const StepComponent = step.component;
 
     return (
       <div className='forms-container flex-item flex-column'>
--- a/src/forms/STEP_VALS.js
+++ b/src/forms/STEP_VALS.js
@@ -6,28 +6,28 @@ import { CurrentBenefitsStep } from './CurrentBenefits';
 
 export const STEP_VALS = [
   {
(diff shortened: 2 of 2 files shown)


knod:
Rather, we probably want to change it. Maybe a better name than `component` would be `section`, but `component` just seemed to most clearly express what it was.

dylanesque (approved):

dylanesque:
> Rather, we probably want to change it. Maybe a better name than `component` would be `section`, but `component` just seemed to most clearly express what it was.

I concur, component is definitely an improvement over form

PR #965, opened 2018-11-04
Rearrange and split functions in Expenses

knod:
For testing and maintenance. Splitting the file itself is an option, but I'm not sure there's a reason to. None of the components need to be used anywhere else. Something like splitting into two scripts so that the name of the file isn't so misleading might worth it, though.


knod:
Can't be really checked out till #967 is resolved.

knod:
imo, the diff is probably best viewed one commit at a time.

Also, the styles are messed up right now, but that's because of an unrelated issue that's fixed in another PR.

knod:
Not WIP anymore

src/forms/CurrentExpenses.js
```diff
@@ -41,203 +41,323 @@ import { getUnder13Expenses } from '../utils/cashflow';
...
 
-  if (current.housing === 'voucher') {
-    return (
-      <div>
-        <ContractRentField { ...sharedProps } />
-        <RentShareField { ...sharedProps } />
-        <Utilities { ...sharedProps } />
-      </div>
-    );
+const ElderlyOrDisabled1 = function ({ current, type, sharedProps, updateClientValue }) {
```
turnerhayes: Can we give this a more descriptive name than suffixing it with "1"? Maybe something like `ElderlyOrDisabledAssistance` and the component below as `UnreimbursedMedicalExpenses`?

src/forms/CurrentExpenses.js
```diff
@@ -298,272 +418,219 @@ const Housing = function ({ current, type, time, updateClientValue }) {
...
 
+  } else {
+    return null;
+  }
 
-export { CurrentExpensesStep };
+};  // Ends <EarnedFrom>
+
+
+export {
```
turnerhayes, line 623: Is any other file actually making use of any of these other components? Do we need to export anything other than the top level component?

src/forms/CurrentExpenses.js
```diff
@@ -41,203 +41,323 @@ import { getUnder13Expenses } from '../utils/cashflow';
...
-    if (evnt.key === `Enter`) {
-      evnt.target.click();
-    }
-  };
+};  // Ends <ExpensesFormContent>
 
-  // May want to change name to 'utilities' and value to what's 'name' now
-  // Will require more work in the change handler
+
+const Under13 = function ({ snippets, type, sharedProps, current, updateClientValue }) {
```
turnerhayes, line 202: I think `sharedProps` is a bit too vague; I think it would be be better to have the component actually list the props that it needs to use. The consumer of the component can still use a shared props object by calling `<Under13 {...sharedProps} />`, but the component itself should specify the minimal set of information it needs (especially if we add propTypes in the future).

This goes for the other components in the file, too.

turnerhayes (approved):
I'll open another task for the `sharedProps` prop (see [comment](https://github.com/codeforboston/cliff-effects/pull/965/files#r232471172)), unless you decide you want to address it in this PR. Same for the extraneous exports.

src/forms/CurrentExpenses.js
knod: Oh, sorry, forgot I left those in. I'll look at them and try to figure out useful names.

src/forms/CurrentExpenses.js
knod: I'd like to be able to export them for testing. Is that not the way?

src/forms/CurrentExpenses.js
knod: If it's not already broken out in the code below, we should definitely do so. I'm not sure about cluttering the function parameters more. We can discuss it, though. @todo

src/forms/CurrentExpenses.js
knod: Yeah, those are completely on point 👍

src/forms/CurrentExpenses.js
knod: @todo

turnerhayes: I would test the exported component only; if we use `mount`, we should be able to find the necessary components. These seem like internal components; essentially an implementation detail that should not be part of the module's API.

knod: Hmm, maybe worth a proper discussion. Been considering moving the logic of whether to show stuff or not into the individual components themselves and that would make stuff easier to test individually, etc.

PR #968, opened 2018-11-04
Fixes #967, 'Household' section crashing and burning

### LINKED ISSUE #967: 'Household' section burning
### Replicate:
1. Go to 'Household' section

### Result I'd Expect:
- No error

### Actual Result:
- Error

### Environment:
App version or commit id at the time: 71c168723255c7abb67d9d4c96573ddbb5696bbb

### Other Notes
We really have to switch to non-shallow tests. Thought I was covered on this one!

knod:
Sorry folks, this was my bad. Head's underwater somewhere again.


src/forms/inputs.js
```diff
@@ -184,6 +184,12 @@ class ManagedNumberField extends Component {
     var { valid, focused, focusedVal }                = this.state;
     var { value, name, className, format, otherData } = this.props;
 
+    // @todo Bad var name. Find more useful name for this.
+    let id = name;
```
pporfilio, line 188: Maybe something like `constructedId` or `augmentedId`?

pporfilio (approved):
This appears to resolve the crash on the Household section and seems like a nice change to not require an interval to be specified.

src/forms/inputs.js
knod: Oh, sorry. That wasn't specific and was put in the wrong place. I meant `interval`. Something like `idSuffix` might be useful for that. @todo

Issue #966, opened 2018-11-04
Replace _all_ `shallow()` test renderings with `mount()`


knod:
Other than the shallow tests not catching bunches of stuff, I think we may find some things not right with our components.

knod:
Or do `.not.Throw()` instead. One or the other.

turnerhayes:
I've read the opposite advice; use `shallow()` unless you need something that `shallow()` can't do. One reason for this is that it can help you avoid needing to set up context/etc. for nested components--for example, React Router's `<Link>` component requires a router in its ancestry, and will throw an error if it doesn't find it. But if you're testing a component that contains a `<Link>` but the bit you're testing doesn't need to render the link, you can avoid that. In general, I think the idea is to only test the specific case you're testing.

knod:
I hear some of that. Can't we add context and routers to our test, though? Here's an article about shallow rendering that seemed to resonate: https://blog.kentcdodds.com/why-i-never-use-shallow-rendering-c08851a68bb7

PR #969, opened 2018-11-04
`shallow()` to `mount()` for some tests as much as possible, #966

knod:
`<CashflowInputsRow>` needs `shallow()` for some bits.


src/test/forms/cashflow/CashFlowInputsRow.test.js
```diff
@@ -1,63 +1,75 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { mount, shallow } from 'enzyme';
```
turnerhayes, line 2: Out of curiosity, why not make the last test into `mount()` too?

turnerhayes (approved):

src/test/forms/cashflow/CashFlowInputsRow.test.js
knod: All I remember is that it was hard to resolve the strangeness that was happening. I'm not super familiar with enzyme, though, so maybe it's really a quick fix. I encourage you to try it out!

Issue #892, opened 2018-10-24
Spouse/member dropdown not fully changing on selection of item


knod:
### Replicate:
1. Go to the 'Household' section
2. Add a member
3. Click on the drop down
4. Click on 'Child/Other member'
5. Click on the drop down again
6. Click off of the drop down

### Result I'd Expect:
- 'Child/Other member' to appear as the chosen option

### Actual Result:
- 'Spouse' is the text that appears as the selected option

### Environment:

App version or commit id at the time: 04d75ab296d8648456fdb4fecc5bb84be241f64f

Issue #929, opened 2018-10-30
package-lock.json with npm


knod:
Removing all the carats didn't do it. How do we deal with this?

dylanesque:
Not sure I cross-posted the thread from Boston JS I started about Yarn/package-lock.json issues, but doing that now. Bullish on Yarn being our solution for the headache of keeping everybody's ~~npm~~ Yarn version in sync, but I seem to remember other helpful suggestions in there.

dylanesque:
One possibility is ```npm ci```: https://docs.npmjs.com/cli/ci

knod:
Below is the thread from Boston JS. From what I can tell, it says yarn seem like it probably works for the p-l issue if you've got a lockfile, but
   1. It requires the whole team to have the same yarn version
   1. It's hard to make sure new members aren't using npm instead (though `.gitignore`ing the p-l helps)
   1. Some editor plugins assume you're using npm so they'll freak out if you're using yarn

For npm, removing carats is supposed to work, but we've tried that and it hasn't solved the issue.

Have I missed anything crucial?

09/14/18
Person1
I just got involved with an open-source project, and learned about this ongoing bug with package-lock.json files in npm where they defer to the package.json files and dont' stay, um, locked. Is anybody using Yarn instead of npm? I've used it and it seems fine, but are there any major drawbacks or gotchas in Yarn vs npm?

Person2
I've used yarn pretty much since it came out, it works for me.  The greatest complement I can give it is that I can't remember the last time I have had to delete node_modules and reinstall to get things working again.

Person3
that sounds too good to be true

Person4
Yeah, I can echo what Person2 said
I've been using yarn for about 1.5 years
same

Person2
if there is a drawback, I would say that because it an external dependency that you need to make sure everyone keeps it up to date

Person4
Interestingly, I've seen some people get around that by having a versioned copy of yarn in their repo
that way you can do utils/yarn.js [command]

Person2
hmm thats a good idea

Person4
sorta hacky and it bloats the repo but it works!

Person5
wouldn't yarn behave the same? the `package.json` is still the source of truth no? I know that yarn has a `--frozen-lockfile` option you can pass that will cause the install to fail if it detects a change to the lockfile.

	Person1
	I did not know this, and glancing at the npm docs, they don't seem to have an equivalent feature

	Person5
[...]

PR #977, opened 2018-11-07
solution to Spouse/member dropdown not fully changing on selection of item #892

josephdlawson21:
This is my solution to issue #892. there was an issue with Semantic UI and i replaced the Dropdown component with the react-select package. had to make a new callback because the data returned from the dropdown was different.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "react-router": "4.2.0",
     "react-router-dom": "4.2.2",
     "react-scripts": "1.1.5",
+    "react-select": "^2.1.1",
     "semantic-ui-css": "2.2.12",
     "semantic-ui-react": "0.75.1"
   },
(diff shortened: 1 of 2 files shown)


knod:
Sorry we misunderstood the situation yesterday. I think for this we _should_ update the package-lock.json since the package.json is actually changing. It's a shame we have to add a whole new lib just for this one component. You said semantic-ui-react has an issue for this. Maybe we can link this lib in there. It may help them along.

Other than that, the style doesn't match ours, but that can probably be tweaked later.

knod:
From what I can see, an inaccurate p-l is the reason the tests are failing. Which may be another note to add to #929.

knod:
@turnerhayes , @dylanesque , or anyone else, this one needs to be taken up and completed.

knod:
Taken over by #1034

PR #978, opened 2018-11-07
Redirect before rendering VisitPage. Fixes #972

### LINKED ISSUE #972: Is there a way to 'redirect' before the whole VisitPage renders?
Whenever we go to the Current Benefits section or refresh the page while in the form part of the site, it looks like the `VisitPage` might be rendering and _then_ 'redirecting' and rendering the first section. Can we intercept it before the render? The 'redirect' is currently happening in `VisitPage` in `.componentDidMount()` by replacing history.

At worst, could we just do an `if` statement to prevent rendering if the history doesn't have the right section?

turnerhayes:


knod (approved):
Hmm. A little translucent, but I think it does what we need.

knod:
#972 isn't showing a reference to this, which is strange.

PR #979, opened 2018-11-07
Add .nvmrc file. Fixes #931

### LINKED ISSUE #931: Use .nvmrc for node version
As mentioned in https://github.com/codeforboston/cliff-effects/pull/824#issuecomment-433199711 we might want to use an .nvmrc file rather than piping file contents into `nvm`.

turnerhayes:
Test plan:

- [Install NVM](https://github.com/creationix/nvm#installation)
- Run `nvm install` from the project root directory
- It should install and use the latest LTS version of Node (at the moment, this is `v10.13.0`)
- Open a new terminal and go to the project directory
- Run `nvm use`
- It should switch to use the LTS version (verify that it is with `node --version`)


.nvmrc
```diff
@@ -0,0 +1 @@
+lts/*
```
knod: Hmm, I think [we](https://github.com/codeforboston/cliff-effects/pull/824#issuecomment-426595109) wanted to specify an actual version number. Would that be acceptable?

knod (changes requested):

knod:
Also, is there a way to add a pre-... `npm install` hook to make sure this happens for new folks right off?

.nvmrc
turnerhayes: Ah, missed that comment. Will fix!

turnerhayes:
That might be trickier, because it depends on `nvm` being installed, and as far as I know there's not a good way to do that from a script--I think they will have to install nvm first (if they don't have it).

*Maybe* there's a way to detect if `nvm` is installed and run the necessary commands only in that case?

knod:
Dang, forgot about that part. Hmm. Maybe a warning if they're using the wrong version of node/npm?

turnerhayes:
That can be done; the problem is that `create-react-app`'s `start` script clears the console, so at most you'd see the warning appear for a couple of seconds and that might not be enough to register it.

knod:
I was actually thinking of doing it at other times. For example, when they first clone the repo, when they do `npm install`, when they push. It seems like there's a lot of different options.

knod (approved):
Sorry, didn't see this change! The discussion about a script to warn folks can happen elsewhere. @todo!

Issue #952, opened 2018-11-01
Client data vs. form data?


knod:
So far, we use client data to determine what to show in the form and client data is supposed to contain just the data for the calculations. We're slowly getting into territory where we many need non-calculation crucial data in order to choose what to show in the forms.

How do we want to handle that?

dylanesque:
Could you elaborate on this? What is an example of crucial data that that affects form output?

knod:
One example is the 'other expenses' questions we've partially implemented. It doesn't affect benefit calculation results, it just affects what the form shows at the end.

PR #981, opened 2018-11-08
Convert benefits to a list.

### LINKED ISSUE #964: Turn client data current benefits into a list/array of benefit names
To be most flexible for adding future benefits, it'd work better to build a list of what benefits the person has.

turnerhayes:
Fixes #964


src/forms/output/BenefitsLineGraph.js
```diff
@@ -48,8 +48,12 @@ class BenefitsLineGraph extends Component {
         interval  = Math.ceil((max / 100) / 10) * 10;
 
     var xRange      = _.range(limits.min, max, interval),  // x-axis/earned income numbers
-        extraProps  = { snap: { fill: false }, section8: { fill: false }},
+        extraProps  = {},
         datasets    = getChartData(xRange, multiplier, client, activePrograms, extraProps);
+    
+    for (let benefitIndex = 0; benefitIndex < activePrograms.length; benefitIndex++) {
+      extraProps[ activePrograms[ benefitIndex ] ] = { fill: false };
```
knod, line 53: This seems not to be working currently.

Note for future: May be worth abstracting some of the specific benefit chart frostings. @todo

src/localization/en.js
```diff
@@ -122,12 +122,11 @@ export default {
...
-      selectBenefits_v1:   `Select the benefits you currently receive`,
-      hasSection8Label_v1: `Do you have Section 8 Housing?`,
-      hasSection8Hint_v1:  `Section 8 provides rental housing assistance.`,
-      hasSnapLabel_v1:     `Do you have SNAP?`,
-      hasSnapHint_v1:      `SNAP provides assistance with buying food`,
+      currentBenefits_v1:      `Current Benefits`,
+      selectBenefits_v1:       `Select the benefits you currently receive`,
+      has_section8_label_v1:   `Do you have Section 8 Housing?`,
+      has_snap_label_v1:       `Do you have SNAP?`,
+      has_xrayVision_label_v1: `Do you have X-ray vision?`,
```
knod: lol! Maybe we should keep it in, though. I kind of like it. It'll be an easter egg for our fans :D

knod: And it'll show up on the localization report! I love it!

knod (changes requested):
Some awesome work here!

One note I couldn't add is that the `updateClientValue` needs to be changed too. The way it's setting shelter and housing stuff is weird and this messes it up. You'll see when you look at it. Let me know if you need another eye on it since it's a bit wacky. Generally searching the code for `hasSNAP` and `hasSection8` will probably reveal any stray problems.

Other than those two things, I _think_ we're good...

knod:
Btw, that's the kind of problem I think figuring out #952 would help.

src/localization/en.js
turnerhayes: Haha, damn it, I added that to test arbitrary numbers of benefits. I thought I had removed it!

src/forms/output/BenefitsLineGraph.js
turnerhayes: What effect is this supposed to have?

turnerhayes: Ah, nevermind, I see it

src/localization/en.js
```diff
@@ -122,11 +122,10 @@ export default {
     },
 
     currentBenefits: {
-      currentBenefits_v1:      `Current Benefits`,
-      selectBenefits_v1:       `Select the benefits you currently receive`,
-      has_section8_label_v1:   `Do you have Section 8 Housing?`,
-      has_snap_label_v1:       `Do you have SNAP?`,
-      has_xrayVision_label_v1: `Do you have X-ray vision?`,
```
knod: Dangit

knod (approved):
So because of my own past mistakes the benefit orders are wrong, but that's on me. The work in this PR is lookin' good to me 👍

Issue #956, opened 2018-11-03
New chart library/new charts


knod:
So far two have been suggested:
1. Highcharts with a non-commercial license: https://shop.highsoft.com/faq#Non-Commercial-0
2. Chartist.js https://gionkunz.github.io/chartist-js/index.html

Highcharts looks pretty swank and quite worth checking out, imo (thanks @turnerhayes)

- [x] Line graph
- [x] Stacked area graph
- [x] Bar chart

Wanted:
1. Zooming/panning (chartjs doesn't).
2. Overlaying multiple types of charts on each other (chartjs does).
3. Ability to stay the same size even when labels text increases in length, like when switching between weekly, monthly, and yearly amounts (chartjs doesn't).
4. Accessibility would be awesome (chartjs probably doesn't).
5. Also printing, though that might be a bit messy depending on if we want to be able to have an option to print multiple result types (chartjs probably doesn't manage that itself).

knod:
For Highcharts, I've found:
1. https://github.com/whawker/react-jsx-highcharts
2. https://github.com/kirjs/react-highcharts

knod:
[Using react-jsx-highcharts]

~So far, one issue with highcharts is its width is annoying. It only takes pixel or percent values. A percent value would mean us really digging deep to make sure there's a solid width property for the chart's parent container. A pixel value means less responsiveness. We can add some responsiveness back in with their responsive settings, but that's limited. We can also add a script that will dynamically update the pixel width of the chart when the component is mounted and when the window is re-sized (it's been done before, though I'd have to dig up where I found it), but that too is a pain.~

~Height has the same problem, but I'm not as concerned about that right now.~

~[Hmm, width with highchart itself actually seems responsive: [https://jsfiddle.net/1p7s5yw6/](https://jsfiddle.net/1p7s5yw6/). I'm not sure what's going on with the react package.]~

So no idea what was going on because this wasn't working before, but I now, for the 10th time, have no width set whatsoever and the container style as `width: 100%;` and its size is adjusting just fine :P

knod:
[Using react-jsx-highcharts]

Things that should be dug into more:
- [x] 1. Controlling number of tick marks and points on the lines better. [Discovery: if there are too many close together, it decreases the number of them automatically.]
- [x] 2. Hover/tap to see details about a point [not completely sure about tapping]
- [x] 3. Same math/resizing problem as our current version when the time interval is changed. [`endOnTick={false}`!]
[...]

Issue #546, opened 2018-05-31
UI bug: graph titles in summary


ohmegasquared:
The graph titles in the summary page is center aligned.  This can bleed into the day/month/year toggles. Thinking of i18n, titles like Individual Benefit Amounts for Housing as Income Changes will be longer and should either be left aligned or have a set width and allow for multiple lines.

knod:
I think this may be a issue about the site's size responsiveness, which hasn't been addressed yet. It's not currently in our scope, but if someone feels that that would be their main interest in contributing, it would be a great addition to the project.

ethanbb:
Just want to add that the concern wasn't just about responsiveness, but also that the text may be longer when translated into other languages. That being said we haven't seen that be a problem yet, but it could be soon.

knod:
Ah, got it. We could think about rethinking/restructuring those elements. It's a complex situation because we don't really control the chart stuff and the time buttons are basically an overlay on top of it. We'd need to re-design the look of it so that that was taken into account.

I'm concerned that there's so much to do right now that this might need to be put on the back-burner for the time being. If someone has a specific re-design in mind, that could help.

knod:
We may be changing our chart lib, so we'll see what can happen with that.

knod:
For now, we've done what work we can while keeping the time interval buttons. We've attempted to make adjustments for different screen sizes. Those measures will have to be re-done if layout changes because there doesn't seem to be a reasonable way to add buttons to the Highcharts element itself, but until some method for that becomes clear this is the best we can do.

PR #984, opened 2018-11-10
New charts lib for Benefits Lines graph, #956. Feedback welcome

knod:
Building it up.


knod:
Not a WIP anymore! Ready for review!

package.json
```diff
@@ -5,12 +5,14 @@
   "dependencies": {
     "ajv": "6.5.4",
     "chart.js": "2.7.2",
+    "highcharts": "^6.2.0",
```
knod, line 8: Not sure whether to mess with this caret or not. This is what the react-jsx-highcharts instructions said to install.

package.json
```diff
@@ -5,12 +5,14 @@
...
     "lodash": "4.17.10",
     "moment": "2.22.1",
     "react": "16.3.2",
     "react-chartjs-2": "2.7.2",
     "react-dom": "16.3.2",
     "react-helmet": "5.2.0",
+    "react-jsx-highcharts": "^3.3.1-alpha.1",
```
knod, line 15: Not sure why `npm install...` pulled in an alpha version. Not sure what to do about that either.

knod:
Also helps with #546. Doesn't actually fix it, dangit. Just makes it more likely to be less of a problem. Handling the display based on the width of an actual title is a much trickier proposition.

src/index.css
```diff
@@ -240,15 +240,49 @@ input[type=number]::-webkit-outer-spin-button {
 .graph-holder {
   position: relative;
   min-height: 7rem;
-  display: flex;
-  justify-content: flex-end;
-  align-items: flex-end;
+  /*display: flex;*/
```
turnerhayes: Can you remove these lines rather than comment them out?

src/test/forms/output/GraphHolder.test.js
```diff
@@ -9,7 +9,7 @@ const passedActivePrograms = (wrapper) => {
   return wrapper.find('Graph').prop('activePrograms');
 };
 
-describe('<GraphHolder>', () => {
+describe.skip('<GraphHolder>', () => {
```
turnerhayes, line 14: Is this supposed to be skipped? If so, could you add a comment for why it's being skipped?

turnerhayes (approved):
Nothing here that blocks the PR; mostly formatting/comments. It would be great if you wanted to address them here, but if not we can have a separate issue to address them.

src/test/forms/output/GraphHolder.test.js
knod: Great idea! Done!

src/index.css
knod: Thanks!

package.json
knod: @todo

package.json
knod: @todo

PR #992, opened 2018-11-12
Add node version checking script. Fixes #983

### LINKED ISSUE #983: Warn coders when they have the the wrong version of node/npm
As discussed in #979.

turnerhayes:
TBD: all the places to run this


turnerhayes:
For some reason, this didn't seem to show any link to #983

turnerhayes:
Another question to consider; do we want to do exact version match? Or accept higher versions/different minor/patch versions? That is, if our current node version is set to `v10.13.0`, should we warn if the user is on `10.13.4`? `10.15.0`? `12.0.0`?

turnerhayes:
Added check to pre-commit and lint. Keeping version checking to exact match.

check-node-version.js
```diff
@@ -0,0 +1,14 @@
...
+const path = require('path');
+
+const targetVersion = fs.readFileSync(path.resolve(__dirname, '.nvmrc'), { encoding: 'utf8' })
+  // strip trailing and leading whitespace
+  .replace(/\s*$/, '').replace(/^\s*/, '');
+
+if (process.version !== targetVersion) {
+  const separator = '===========================================================';
+  console.warn(
+    separator + '\nUsing the wrong Node version; expected ' + targetVersion + ', got ' + process.version +
```
knod, line 11: @todo Maybe we make a wiki page about where to find info for updating and add a link to that in this message so people have a next step to take if they have the wrong version?

knod:
It's weird. I'd think on `commit` you'd get two messages, since it also lints, but you only get one.

knod (approved):
Pulled and tested and seems to be working for me! I love it!

turnerhayes:
I think that's because the pre-commit hook runs `eslint --fix`, rather than `npm run lint`

PR #995, opened 2018-11-13
Fixes #953, shifts focus of About page, notes on future fixes

### LINKED ISSUE #953: Edit `AboutPage` focus
It currently sounds like everyone hits an implicit tax over 100% (a dramatic cliff), but not everyone does. Right now, a lot of people are throwing around the idea that the safety net doesn't work. Even though it's got flaws, not every situation is a disaster.

This line has been suggested to be most problematic: “The household is working to earn more, but they end up further behind. These cliff effects stop a lot of families from actually getting off of public assistance programs.”

While case managers do seem excited by the name (maybe because it directly talks about what they're worried about), we don't want to be alarmist. It's been suggested that at least in our description we focus on the fact that we're helping with the uncertainty about the effect pay changes have on benefits.

knod:
Links still need to be removed/edited. If this is merged, the google docs will need to be updated.


src/localization/en.js
```diff
@@ -54,11 +54,13 @@ export default {
     ],
 
     whyHeader_v1: `Why is this tool important?`,
-    why1_v1:      `A cliff effect occurs when a slight change in a household’s circumstances - say, a slight pay raise - disproportionately lowers their benefits. The household is working to increase what they earn, but they end up with a net loss that actually puts them further behind. These cliff effects prevent many families from actually getting off of public assistance programs.`,
-    why2_v1:      `Cliff effects are also difficult to predict. The interactions between income, household size, many other criteria, as well as the effects of the programs themselves impact each other in unexpected ways. We're exploring ways to deal with this issue of complexity and help families better understand and predict their situation.`,
+    why1_v2:      `All the ways that benefits interact with a family's circumstances and with each other are impossible fo one person to calculate accurately. A family making fanancial choices these days is making them blind. We're exploring ways to deal with this issue of complexity and help families better understand and predict their situation.`,
```
turnerhayes: "are impossible fo one person" => "are impossible for one person"

src/localization/vi.js
```diff
@@ -45,7 +46,9 @@ export default {
     ],
 
     'whyHeader_v1': `Tại sao công cụ này quan trọng?`,
+    // v2: `All the ways that benefits interact with a family's circumstances and with each other are impossible fo one person to calculate accurately. A family making fanancial choices these days is making them blind. We're exploring ways to deal with this issue of complexity and help families better understand and predict their situation.`
```
turnerhayes: (see above comment on English localization file)

src/localization/zh.js
```diff
@@ -45,7 +46,9 @@ export default {
     ],
 
     'whyHeader_v1': `這個工具爲什麽重要？`,
+    // v2: `All the ways that benefits interact with a family's circumstances and with each other are impossible fo one person to calculate accurately. A family making fanancial choices these days is making them blind. We're exploring ways to deal with this issue of complexity and help families better understand and predict their situation.`
```
turnerhayes: (see above)

turnerhayes (changes requested):
Mostly fine; just one typo to fix.

knod:
Should be better now!

turnerhayes (approved):
Looks good!

PR #996, opened 2018-11-14
Updates de.js

ValerieKenyon:
Brings de.js up-to-date with en.js.  It's been a while since I've done one of these, I hope I did it okay.

### DIFF
--- a/src/localization/de.js
+++ b/src/localization/de.js
@@ -6,13 +6,13 @@ export default {
   header: {},
 
   footer: {
-    header: [
+    header_v1: [
       {
         name: `__githubRepoLink__`,
         text: `Cliff Effects Werkzeug`,
       },
     ],
-    cfbCredit: [
+    cfbCredit_v1: [
       `Mit `,
       { name: `__heartIcon__` },
       ` von Code for Boston gemacht`, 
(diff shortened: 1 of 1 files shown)


knod (approved):
I don't know German! It looks great!

knod:
None of this needs to block the merge. This is all fine. Unless I've missed typos, in which case it's my bad.

From #997, @ValerieKenyon says:
>(BTW: I was working from en.js in the google docs, updating to de.js in the google docs, and then creating a new, up-to-date branch. I copied and pasted my German from the google doc into the new branch and tweaked it in VS Code. I also noticed that the en.js in dev branch (and in germanTranslation18_11_13) has a lot more text in it than the en.js google doc, so my updated de.js doesn't actually contain this newer text.)

I can't tell exactly what you mean, but it sounds like the google doc is out of date, so someone needs to go through and update it carefully. There are about to be more changes to it, so after the next couple merges, which should happen in just a few minutes, anyone can feel free to take that on. There's actually an issue for that already.

For English, we can basically just cut and paste the local changes into the google doc. The updates to the other languages need to happen more carefully.

@ValerieKenyon : Is that something you'd be interested in doing/have time to do?

PR #1000, opened 2018-11-17
Dev utilities. Fixes #999

### LINKED ISSUE #999: Persist loaded client
Allow loaded clients to be persisted locally for a time so that developers don't have to copy and paste clients into the HUD each time they reload

turnerhayes:
- Persist loaded client for 1 day
- Add command line utilities for viewing client (read-only) and
enabling Dev HUD


turnerhayes:
Open to discussion about appropriate timeout lengths.

Also if we're really worried, maybe look into using a crypto library so that people can't make anything of the data if they just have access to local storage? But I don't think that is actually a reasonable concern; I'm not sure how someone would have access to the site's local storage without having access to the app.

Let me know what you think about the command line utilities; I did them on a whim and I think they could be handy but I know we didn't discuss them beforehand.

turnerhayes:
Fixes #999 .

knod (approved):
I'm still a little leery of keeping the data, but it does drive me absolutely crazy to put it in again each and every time. The changes you've made from our discussions look good.

I declare this functional and useful!

PR #1005, opened 2018-11-19
Implements bar graph. Fixes #956.

### LINKED ISSUE #956: New chart library/new charts

knod:
Cleans up/clarifies some related stuff. Also changes settings for
the graph holder.

Will now need to clean up from having the old charts.

### DIFF
--- a/src/forms/output/BenefitsLines.js
+++ b/src/forms/output/BenefitsLines.js
@@ -131,7 +131,8 @@ class BenefitsLinesComp extends Component {
       lines.push(line);
     }
 
-    // Get 'Unexpected template string expression' warning otherwise
+    // Label for split tooltip 'labels'/'label headers' that appear
+    // at the bottom. Really long.
     // @todo Abstract commonalities between graphs
     const labelHeaderFormatStart = `<span style="font-size: 10px">${getText(snippets.i_beforeMoney)}`,
           labelHeaderFormatEnd   = `{point.key:,.2f}${getText(snippets.i_afterMoney)}</span><br/>`,
--- a/src/forms/output/ResourcesColumns.js
+++ b/src/forms/output/ResourcesColumns.js
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { range } from 'lodash';
+import { cloneDeep } from 'lodash';
 
 // HIGHCHARTS
 import Highcharts from 'highcharts';
@@ -11,8 +11,7 @@ import {
   Tooltip,
   XAxis,
   YAxis,
-  PlotLine,
-  AreaSeries,
+  ColumnSeries,
   withHighcharts,
 } from 'react-jsx-highcharts';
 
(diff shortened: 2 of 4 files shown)


dylanesque (approved):

src/forms/output/BenefitsLines.js
```diff
@@ -131,7 +131,8 @@ class BenefitsLinesComp extends Component {
       lines.push(line);
     }
 
-    // Get 'Unexpected template string expression' warning otherwise
+    // Label for split tooltip 'labels'/'label headers' that appear
+    // at the bottom. Really long.
     // @todo Abstract commonalities between graphs
     const labelHeaderFormatStart = `<span style="font-size: 10px">${getText(snippets.i_beforeMoney)}`,
```
turnerhayes, line 137: :( inline styles

(I know this is pre-existing, just makes me sad)

src/forms/output/ResourcesColumns.js
```diff
@@ -0,0 +1,183 @@
...
+              name        = { dataset.label }
+              data        = { dataset.data }
+              legendIndex = { dataseti }
+              x           = { dataseti }
+              label       = { dataseti } />
+          );
+
+      columns.unshift(column);
+
+      if (dataset.label === `Earned`) {
```
turnerhayes, line 95: This should probably use an ID or something; when this is localized, this will be a pain to use as the conditional.

src/forms/output/ResourcesColumns.js
```diff
@@ -0,0 +1,183 @@
...
+   * 
+   * @example
+   * let toShow = formatMoneyWithk({ axis: { defaultLabelFormatter: function () { return '10k'; }} });
+   * console.log(toShow);
+   * // If app's language code is 'en':
+   * // <span class="graph-label">$10k</span>
+   * // If app's language code is 'vi':
+   * // <span class="graph-label">10k$</span>
+   */
+  formatMoneyWithK = (highchartsObject) => {
```
turnerhayes, line 174: I think it might be better to name this something like `formatAsHumanReadableMoneyString` (ugly, but you get the idea)--I imagine that not all localizations will use a "K".

(not blocking)

src/forms/output/BenefitsLines.js
knod: Yeah, sorry. I meant to hunt down that selector.

src/forms/output/ResourcesColumns.js
knod: Good point. @todo.

src/forms/output/ResourcesColumns.js
knod: It's not forever, but Highcharts adds a 'k' and that's why this is here. In future, we'll probably want to just do all the formatting ourselves and this won't be needed.

src/forms/output/BenefitsLines.js
knod: If you have the time, feel free to tackle this one.

knod: @todo

Issue #936, opened 2018-10-31
Change `snippets` to `translations` everywhere


knod:
It's not quite a simple find and replace because in some places (definitely in 'Summary.js'), I've shortened `snippets` to `sn` or something for convenience. Sorry folks.

knod:
Closed by #1020

PR #1015, opened 2018-11-23
Rename 'snippets' to 'translations' in a lot of places. #936

turnerhayes:
Addresses #936 .

I *think* I got everything. Might have missed something.


knod (changes requested):
Awesome that you've taken this on! You did find a lot of them, but there are some still missing. Try searching in '/src' for ~'snips'~ 'snip'. There'll also still be a couple other names that won't make sense after this, but we can work on hunting them down separately.

knod:
We can possibly choose to do this piecemeal and just accept this as step one in a series.

knod (approved):

knod:
Didn't quite get everything, but gets us most of the way there!

PR #1017, opened 2018-11-24
Code style cleaning in '/src/components/dev',  #1018

knod:
Uses code conventions as they've been discussed so far. We can open up the discussion about changes, but that would be separate. Also using this opportunity to make personal notes for a code style type document. Let me know if it would be useful for me to make a gist.

### DIFF
--- a/src/components/dev/LocalizationReport.js
+++ b/src/components/dev/LocalizationReport.js
@@ -7,21 +7,31 @@ import {
   List,
   Modal,
 } from 'semantic-ui-react';
+
 import _ from 'lodash';
 import { getKeyPathsArray, getKeyPathStrings } from '../../utils/objectKeyPaths';
 import { getLocalizationData } from '../../localization/all';
+
 
 // Get copy of localization data
 const localizations = getLocalizationData();
 
 const ReportItem = function ({ keyPath, test, locKey, pass }) {
+
+  let iconName = `check square`,
+      color    = `green`;
+  if (!pass) {
+    iconName = `window close`;
+    color    = `red`;
+  }
+
   return (
     <List.Item>
       <List.Icon
-        name={ pass ? 'check square' : 'window close' }
-        size='large'
-        color={ pass ? 'green' : 'red' }
-        verticalAlign='middle' />
+        name          = { iconName }
+        color         = { color }
+        size          = { `large` }
+        verticalAlign = { `middle` } />
       <List.Content>
         { keyPath } { test } '{ locKey }'.
       </List.Content>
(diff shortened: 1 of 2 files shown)


src/components/dev/LocalizationReport.js
```diff
@@ -7,21 +7,31 @@ import {
...
   return (
     <List.Item>
       <List.Icon
-        name={ pass ? 'check square' : 'window close' }
-        size='large'
-        color={ pass ? 'green' : 'red' }
-        verticalAlign='middle' />
+        name          = { iconName }
+        color         = { color }
+        size          = { `large` }
```
turnerhayes, line 33: Question for discussion: do we want to change string prop values to be `propName="propValue"` instead of ```propName={ `propValue` }```? It's the convention I see most often, and makes JSX look a little more like HTML.

knod: Definitely @todo for a discussion. Feel free to add the issue if you want.

src/components/dev/LocalizationReport.js
```diff
@@ -30,31 +40,25 @@ const ReportItem = function ({ keyPath, test, locKey, pass }) {
...
-      keyPath = { keyPath }
-      test    = { test }
-      locKey  = { locKey }
-      pass    = { pass } />
-  );
-};
+const ReportList = function ({ results }) {
 
+  let items = [];
+  for (let result of results) {
```
turnerhayes, line 46: I would like to add `.map()`(/`.reduce()`) to the discussion topics for coding conventions :)

src/components/dev/LocalizationReport.js
knod: *future discussion

src/components/dev/LocalizationReport.js
```diff
@@ -65,22 +69,27 @@ class LocalizationReport extends Component {
     modalOpen:        false,
     modelLocKey:      null,
     compareLocKey:    null,
-    filter:           'false',
+    filter:           `false`,
```
turnerhayes, line 72: Also should discuss quoting conventions; most projects I see use single quotes for anything without interpolation, backticks only if it's interpolating or literals with newlines. I (marginally) prefer defaulting to double quotes, but I admit it doesn't make much difference.

src/components/dev/LocalizationReport.js
```diff
@@ -96,17 +105,22 @@ class LocalizationReport extends Component {
...
  
-  filterResults = (results, filter) => {
-    if (filter === 'all') { 
+  filterResults (results, filter) {
+    if (filter === `all`) { 
       return results;
     }
-    let pass = filter === 'true' ? true : false;
+
+    let pass = true;
```
turnerhayes: In this specific case, this could instead be replaced with `const [or let] pass = filter === 'true';`. Is this intentionally avoiding that construct for clarity?

knod: No, it's not, lol. Just blanked, thanks!

src/components/dev/LocalizationReport.js
```diff
@@ -96,17 +105,22 @@ class LocalizationReport extends Component {
...
+      pass = false;
+    }
+
     return results.filter((result) => {
       return result.pass === pass;
     });
   };
 
-  compareLocalizations = (modelLocKey, compareLocKey) => {
+  compareLocalizations (modelLocKey, compareLocKey) {
```
turnerhayes, line 119: :+1: I like replacing bound instance methods with regular instance methods wherever they don't need to be passed as callbacks (or don't use `this`)

src/components/dev/LocalizationReport.js
```diff
@@ -184,36 +206,36 @@ class LocalizationReport extends Component {
     });
 
     const filterOptions = [
-      { text: 'All', value: 'all' },
-      { text: 'Passing', value: 'true' },
-      { text: 'Failing', value: 'false' },
+      { text: `All`,     value: `all` },
```
turnerhayes, line 205: I wonder if there's an eslint rule that can fix these alignments automatically...

turnerhayes (approved):
Looks good. A couple of potential discussion topics, but nothing blocking.

src/components/dev/LocalizationReport.js
knod: That would be amazing.

PR #1020, opened 2018-11-25
Replace more snippet references. Fixes #1019

### LINKED ISSUE #1019: Finish renaming `snippets` to `translations`
#1015 got most of the places but there are still a few stragglers; search for `snip` (case insensitive) to find some.

turnerhayes:
Fixes #1019 .


knod (approved):
There may be other deeply hidden ones, but we'll have to burn those if and when we find them. It's looking good to me!

Issue #1018, opened 2018-11-24
Clean up code style stuff throughout the app


knod:
We'll use our current understanding of the code style for now. Further discussions can be had if needed, but they'll be saved for future changes.

Maybe we can use this opportunity to make an explicit style guide.

Assume '/src' comes before all these directories. Some directories still need to be broken up into smaller chuncks.

- [x] /components
   - [x] /dev for @knod 
   - [x] /prompts for @knod
   - [x] through ExternalLink.js for @knod
   - [x] through headings.js for @knod
   - [x] through Surrounder.js for @knod
- [x] /containers
- [x] /data
- [x] /forms
   - [x] /output
   - [x] others
- [x] /localization?
- [x] /programs
- [ ] /utils
- [x] App.js
- [x] index.js
- [x] index.css (`!important`, etc)

Issue #974, opened 2018-11-06
Remove anonymous functions from event handler props


turnerhayes:
When your `render()` method passes an anonymous function as a prop, a new function instance will be created and that prop's value will be different every time `render()` is called, which means that React will assume it needs to refresh the tree, even if nothing has "actually" changed. This can result in unnecessary DOM manipulation.

Wherever possible, we should have the function be a member of the component's class and bind it to `this` in the constructor, and just pass that function reference.

e.g.

Before:

```
class MyComponent extends React.Component {
  render() {
    return (
      <div>
        <button onClick={ () => alert(this) } />
      </div>
    );
  }
}
```

After:

```
class MyComponent extends React.Component {
  constructor(...args) {
    super(...args);

    this.handleClick = this.handleClick.bind(this);
  }

  handleClick() {
    alert(this);
  }

  render() {
    return (
      <div>
        <button onClick={ this.handleClick } />
      </div>
    );
  }
}
```

PR #1021, opened 2018-11-25
Clean up code style, part 2 of many parts. The prompts folder. #1018

knod:


knod:
No longer [WIP]. Probably.

src/components/prompts/PredictionsWarning.js
```diff
@@ -7,14 +7,15 @@ import {
...
+
 /**
  * Displays a model that requires the user to accept the terms and conditions before using the app
  * @extends React.Component
  * @param {boolean} termsAccepted - boolean indicating whether terms and conditions have been accepted by the user
  * @param {function} toggleDistrustConfirmed - function to set the termsAccepted in app state
  * @param {object} translations - object containing translations
  */
-class TermsAndConditions extends Component {
+class PredictionsWarningComp extends Component {
```
turnerhayes: I'm not a huge fan of the "Comp" suffix; can we just call it `PredictionsWarning`? Or `PredictionsWarningForm` or something?

src/components/prompts/PredictionsWarning.js
```diff
@@ -50,65 +52,63 @@ class TermsAndConditions extends Component {
...
+            key       = { `ReqCkBx1` }>
             <Form.Field>
               <Checkbox
                 checked = { this.state.Checkbox1 }
                 name    = { `checkbox1` }
-                onClick = { () => {return this.handleChange('checkbox1');} } />
-            </Form.Field>
-            <Form.Field>
-              { translations.i_checkboxLabel1 }
+                onClick = { () => { return this.handleChange(`checkbox1`); } } />
```
turnerhayes: As long as we're cleaning up code, we might as well make these handlers member functions instead of anonymous functions (see #974)

turnerhayes (changes requested):

src/components/prompts/PredictionsWarning.js
knod: For that one look at the bottom of the file. `PredictionsWarning` is indeed what's being exported. `PredictionsWarningComp` is being used with `withRouter()` to create `PredictionsWarning`. Can you find another way to express that difference?

turnerhayes: We could do `export { withRouter(PredictionsWarning) as PredictionsWarning };` (I think that syntax works). Or maybe call it `const WrappedPredictionsWarning = withRouter(PredictionsWarning);` at the bottom and export that with an alias instead, if you don't like inlining the `withRouter()`.

knod: Hmm, I'm not sure about this aliasing thing. I'm willing to go with that first option for now, but I'm very suspicious.

knod: Follow-up philosophical question - why the heck can't export syntax just be object syntax? What's up with that?

knod: Hmm, it doesn't seem to work inline that way. Dang. I really do prefer it that way as opposed to naming it before, but I guess that's a no go.

turnerhayes (approved):
Looks good!

src/components/prompts/PredictionsWarning.js
knod: @todo Discuss this code style.

PR #1023, opened 2018-11-25
Style cleans '/src/components' through 'ExternalLink.js'. #1018

knod:


dylanesque:
I don't understand why the const declarations here were changed to let.

src/components/CustomClient.js
```diff
@@ -24,7 +27,7 @@ class CustomClient extends React.Component {
   state = {
     toLoad: null,  // parsed json
     error:  null,
-    json:   '',
+    json:   ``,
```
turnerhayes, line 30: I feel like we should configure eslint to use backticks for string literals (I'm assuming that's possible).

src/components/CustomClient.js
```diff
@@ -46,15 +49,15 @@ class CustomClient extends React.Component {
   };
 
   submit = (event) => {
-    const { toLoad } = this.state;
+    let { toLoad } = this.state;
```
turnerhayes: Echoing @dylanesque 's question of why these were changed to `let`s

src/components/BetaWarning.js
```diff
@@ -1,22 +1,26 @@
...
       <Responsive
-        as='strong'
+        as = { `strong` }
         { ...Responsive.onlyComputer }>
         This tool is a prototype and should not be used to make financial decisions.
       </Responsive>
     </Menu.Item>
   );
-};  // End BetaWarning(<>)
+};
```
turnerhayes, line 23: Why remove the end comment? Is it because the component definition is short enough to fit on a screen? I kind of feel like if we want to use these end comments, we should use them consistently.

src/components/details.js
```diff
@@ -79,16 +80,18 @@ class HeadingWithDetail extends Component {
...
     );
   };
-};  // End <HeadingWithDetail>
+};  // Ends <HeadingWithDetail>
 
 
-/** @todo Have rows that can have details. Maybe other things too. */
-class RowWithDetail extends Component {};  // <End RowWithDetail>
+/* @todo Have rows that can have details. Maybe other things too. */
+class RowWithDetail extends Component {};
```
turnerhayes, line 94: I know this is existing code and not a part of this PR, but I am confused why we have an empty class at all. Is this event used anywhere? It doesn't look like it is. We can open a separate issue for this if necessary.

turnerhayes (changes requested):

src/components/CustomClient.js
turnerhayes: (this is a @todo, not blocking this PR)

knod: Definitely feel free to make an issue.

src/components/BetaWarning.js
knod: Well, if there were a three line function I don't imagine using and `// Ends...` comment, so the question is really where do we draw the line. Any ideas?

knod: Or maybe we use them at the end of every function, if statement, and for loop. That's also an option.

src/components/details.js
knod: Yeah, this can be a separate issue. I've been debating about that. I think the first one may be generic enough to be used with a row item as well, so we may just need to change the name.

@todo 

It'd be great to have an issue about this.

knod:
As far as `const` vs. `let`, I had thought the pre-existing choice (which hadn't been implemented) was to use `let` by default. I can't seem to find my notes on it now and I might be wrong about that.

I know it's something that needs more discussion. Is it a blocking change for this PR?

src/components/CustomClient.js
knod: See below.

dylanesque:
> As far as `const` vs. `let`, I had thought the pre-existing choice (which hadn't been implemented) was to use `let` by default. I can't seem to find my notes on it now and I might be wrong about that.
> 
> I know it's something that needs more discussion. Is it a blocking change for this PR?

For me, this is blocking, it makes the code that was altered more fragile. Values like destructured props should not change, and really should always be declared with const.

knod:
@dylanesque : I hear you feel strongly about that. I'll change back the destructuring ones. If you want a change, though, it's often a good idea to comment on the relevant lines with a 'request changes' review.

knod:
Dang, also forgot to destructure some props in there!

knod:
> Values like destructured props should not change

Did you mean destructured `this.state`? I don't think I changed any destructured props.

dylanesque:
> > Values like destructured props should not change
> 
> Did you mean destructured `this.state`? I don't think I changed any destructured props.

I meant "like" props, so state is similar enough in this example.

dylanesque (approved):

src/components/BetaWarning.js
knod: Thoughts?

turnerhayes: Yeah, it is kind of a difficult line to draw. I guess it's fine to remove it--I don't even have those sorts of comments in code I write (although maybe I should).

turnerhayes (approved):

PR #1024, opened 2018-11-25
Cleans code style of '/src/components' through 'headers.js`. #1018

knod:


src/components/headings.js
```diff
@@ -83,38 +83,45 @@ const ContentSubH1 = function ({ children }) {
...
 
-  let columnTitle = type.toLowerCase().replace(/\b[a-z]/g, (letter) => {
+  let styles      = { fontSize: `14px` },
+      toUpper     = function (letter) {
         return letter.toUpperCase();
-      }) + ' Type',
-      styles      = { fontSize: '14px' };
+      },
+      asLowerCase = type.toLowerCase(),
+      toReplace   = /\b[a-z]/g,
```
turnerhayes, line 91: I'm not clear why these were broken out into different vars when they're only used in one place.

turnerhayes (commented):
Only one comment; I wanted to explore the reasoning for breaking out variables

src/components/headings.js
knod: Readability. We try to limit the number of operations happening on one line and name things in a way that makes it clear what's going on. Thoughts?

dylanesque: I feel that readability can be serviced here by a comment summarized why this function is here. The breaking up would make sense if these sub-functionalities were used elsewhere, which doesn't seem the case.

turnerhayes: Personally, I think I found it easier to follow before--this way, I have to look around to find out what toReplace actually is, how toUpper is defined, etc. This may just be a personal preference thing; I'd be interested to hear what other people think about it. @dylanesque ?

knod: I love the sound of this discussion! In the meantime, are we holding off on merging for this? Or is this a discussion to have separately? Would you prefer that I change it back till we've discussed it more? Just so we can get this PR through.

dylanesque: Yeah, I don't see this as blocking, just putting in my two cents.

dylanesque (approved):

src/components/headings.js
turnerhayes: Agreed, not blocking.

turnerhayes (approved):

src/components/headings.js
knod: @todo then!

PR #1025, opened 2018-11-25
Cleans code style of '/src/components' through 'Surrounder`. #1018

knod:


turnerhayes (approved):

dylanesque (commented):
Unsure why the function style in StepBar.js was changed. Is there a scoping issue?

knod:
@dylanesque : Right now, our decision is to use arrow functions only when necessary.

PR #1026, opened 2018-11-25
Cleans code style of '/src/containers'. #1018

knod:


src/containers/VisitPage.js
```diff
@@ -271,78 +277,74 @@ class VisitPage extends Component {
...
+          shouldRequestConfirmation = { this.shouldConfirmLeave }
+          isBlocking                = { this.state.isBlocking } />
 
         {/* = LINKS? = */}
         {/* We should probably remove this. If we want to
          * do this we might do this a different way at this
          * point. Perhaps a user's page should be a route
          * in VisitPage? */}
-        { this.state.redirect ? (
+        { (this.state.redirect) ? (
```
turnerhayes: Why add the extra parens?

src/containers/VisitPage.js
```diff
@@ -355,20 +357,18 @@ class VisitPage extends Component {
...
-          distrustConfirmed === false ? (
-            <PredictionsWarning
-              distrustConfirmed       = { distrustConfirmed }
-              toggleDistrustConfirmed = { this.props.funcs.toggleDistrustConfirmed }
-              translations={{ ...translations.warningModal }} />
-          ) : (
-            null
-          )
-        }
+        { (distrustConfirmed === false) ? (
```
turnerhayes: (same question as above)

src/containers/VisitPage.js
knod: Well, in more complex conditionals I find it easier to read if there are parens around the conditional. If I'm going to do it elsewhere, I thought I'd try being consistent and do it here too. Thoughts?

src/containers/VisitPage.js
knod: Same answer. This isn't a super complex one, though it's a bit more complex and the parens help even here. I also think it makes it look even more like a regular if statement, which is what I find easiest to read.

src/containers/VisitPage.js
turnerhayes: I see. I kind of feel like it clutters things up to have extra parens. I think for conditionals that don't involve operations or multiple lines (for the conditional part), I would prefer leaving off the parens.

src/containers/VisitPage.js
knod: @turnerhayes : Would we remove the parens here, or do they stay because this has an operation?

dylanesque (approved):

src/containers/VisitPage.js
turnerhayes: Debatable--I would say equality is a simple enough operation, but that does get into the question of "what's simple enough" and that's probably not something we want to wade into.

Would this be okay to replace with `!distrustConfirmed`? Or can it have falsy-but-not-`false` values?

knod: I can switch it up

knod: I don't think it has falsy/truthy values...

turnerhayes (approved):

PR #1027, opened 2018-11-25
Cleans code style of '/src/data'. #1018

knod:
Code style clean up based on conventions that have been agreed upon most recently.

### DIFF
--- a/src/data/federal/2017/SNAPData.js
+++ b/src/data/federal/2017/SNAPData.js
@@ -2,22 +2,17 @@
  *
  * RESOURCES:
  * Derived from {@link https://www.masslegalservices.org/content/online-snap-calculator}
- * See more resources in the actual code since some
- *      can't be referenced in these comments.
+ * {@link http://www.mass.gov/eohhs/consumer/basic-needs/financial/program-eligibility-charts-and-tables.html}
+ * {@link http://www.mass.gov/eohhs/gov/departments/dta/program-eligibility-charts-and-tables.html}
+ * Standard deductions: {@link http://www.mass.gov/eohhs/docs/dta/eligibility-charts/c-snap-364-400.pdf}
+ * Utility allowance categories: {@link http://www.mass.gov/eohhs/docs/dta/eligibility-charts/c-snap-364-945.pdf}
+ * Net income limits: {@link http://www.mass.gov/eohhs/docs/dta/eligibility-charts/c-snap-364-970.pdf}
+ * SNAP allowance maximums: {@link http://www.mass.gov/eohhs/docs/dta/eligibility-charts/c-snap-364-980-hh-1-10.pdf}
  *
  * @namespace
  */
 const SNAPData = {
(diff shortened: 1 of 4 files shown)


knod:
jsdoc doesn't seem to like the `.pdf`s and `.html`s. Not sure how to fix that while keeping at least the text there so the files can at least be named.

knod:
Why is netlify still failing?

src/data/federal/2017/SNAPData.js
```diff
@@ -1,10 +1,24 @@
...
+ * Derived from {@link https://www.masslegalservices.org/content/online-snap-calculator}
+ * See more resources in the actual code since some
+ *      can't be referenced in these comments.
+ *
+ * @namespace
  */
 const SNAPData = {
+  // Code-style note: Properties have to use quotes
+  /* MORE RESOURCES
+   * (Since pdfs and htmls can't seem to be referenced safely in jsdoc comments)
```
turnerhayes: I think you can put them in the docblock as long as it's before the `@namespace` tag

src/data/federal/federalPovertyGuidelines.js
```diff
@@ -1,6 +1,12 @@
+/** Data tables for Federal poverty guildelines/income limits.
+ *
+ * RESOURCES:
+ * {@link https://aspe.hhs.gov/poverty-guidelines}
+ *
+ * @namespace
+ */
 const federalPovertyGuidelines = { 0: 0, 1: 12060, 2: 16240, 3: 20420, 4: 24600, 5: 28780, 6: 32960, 7: 37140, 8: 41320, 'eachAdditional': 4180 };
+// {@link http://www.mass.gov/eohhs/docs/masshealth/deskguides/fpl-deskguide.pdf}
```
turnerhayes: Can this be put into some JSDoc block somewhere?

knod: I think not. Because it's a file name (with '.pdf' in its name), jsdoc seems to want to find that file in our folders and gets upset when it can't find it. It created errors when I tried to include that and then did `npm run build`. If that answers your questions, mark this comment as resolved.

src/data/federal/2017/SNAPData.js
knod: See below.

src/data/federal/federalPovertyGuidelines.js
knod: I could always have done something else wrong...

knod: Trying it again now that other problems have been resolved.

knod: If it works, it'll be very strange behavior since removing those seemed to fix the local problems in that first commit, even though other things were causing other issues.

knod: Welp, I guess it's all good now, lol.

turnerhayes (approved):

PR #1028, opened 2018-11-26
Cleans code style of '/src/forms/output'. #1018

knod:


src/forms/output/BenefitsLines.js
```diff
@@ -112,12 +112,12 @@ class BenefitsLinesComp extends Component {
     // Label for split tooltip 'labels'/'label headers' that appear
     // at the bottom. Really long.
     // @todo Abstract commonalities between graphs
-    const labelHeaderFormatStart = `<span style="font-size: 10px">${getText(translations.i_beforeMoney)}`,
-          labelHeaderFormatEnd   = `{point.key:,.2f}${getText(translations.i_afterMoney)}</span><br/>`,
-          labelHeaderFormat      = labelHeaderFormatStart + labelHeaderFormatEnd;
+    let labelHeaderFormatStart = `<span class="tooltip-label-header">${getText(translations.i_beforeMoney)}`,
```
turnerhayes, line 115: :tada: Hooray for getting rid of inline styles! :tada:

src/forms/output/BenefitsLines.js
```diff
@@ -81,21 +81,21 @@ class BenefitsLinesComp extends Component {
       classes += ` ` + className;
     }
 
-    const multiplier    = multipliers[ timescale ],
-          resources     = activePrograms,
-          currentEarned = client.current.earned * multiplier,
-          getText       = textFromTranslatedElement;
+    let multiplier    = multipliers[ timescale ],
```
turnerhayes, line 84: Not clear on why these were all changed to `let`? Notwithstanding the `prefer const` question, it seems like a number of these could probably be considered consts (e.g. the component we're exporting below).

knod: Hmm. I agree this is an important question. How about we save this for a conversation on how to identify what variables need to be prevented from being reassigned?

knod: How about this - after these PRs that already exist, I won't change any `const`s to `let`s until we've had a chance to discuss this further.

turnerhayes: :+1:

turnerhayes (approved):

PR #1031, opened 2018-11-29
Cleans through 'src/forms/CurrentExpenses.js'. #1018

knod:


src/forms/CurrentBenefits.js
```diff
@@ -23,45 +25,48 @@ import { allBenefitOrders } from '../programs/allBenefitOrders';
...
+          value       = cloneDeep(this.props.current.benefits);
 
     if (inputProps.value) {
       value.push(benefitName);
       value.sort(
         // Make sure benefits are in the correct order
         (a, b) => {
-          const aIndex = this.props.benefits.indexOf(a);
-          const bIndex = this.props.benefits.indexOf(b);
+          // @todo Is this code right? Don't we need `.current` in here?
```
turnerhayes: `props.benefits` here is a list of all available benefits, rather than the client's benefits, so no `current` needed.

src/forms/CurrentBenefits.js
```diff
@@ -23,45 +25,48 @@ import { allBenefitOrders } from '../programs/allBenefitOrders';
...
   
     for (let benefitIndex = 0; benefitIndex < benefits.length; benefitIndex++) {
-      const benefit = benefits[ benefitIndex ];
-
-      const translationKey = `i_has_${benefit}_label`;
+      const benefit        = benefits[ benefitIndex ],
+            translationKey = `i_has_${benefit}_label`;
 
+      // @todo This would have to be handled in every translation situation.
+      // @todo Also, is there a way we can include this in the translation report?
```
turnerhayes, line 69: That seems doable. This is, IMO, one of the weaknesses in the current i18n approach--there doesn't seem to be a way to pass variables to the translation messages (only static placeholders). But that's a whole other discussion!

src/forms/cashflow.js
```diff
@@ -83,7 +85,7 @@ class CashFlowInputsRow extends Component {
     return (str) => {
       if (!isNonNegNumber(str)) {
         let message    = `That number doesn't look right`,
-            numPeriods = str.match(/\./g) || [];
+            numPeriods = str.match(PERIODS_REGEX) || [];
```
turnerhayes: The `num` prefix makes it sound like this variable is storing a count of periods, rather than the results of the match. We could do something like `numPeriods = (str.match(PERIODS_REGEX) || [undefined]).length` (although that's super ugly). Probably better to just name it something like `periodsInString`.

turnerhayes (changes requested):

src/forms/CurrentBenefits.js
knod: It is! Still an important discussion to have. @todo

knod: And thanks for the brainstorming.

src/forms/cashflow.js
knod: True, I'll change it. Generally, though, I think variable name stuff isn't going to be the focus of these PRs.

knod: I think I'll just call it `periods` to simplify.

src/forms/CurrentBenefits.js
turnerhayes: It might be good to remove this comment, just so we're not chasing geese when we pick this up later (assuming your concerns are assuaged).

knod: Oh, true. My bad.

turnerhayes (approved):
Looks good! :+1:

PR #1032, opened 2018-12-01
Cleans '/src/forms' through 'Household.js'. #1018

knod:
Many inline styles removed.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -153,7 +153,7 @@ Used to fill vertical space.
 .ui.form .cashflow .cashflow-column-label {
   text-align: left;
   width: auto;
-  padding-left: 1em !important;
+  padding-left: 1em;
 }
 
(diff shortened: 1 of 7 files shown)


src/index.css
```diff
@@ -153,7 +153,7 @@ Used to fill vertical space.
 .ui.form .cashflow .cashflow-column-label {
   text-align: left;
   width: auto;
-  padding-left: 1em !important;
+  padding-left: 1em;
```
turnerhayes, line 156: Yay for getting rid of `!important`!! :tada:

turnerhayes (approved):
Looks good!

src/index.css
knod: Agreed. I may add 'index.css' to the code style list. We've got a couple more of those in there from what I could see.

PR #1036, opened 2018-12-02
Cleans '/src/programs'. #1018

knod:


turnerhayes (approved):
I really like a lot of the variable renames, makes them much more clear!

PR #1037, opened 2018-12-02
Cleans '/src/utils' through 'interpolation.js'. #1018. Also #708.

knod:


src/utils/CLIENT_DEFAULTS.js
```diff
@@ -9,20 +9,20 @@ import { cloneDeep } from 'lodash';
  */
 const CLIENT_DEFAULTS = {
   
-  // Property Name                | Default Value   | Valid Value Description
+  // Property Name                | Default Value    | Valid Value Description
 
-  // @todo get this value from the app somewhere  
-  USState: 'MA',                   // Two-letter code denoting the state the client is in
+  // @todo get this value from the app somewhere
+  USState: `MA`,              // Two-letter code denoting the state the client is in
```
turnerhayes: Might want to look into ESLint config option [spacing?](https://eslint.org/docs/rules/key-spacing#align) to fix this. I'm reluctant to disable ESlint for the whole file; at most, I hope we could disable just a specific rule to fix this, preferably in a block (`/* eslint-disable */ ... /* eslint-enable */`) rather than whole-file.

turnerhayes (changes requested):

src/utils/CLIENT_DEFAULTS.js
knod: The spacing rules are the thing that's causing the problem. I didn't find any spacing rules that would be more useful in there. I disabled them in this block.

turnerhayes: Can you disable only the rule that's causing the problem in that block, rather than a total disabling of ESLint? Also add a comment explaining why it's being disabled in that block?

knod: Ok, I've found a way to disable for a specific rule. I'm going to change this up, though, and disable the rule for the whole object. If/when new values are added, I don't want them to have the same unexpected result as before and have to figure it out all over again.

This also makes me question whether this column thing should be around at all. Maybe we remove the column headings and just go back to regular spacing rules. Thoughts?

dylanesque (commented):
What's the purpose of this forUpdate function in convertForUpdate.js? All it seems to be doing is returning itself?

src/utils/CLIENT_DEFAULTS.js
turnerhayes: I kind of agree about the column headings--I see their value, but we could probably accomplish the same thing without them (I don't think it's too confusing what part of the code is what without the columns).

knod:
@dylanesque : I don't see a `forUpdate` function, just an object. Are we talking about the same thing?

src/utils/CLIENT_DEFAULTS.js
knod: I'll just get rid of them then.

turnerhayes:
@dylanesque are you referring to this?:

```
const convertForUpdate = function ({ name, route, ...otherProps }) {
   let forUpdate = {
     ...otherProps,
     route: route || name,
   };
   return forUpdate;
 };  // End convertForUpdate()
```

dylanesque:
> @dylanesque are you referring to this?:
> 
> ```
> const convertForUpdate = function ({ name, route, ...otherProps }) {
>    let forUpdate = {
>      ...otherProps,
>      route: route || name,
>    };
>    return forUpdate;
>  };  // End convertForUpdate()
> ```

That's what I meant, but the quoted code you have above doesn't match at all what I'm seeing in the PR, namely: 
```const convertForUpdate = function ({ name, route, ...otherProps }) {
   let forUpdate = {
       return forUpdate;	 
 };
 export { convertForUpdate };	export { convertForUpdate };
```

knod:
@dylanesque : that's just github skipping lines - it avoids showing the whole file and concentrates on what's been changed. If you see, there's a line highlighted in blue that has icons at the left edge. If you click on those, it'll expand to show more of the code. You can also click 'view' for the file and see the new version in full.

src/utils/CLIENT_DEFAULTS.js
knod: That's done. Is this ready for merging?

turnerhayes (approved):

PR #1038, opened 2018-12-02
Cleans /src/utils/charts. #1018

knod:


src/utils/charts/chartFormatting.js
```diff
@@ -1,23 +1,22 @@
 import _ from 'lodash';
 
-/** Helper functions to format vlaues
+/** Helper functions to format vlaues for charts
```
turnerhayes: "to format vlaues" => "to format *values*"

turnerhayes (changes requested):
One minor typo (preexisting, but might as well fix it while we're cleaning stuff up here).

src/utils/charts/chartFormatting.js
knod: Done! Better?

turnerhayes (approved):

src/utils/charts/chartFormatting.js
turnerhayes: :+1:

PR #1041, opened 2018-12-03
Removes `!important` from index.css wherever possible.

knod:
Only two uses remaining and those are fighting semantic-react-ui's own use of `!important`. Don't know why the heck that's in their library at all :P


turnerhayes (approved):
:+1: :+1:

PR #1051, opened 2018-12-14
Refactor charts to abstract operations. Fixes #1010

### LINKED ISSUE #1010: Abstract chart/graph commonalities (in same dir as chart files)

knod:


knod:
Not WIP anymore. Reviews welcome!

src/forms/output/chartStringTransformers.js
```diff
@@ -2,32 +2,38 @@
...
+ *     options.
  *
- * @params {object} translationObj React element containing
- *     children that are strings or other React elements.
+ * @params {object} translations The strings for the app's current
+ *     language.
+ * @params {object} Highcharts Where to set the options.
  *
- * @returns {string}
+ * @returns {nothing}
```
turnerhayes, line 12: I *think* the convention for functions with no return value in JSDoc is the type `void`, but Googling it comes up with fairly disparate suggestions (e.g. Closure Compiler says don't include the tag at all if it returns nothing). So maybe there's no consensus.

src/forms/output/chartStringTransformers.js
```diff
@@ -2,32 +2,38 @@
...
-  if (typeof children === `string`) {
-    return children;
-
-  // To handle more complex translationObj objects
-  } else if (Array.isArray(children)) {
+const setThousandsSeparator = function (translations, Highcharts) {
+  // We really need our own number formatting functions
+  let separator = textFromTranslatedElement(translations.i_thousandsSeparator);
+  // This is for Highcharts to format pure numbers
+  Highcharts.setOptions({ lang: { thousandsSep: separator }});
```
turnerhayes, line 18: Since this is setting a universal Highcharts option, rather than a per-chart option, could this be called in a single "config global Highcharts" section, rather than on each graph? It looks like right now it's being called multiple times by each of the different Highcharts charts.

turnerhayes (approved):
A couple of minor comments, but nothing blocking.

src/forms/output/chartStringTransformers.js
knod: Hmm, good note. We should decide on our own convention and then document it. @todo discussion.

src/forms/output/chartStringTransformers.js
knod: I thought this was creating an instance of Highcharts, but if it is global, that's a great idea! @todo.

PR #1053, opened 2018-12-19
de-nests the first section of CurrentExpenses.js and en.js

ValerieKenyon:
I only de-nested un-reimbursed non-medical child care.


knod (approved):
Looking great!

PR #1054, opened 2019-01-01
[WIP] Clarifies 'Predictions' section, fixes #866

### LINKED ISSUE #866: Clarify Predictions section
With the help of a moonlighting member, we've have identified a few specific problems in the Predictions section that we can try to work on.

1. Things are a bit out of order. The summary for the given information is only presented after asking for information about the future.
1. The input asks for future income as opposed to asking about the amount of a raise.
1. (No great solution yet) We want to provide our users with two things, not just one:
   1. Information about how a specific change will affect their circumstances
   1. The long-term view of their benefit changes

That last one is tough because we do really want to provide both those things, but they're confusingly similar. I'm not sure separating them into two sections would do much for us.

For the problem of order of presentation, we're going to switch so that a summary of what they've told us is at the top and, in the future, we might break that out into its own section. As for the raise, we'll switch the question we're asking and how its processed.

-----------

Previous title and text:
Title: Make user more aware of/clarify future inputs in Predictions section or remove them

One idea: add a colored pulse to Predictions input until user interaction.

Anyone have others? Better wording, maybe? Take away the need for it completely? It does seem a bit weird that they need to put in a future amount before they get any results since we can extrapolate ~them~ [the results] ourselves.

knod:
An attempt, anyway. We'll see how it goes. See #866 for more details.


wacii:
The design is problematic. Although I'm not sure how much I can help with that.

At the very least, the vertical spacing between "Right now" and the future income input should be increased. There is more vertical space between the future income input and "With the new pay, how could your benefits change?" than with "Right now", and the first two are strongly related.

Now that the future income input is not the first entry on the page it should get a stronger label.

In general, the current summary and future predictions aren't well differentiated. Not sure how to fix that as the page is already quite busy. I do like the idea of making the current summary more concise and rendering on every page. Although I'm not sure where I'd put that. Super helpful!

knod:
@wacii : Thanks for the feedback! Yeah, this is still a WIP and that was just a step along the way of metamorphosis. It's still not done. I'm working from the top down. I've gotten to the basics of the 'Raise' section. Ignore the graphs part for now and let me know what you think. If you want to see it with a cliff, use the data from row 8 of https://docs.google.com/spreadsheets/d/15LyR9yELAfcngj-c7vMdI630b6DwuuXV-dQEJDvU4gE/edit?usp=sharing.

If you look at where the 'Raise' section is now, that's going to be the raise and, if needed, the summary. The 'charts' section is just going to be a 'long view' and have the two long-term graphs. I think I'm going to play with dropping the table and the bar graph to see how that goes over because I don't really know how to include them anymore.

If you have any more thoughts, on what I've said or what I generate, keep 'em coming!

Issue #937, opened 2018-10-31
Check for `instanceof Error`


knod:
In 'src/utils/valueFixers.js' `isNumberlike()` change from
```js
if (typeof trueOrError === `object`) { //...
```
to
```js
if (typeof trueOrError instanceof Error) { //...
```

PR #1055, opened 2019-01-29
Fix error handling, Issue #937

migueldcepeda:
Let me know if I misunderstood and this fix is not as simple as what I did here.

### DIFF
--- a/src/utils/valueFixers.js
+++ b/src/utils/valueFixers.js
@@ -37,7 +37,7 @@ const toNumber = function (numberOrString) {
 
   let trueOrError = isNumberlike(numberOrString);
 
-  if (typeof trueOrError === `object`) {
+  if (typeof trueOrError instanceof Error) {
     throw (trueOrError);
   }
 
--- a/src/utils/valueFixers.js
+++ b/src/utils/valueFixers.js
@@ -37,7 +37,7 @@ const toNumber = function (numberOrString) {
 
   let trueOrError = isNumberlike(numberOrString);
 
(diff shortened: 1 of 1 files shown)


knod:
This is what we were looking for! And it looks like it may be revealing a flaw in our code. We'll have to look into the test failure to see what's going on. We'll try to do that soon and let you know.

Thank you!

src/utils/valueFixers.js
```diff
@@ -37,7 +37,7 @@ const toNumber = function (numberOrString) {
 
   let trueOrError = isNumberlike(numberOrString);
 
-  if (typeof trueOrError === `object`) {
+  if (typeof trueOrError instanceof Error) {
```
knod: I see it now! More needs to be changed here. `typeof Error === 'object'`, so when `isNumberlike()` returns an `Error`, this will come out false and won't behave the way we want.

Not sure I described that clearly. Feel free to ping me if I need to clarify.

knod (changes requested):

src/utils/valueFixers.js
migueldcepeda: Ah I see now. The line should be written as `if (trueOrError instanceof Error)` since `typeof trueOrError` will always return `object` and `object instanceof Error` will always return `false`.

I can work on making that fix and resubmitting.

knod:
Locally passing. Re-running the Travis PR build.

knod (approved):

knod (approved):
Looks great! Thanks a ton!

